PR #12, opened 2025-02-05
getBalance implementation

Dodecahedr0x:
Uses `getAccountInfo` to potentially reuse the account fetching logic

### DIFF
--- a/crates/core/src/rpc/minimal.rs
+++ b/crates/core/src/rpc/minimal.rs
@@ -23,14 +23,6 @@ use solana_sdk::{
 pub trait Minimal {
     type Metadata;
 
-    #[rpc(meta, name = "getBalance")]
-    fn get_balance(
-        &self,
-        meta: Self::Metadata,
-        pubkey_str: String,
-        config: Option<RpcContextConfig>,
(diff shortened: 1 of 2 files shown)


crates/core/src/rpc/minimal.rs
```diff
@@ -23,14 +23,6 @@ use solana_sdk::{
 pub trait Minimal {
     type Metadata;
 
-    #[rpc(meta, name = "getBalance")]
-    fn get_balance(
```
lgalabru: The shape of these traits are coming from the canonical implementation.
I personally don't necessarily agree with the way they've been designed, but I think staying consistent with the canonical design would help us maintaining this set of methods.
Could we please keep the `get_balance` implementation in the `Minimal` trait?

Dodecahedr0x: Sure, I wasn't aware of the canonical design and did that because I wanted to reuse the implementation from accounts data. Ideally there should be a trait implemented by the run loop that fetches and updates local accounts if they are missing but I was facing ownership issues.

Could you point me to the official design please?

lgalabru (approved):
Love it! Thank you @Dodecahedr0x!!

PR #34, opened 2025-02-12
Basic integration test

Dodecahedr0x:
starting to work on tests, created this PR to have some feedbacks on the structure

## Summary by CodeRabbit

- **New Features**
  - Enhanced the simulation command interface with updated aliases and a new option for specifying plugin configurations.
  - Improved transaction reporting in the UI by tracking successfully processed transactions.

- **Improvements**
  - Refined configuration handling with updated defaults for simulation, networking, and RPC, leading to a smoother user experience.
  - Streamlined background operations and error management for more reliable interactions.

- **Tests**
  - Added integration tests to validate simulation readiness, timing accuracy, and SOL transfers.


crates/core/tests/integration.rs
```diff
@@ -0,0 +1,74 @@
...
+            if ticks > 100 {
+                let _ = test_tx.send(true);
+            }
+        }
+    });
+
+    match test_rx.recv_timeout(Duration::from_millis(1000)) {
+        Ok(_) => (),
+        Err(_) => panic!("not enough ticks"),
+    }
```
lgalabru, line 92: I think it makes sense to me. As we go, we should create as many helpers as possible, to make testing easy and boring. 
@MicaiahReid, is there anything specific you'd like to see?
Do we want to setup https://crates.io/crates/test-case ?

crates/core/tests/integration.rs
```diff
@@ -0,0 +1,74 @@
...
+                _ => (),
+            }
+
+            if ticks > 100 {
+                let _ = test_tx.send(true);
+            }
+        }
+    });
+
+    match test_rx.recv_timeout(Duration::from_millis(1000)) {
```
MicaiahReid: Hmm, I wonder if this will be flaky in CI. The race condition here seems like it could be problematic, right?

crates/core/tests/integration.rs
MicaiahReid: For unit testing, I'd love to see each rpc endpoint tested with a few edge cases. `test-case` is great for that, when we get there.

For integration tests, I'd love to see something like:
 - start surfpool
 - get some vm data and assert pre-condition
 - send some transaction(s)
 - get some vm data and assert post-conditions

If we have easy interfaces for starting surfpool, creating/sending txs, and calling the rpc to get the data, we probably don't need to go overboard with `test_case` for that. We'd have to have very generic interfaces for the post conditions.

---

Some other considerations. It probably is not the best idea to point to a real mainnet rpc for tests? Or does that matter.

We don't want to have flaky tests if an rpc is slow or we're being rate limited.

This makes me wonder if we could eventually pull off a few things:
 - surfpool for CI
 - surfpool without an upstream
 - surfpool pointing to surfpool as its upstream

With this, for our test suite, we could:
 - start sp with no upstream
 - do some writes to update the state
 - start another sp, pointing to sp1
 - continue with the integration tests

crates/core/tests/integration.rs
Dodecahedr0x: I was worried the test would hand on forever (and consume CI credits) if it failed. I can increase that value to reduce chances of problems

crates/core/tests/integration.rs
lgalabru: > surfpool pointing to surfpool as its upstream

🤯 I love it. that upstream could return some known hard coded data, so that we test that the "fetch remote fallbacks" are kicking in when expected.

Dodecahedr0x:
I wanted to implement the integration test suggested by @MicaiahReid before moving to build simple Anchor test programs but ran into an issue I can't figure out... (https://github.com/txtx/surfpool/pull/34/files#diff-a99b0f7e34db21b03fc510b5abebc18c48817d00dc8f2b32ef6f8aa866dd8021R194)
So I'm pushing it here to get some feedback, implemented a flaky workaround in the mean time

crates/core/tests/integration.rs
```diff
@@ -0,0 +1,228 @@
...
+    }
+}
+
+#[tokio::test]
+async fn test_simnet_some_sol_transfers() {
+    let airdrop_keypairs = (0..10).map(|_| Keypair::new()).collect::<Vec<_>>();
+    let airdrop_addresses: Vec<Pubkey> = airdrop_keypairs.iter().map(|kp| kp.pubkey()).collect();
+    let airdrop_token_amount = LAMPORTS_PER_SOL;
+    let bind_host = "127.0.0.1";
+    let bind_port = 8899;
```
MicaiahReid: This is good for now, but for the future, we may want to find a free port. If we have a full test suite, tests will take forever without parallelization, so we'll need variable ports. I did this in a past project: https://github.com/hirosystems/chainhook/blob/4e4960897983206c376bb7d94a9ba727802ec043/components/chainhook-cli/src/service/tests/helpers/mod.rs#L25-L34

crates/core/tests/integration.rs
```diff
@@ -0,0 +1,228 @@
...
+
+        Box::pin(future::ready(Ok(full_client.send_transaction(data, None))))
+    }))
+    .await
+    .into_iter()
+    .collect::<Result<Vec<_>, _>>()
+    .expect("Transfers failed");
+
+    // Wait for all transactions to be received
+    // let mut processed = 0;
```
MicaiahReid: Is there a chance that because we've `await`ed above, one of them has already emitted the processed event, so we're too late when we start the loop below?

Maybe we need to start that loop in a thread _before_ sending the transfers, and the test will fail if that thread isn't closed within a certain amount of time?

MicaiahReid: Sorry I forgot to ping: @Dodecahedr0x

lgalabru (approved):
Thanks @Dodecahedr0x !!

PR #38, opened 2025-02-13
feat: draft --watch option

lgalabru:
Todo: update and publish svm addon - should expose the rust code, instead of just the diffs between the `.so` files.


lgalabru:
Superseded by #41

PR #39, opened 2025-02-13
Support for deployments from anchor cli

Dodecahedr0x:
This PR adds features and fixes to enable deploying and testing programs.

However, Anchor by default uses websockets and TPU client, which are not available in Surfpool yet. So deploying requires running `solana program deploy -u localhost target/deploy/surfpool_test.so --use-rpc` to do everything through the RPC, and testing programs works (transactions are correctely received and processed by Surfpool) but Anchor tries to stream program logs through websockets and so it currently gets no response.

## Summary by CodeRabbit

- **New Features**
  - Refined the command line interface by simplifying command aliases and updating the shortcut for network host configuration.
  - Now provides accurate block height reporting.

- **Bug Fixes**
  - Streamlined account data retrieval for more consistent and efficient responses.

- **Chores**
  - Removed redundant logging statements during request processing.

### DIFF
--- a/crates/core/src/rpc/accounts_data.rs
+++ b/crates/core/src/rpc/accounts_data.rs
@@ -1,3 +1,5 @@
+use std::collections::HashMap;
+
 use crate::rpc::utils::verify_pubkey;
 use crate::rpc::State;
 
@@ -5,6 +7,7 @@ use jsonrpc_core::futures::future;
 use jsonrpc_core::BoxFuture;
 use jsonrpc_core::{Error, Result};
 use jsonrpc_derive::rpc;
+use solana_account::Account;
 use solana_account_decoder::parse_token::UiTokenAmount;
 use solana_account_decoder::{encode_ui_account, UiAccount, UiAccountEncoding};
 use solana_client::rpc_config::RpcAccountInfoConfig;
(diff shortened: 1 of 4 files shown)


crates/core/src/rpc/accounts_data.rs
```diff
@@ -165,37 +170,31 @@ impl AccountsData for SurfpoolAccountsDataRpc {
...
-            // Fetch all accounts at once and then only use those that are missing
-            let fetched_accounts = join_all(accounts.iter().map(|(pk, _)| {
-                let rpc_client = Arc::clone(&rpc_client);
-
-                async move { (pk, rpc_client.get_account(&pk).await.ok()) }
-            }))
-            .await;
-
-            // Save missing fetched accounts
+            let fetched_account =
```
lgalabru: Should we first load what we have locally, then fetch from the remote the ones we don't have, then merge?

crates/core/src/rpc/mod.rs
```diff
@@ -110,10 +110,7 @@ impl Middleware<Option<RunloopContext>> for SurfpoolMiddleware {
             state: self.context.clone(),
             mempool_tx: self.mempool_tx.clone(),
         });
-        // println!("Processing request {:?}", request);
-        Either::Left(Box::pin(next(request, meta).map(move |res| {
-            // println!("Processing took: {:?}", start.elapsed());
-            res
-        })))
+        println!("Processing request {:?}", request);
```
lgalabru: Debugging artifact

crates/core/src/rpc/accounts_data.rs
Dodecahedr0x: It costs the same for most RPC providers but I fixed that nonetheless, it's more efficient

Dodecahedr0x:
Addressed feedbacks in the last commit @lgalabru

lgalabru (approved):

PR #50, opened 2025-03-18
Add t2 rpc methods

### LINKED ISSUE #15: RPC Methods to implement - T2: compatibility with explorers
Todo: Identify the RPC methods required by http://github.com/solana-foundation/explorer, and move the remaining ones to https://github.com/txtx/surfpool/issues/20

- [x] getHealth
- [x] getFirstAvailableBlock
- [x] getEpochSchedule
- [x] getTransactionCount
- [x] getBlockTime
- [x] getBlock
- [x] getSupply
- [x] getRecentPerformanceSamples

adpthegreat:
Closes #15 

- getTransactionCount has been implemented already
- getRecentPerformance_samples has been implemented already

- getSupply Done
- getFirstAvailableBlock Done 
- getBlockTime Done
- getBlock Done

- getEpochSchedule 

Was seeing some comments -> `// TODO: Refactor `agave-validator wait-for-restart-window` to not require this method, so
    //       it can be removed from rpc_minimal` does it affect `get_epoch_schedule` too? I should just write the method using the `rpc_client` ?  

Also, writing the tests is a bit tricky, I just hardcoded the assertions against my local validator values, but I confirmed that the rpc client methods work 

Please review @lgalabru

### DIFF
--- a/crates/core/src/rpc/accounts_scan.rs
+++ b/crates/core/src/rpc/accounts_scan.rs
@@ -5,14 +5,13 @@ use solana_client::rpc_config::{
     RpcAccountInfoConfig, RpcLargestAccountsConfig, RpcProgramAccountsConfig, RpcSupplyConfig,
     RpcTokenAccountsFilter,
 };
-use jsonrpc_core::futures::future::{self, join_all};
-// use solana_client::rpc_response::RpcResponseContext;
+use jsonrpc_core::futures::future::{self};
 use solana_client::rpc_response::{
     OptionalContext, RpcAccountBalance, RpcKeyedAccount, RpcSupply, RpcTokenAccountBalance,
 };
-use solana_rpc_client_api::response::{Response as RpcResponse, RpcResponseContext};
-use solana_sdk::commitment_config::CommitmentConfig;
-
+use solana_rpc_client_api::response::{Response as RpcResponse};
+use solana_client::rpc_response::{RpcResponseContext, RpcApiVersion};
+use solana_sdk::commitment_config::{CommitmentConfig,CommitmentLevel};
 use super::not_implemented_err_async;
 use super::RunloopContext;
 use crate::rpc::State;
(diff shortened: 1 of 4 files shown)


lgalabru:
Thanks so much for the PR @adpthegreat, this looks super promising!
Review coming your way today/tomorrow!

crates/core/src/rpc/accounts_scan.rs
```diff
@@ -95,19 +96,36 @@ impl AccountsScan for SurfpoolAccountsScanRpc {
...
+
+                response
+            } else {
+                let response = rpc_client
+                    .supply()
+                    .await
+                    .map_err(|err| Error::invalid_params(format!("failed to get supply: {err:?}")));
+
+                response
+            }
```
lgalabru, line 128: I believe we should first be checking if we have something locally, then fallback on fetching remote if we have nothing in our cache. Does that make sense?
Feel free to ping me on TG if you think a chat convo can help you build context faster!

crates/core/src/rpc/accounts_scan.rs
```diff
@@ -140,3 +158,127 @@ impl AccountsScan for SurfpoolAccountsScanRpc {
...
+                    "6o5v1HC7WhBnLfRHp8mQTtCP2khdXXjhuyGyYEoy2Suy",
+                ]
+                .iter()
+                .map(|str| str.to_string())
+                .collect::<Vec<String>>(),
+            },
+        };
+
+        println!("this is the supply {:?}", res); //works
+        assert_eq!(res, expected_response);
```
lgalabru, line 282: Love the tests thanks!

crates/core/src/rpc/minimal.rs
```diff
@@ -17,6 +17,7 @@ use solana_sdk::{
     clock::{Clock, Slot},
     epoch_info::EpochInfo,
 };
+use solana_version::Version;
```
lgalabru, line 20: Could you please run
```bash
cargo fmt --alll
```

I think the CI is being angry at this 😂

adpthegreat: I ran cargo fmt without the -all, will do that

crates/core/src/rpc/accounts_scan.rs
adpthegreat: Yeah i just sent a message, found you on that solana devex telegram gc

adpthegreat:
Also @lgalabru i want to hear your thoughts on the get_epoch_schedule and the comment i made

lgalabru:
The RPC layer went through some pretty serious refactoring, which kind of nuked this PR :(. 
I'll go ahead and close it, let's address one endpoint after the other, it will make the review easier 🙏

adpthegreat:
Alright let me get the other one done

PR #53, opened 2025-04-03
feat(core): add custom rpc module with `svm_setAccount` method

MicaiahReid:


crates/core/src/rpc/custom.rs
```diff
@@ -0,0 +1,164 @@
...
+            account.data = data.clone();
+        }
+    }
+}
+
+#[rpc]
+pub trait CustomRpc {
+    type Metadata;
+
+    #[rpc(meta, name = "surfpool_setAccount")]
```
MicaiahReid: @lgalabru thoughts on `surfpool_setAccount` vs `svm_setAccount`

lgalabru (approved):

PR #54, opened 2025-04-03
fix(simnet): fetch all accounts before processing instructions

MicaiahReid:


lgalabru (approved):
Thanks @MicaiahReid, ship it!

PR #58, opened 2025-04-11
chore: cleanup dependencies

Aursen:


2501babe:
driveby comment, we are deleting `solana-inline-spl` in the near future and i noticed this project is its sole reverse dependency, so im happy to see it is being removed in this pr :grin:

if you ever need this functionality in the future, it is superceded by `spl-generic-token` which is superior to it and intended for public consumption

lgalabru:
Merging in a temp branch - will look into this PR locally, thanks so much @Aursen!

lgalabru:
@2501babe thanks for the insight!

PR #67, opened 2025-04-23
Add --network option for easy Solana network selection

ochikov:
# Add `--network` option for easy Solana network selection

## Motivation
Currently, users must specify the full RPC URL when they want to use a different Solana network (like devnet or testnet). This requires remembering or looking up the exact URLs. Providing a simple option to select common networks makes the CLI more user-friendly.

## Changes
- Added a new `--network` parameter that accepts values: `mainnet`, `devnet`, or `testnet`
- Made `--network` and `--rpc-url` mutually exclusive to avoid confusion
- Added clear help text to explain the options and constraints
- Added constants to make the code more maintainable
- Updated internal configuration methods to handle the new parameter

## Example Usage
```bash
# Use devnet instead of typing the full URL
surfpool start --network devnet

# Error when trying to use both options
surfpool start --network devnet --rpc-url https://example.com
```

## Testing Done
- Verified the new option appears in help output correctly
- Tested that the option works as expected
- Verified that using both `--network` and `--rpc-url` produces an appropriate error
- Ran `cargo fmt` to ensure code style matches the project


lgalabru (approved):
Love it, thanks @ochikov!

PR #71, opened 2025-04-25
fix: revisit config schema, remove rpc client from state

lgalabru:


MicaiahReid (approved):
Ship it! But don't forget to rev the Cargo.toml version so we release!

PR #72, opened 2025-04-27
refactor: skinny controllers, fat models

lgalabru:


crates/core/src/surfnet/mod.rs
```diff
@@ -0,0 +1,696 @@
...
+                            let res = self
+                                .get_account(
+                                    &program_data_address,
+                                    GetAccountStrategy::ConnectionOrDefault(None),
+                                )
+                                .await?;
+                            if let Some(program_data) = res {
+                                let _ = self.inner.set_account(program_data_address, program_data);
+                            }
+                        }
```
MicaiahReid: I think this will lead to a VM error that I ran into when attempting this. We need to insert the program data account _before_ we set the executable account. If we insert `account` first, it's data is pointing to `program_data_address` as having its data, but it doesn't exist yet. SO:
```suggestion

                        if account.executable {
                            let program_data_address = get_program_data_address(pubkey);
                            let res = self
                                .get_account(
                                    &program_data_address,
                                    GetAccountStrategy::ConnectionOrDefault(None),
                                )
                                .await?;
                            if let Some(program_data) = res {
                                let _ = self.inner.set_account(program_data_address, program_data);
                            }
                        }
                        let _ = self.inner.set_account(pubkey.clone(), account.clone());
```

crates/core/src/surfnet/mod.rs
```diff
@@ -0,0 +1,696 @@
...
+    /// - `tick`: A flag indicating whether to update the transaction statuses to "Confirmed" after execution.
+    ///
+    /// # Returns
+    ///
+    /// This function does not return a value. It processes the transactions in-place, sending status updates via the provided
+    /// sender and updating internal performance and epoch information.
+    pub async fn process_transactions(
+        &mut self,
+        transactions_to_process: Vec<(VersionedTransaction, Sender<TransactionStatusEvent>, bool)>,
+        tick: bool,
```
MicaiahReid: I'm not understanding this variable name and how it relates to the functionality. Can we get a better var name here?

crates/core/src/surfnet/mod.rs
```diff
@@ -0,0 +1,696 @@
...
+                    let _ = self.simnet_events_tx.try_send(SimnetEvent::error(format!(
+                        "Transaction verification failed: {}",
+                        transaction.signatures[0]
+                    )));
+                    continue;
+                }
+            }
+
+            // find accounts that are needed for this transaction but are missing from the local
+            // svm cached, fetch them from the RPC, and insert them locally
```
MicaiahReid:
```suggestion
            // svm cache, fetch them from the RPC, and insert them locally
```

crates/core/src/rpc/svm_tricks.rs
```diff
@@ -332,71 +317,44 @@ impl SvmTricksRpc for SurfpoolSvmTricksRpc {
...
+                        Account {
+                            lamports: 0,
+                            owner: system_program::id(),
+                            executable: false,
+                            rent_epoch: 0,
+                            data: vec![],
+                        },
+                    );
+                }
+            }
```
MicaiahReid: In this else case, we're no longer using the account fields the user sent in the request. Here are the cases we need to handle:
 - if the user sent _all_ account fields, we can just write to the SVM the full account state they sent
 - if the user sent some partial set of fields:
   - check if the account is in the local svm
     - if yes, apply the account fields sent by the user onto the local account state, then update the account in the svm with this new account
     - if not, fetch the account from the remote
        - if found at the remote, apply the account fields sent by the user onto the remote account we fetched, then update the account in the svm with this new account
        - if not, create a default account, but apply the account fields they sent on top of our defaults

crates/core/src/rpc/full.rs
```diff
@@ -1382,62 +1375,29 @@ impl Full for SurfpoolFullRpc {
...
-                        err: tx.transaction.meta.map(|m| m.err).flatten(),
-                        confirmation_status: Some(
-                            solana_transaction_status::TransactionConfirmationStatus::Confirmed,
-                        ),
-                    });
-                responses.insert(*i, response);
+            let svm_reader = svm_locker.read().await;
+            let mut responses = Vec::with_capacity(signatures.len());
+            for signature in signatures.into_iter() {
+                let res = svm_reader.get_transaction(&signature, None).await.unwrap();
```
MicaiahReid: Not lovin this unwrap

crates/core/src/rpc/minimal.rs
```diff
@@ -586,42 +588,28 @@ impl Minimal for SurfpoolMinimalRpc {
...
+            let res = svm_writer
+                .get_account_mut(
+                    &pubkey,
+                    GetAccountStrategy::LocalThenConnectionOrDefault(None),
+                )
+                .await
+                .unwrap();
+            let balance = match res {
+                Some(account) => account.lamports,
+                None => 0,
```
MicaiahReid, line 606: I think this should return an `AccountNotFound` error, not 0

crates/core/src/rpc/mod.rs
```diff
@@ -30,80 +30,102 @@ pub struct SurfpoolRpc;
...
+
 trait State {
-    fn get_state<'a>(&'a self) -> Result<RwLockReadGuard<'a, GlobalState>, RpcCustomError>;
-    fn get_state_mut<'a>(&'a self) -> Result<RwLockWriteGuard<'a, GlobalState>, RpcCustomError>;
+    fn get_svm_locker<'a>(&'a self) -> Result<Arc<RwLock<SurfnetSvm>>, RpcCustomError>;
+    fn with_svm_reader<T, F>(&self, reader: F) -> Result<T, RpcCustomError>
+    where
+        F: Fn(&SurfnetSvm) -> T + Send + Sync,
+        T: Send + 'static;
+    fn with_svm_writer<T, F>(&self, writer: F) -> Result<T, RpcCustomError>
```
MicaiahReid: Loving these trait methods, super cool

crates/core/src/surfnet/mod.rs
```diff
@@ -0,0 +1,696 @@
...
+    pub perf_samples: VecDeque<RpcPerfSample>,
+    pub transactions_processed: u64,
+    pub connection: SurfnetDataConnection,
+    pub latest_epoch_info: EpochInfo,
+    pub simnet_events_tx: Sender<SimnetEvent>,
+    pub geyser_events_tx: Sender<GeyserEvent>,
+}
+
+pub enum SurfnetDataConnection {
+    Local,
```
MicaiahReid: Once I started to grok the changes, I think I found this enum variant name to be misleading. Do you think `Uninitialized` would be more accurate?

crates/core/src/surfnet/mod.rs
```diff
@@ -0,0 +1,696 @@
...
+        strategy: GetAccountStrategy,
+    ) -> Result<Option<Account>, Box<dyn std::error::Error>> {
+        let (result, factory) = match strategy {
+            GetAccountStrategy::LocalOrDefault(factory) => {
+                (self.inner.get_account(pubkey), factory)
+            }
+            GetAccountStrategy::ConnectionOrDefault(factory) => {
+                let client = self.expected_rpc_client();
+                let account = client.get_account(&pubkey).await?;
+                let _ = self.inner.set_account(pubkey.clone(), account.clone());
```
MicaiahReid: Should we also be get/setting the program data address here too if executable?

crates/core/src/surfnet/mod.rs
```diff
@@ -0,0 +1,696 @@
...
+
+// #[cfg(not(clippy))]
+// const SUBGRAPH_PLUGIN_BYTES: &[u8] =
+//     include_bytes!("../../../../target/release/libsurfpool_subgraph.dylib");
+
+pub type AccountFactory = Box<dyn Fn(&SurfnetSvm) -> Account + Send + Sync>;
+
+pub enum GetAccountStrategy {
+    LocalOrDefault(Option<AccountFactory>),
+    ConnectionOrDefault(Option<AccountFactory>),
```
MicaiahReid: I'm not actually understanding the existence of this variant. Why would we ever want to get an account from the remote and ignore the local?

MicaiahReid (changes requested):

crates/core/src/surfnet/mod.rs
lgalabru: I'll go with `Offline` and added a few guard conditions in the code

crates/core/src/surfnet/mod.rs
lgalabru: Addressed with latest

crates/core/src/rpc/minimal.rs
lgalabru: Are you sure?

crates/core/src/surfnet/mod.rs
lgalabru: we currently have one use case (executable account), I think it'll be useful for cases where you always want the latest data.

crates/core/src/rpc/minimal.rs
MicaiahReid: Oooo, perhaps not. The solana cli returns an error:
```sh
 ~ solana account zbBjhHwuqyKMmz8ber5oUtJJ3ZV4B6ePmANfGyKzVGX --url devnet
Error: AccountNotFound: pubkey=zbBjhHwuqyKMmz8ber5oUtJJ3ZV4B6ePmANfGyKzVGX
```

but the `getBalance` rpc method when I hit devnet returns 0. Good call pushing back 🙏

crates/core/src/surfnet/mod.rs
MicaiahReid: I think I disagree with this take, both on the executable account and the definition of "latest data".

If we have the account locally, then _it_ is the latest data. We've interacted with the account locally, which is implied by the fact that we've pulled it, so I don't think we'd want to overwrite with the remote.

I think we'd only use `ConnectionOrDefault` in cases where it's convenient - we can prevent a read from the svm because we're _assuming_ the account doesn't already exist locally. 

For example, if we had to get an account from the remote and it is executable, we _assume_ that it's derived program data address does not live locally, so we're safe to fetch instead.

But what if the user has used `svm_setAccount` to write directly to the program data address? We've now written over the account that they explicitly set because we assume it's empty locally.

I think this could lead to some hard to find bugs - I'm wary of keeping this around.

lgalabru: works for me!

PR #75, opened 2025-04-28
Refactor/rpc patch 3

MicaiahReid:


crates/core/src/rpc/accounts_scan.rs
```diff
@@ -478,7 +478,10 @@ impl AccountsScan for SurfpoolAccountsScanRpc {
         meta: Self::Metadata,
         _config: Option<RpcSupplyConfig>,
     ) -> BoxFuture<Result<RpcResponse<RpcSupply>>> {
-        let svm_locker = meta.get_svm_locker().unwrap();
+        let svm_locker = match meta.get_svm_locker() {
+            Ok(locker) => locker,
+            Err(e) => return e.into(),
+        };
```
lgalabru, line 484: how about
```
meta.get_svm_locker() .map_err(Into::into)?
```
?

MicaiahReid: I wish :/

These functions don't return a result, they return a `BoxFuture<Result<...>>`, so the `?` operator doesn't work.

lgalabru (approved):

PR #77, opened 2025-04-29
fix(core): blockhash handling

MicaiahReid:
- create new blockhash with every slot
 - disable litesvm blockhash checks
 - manually check for valid blockhashes before tx execution/sim


lgalabru (approved):
LGTM thanks @MicaiahReid !

PR #82, opened 2025-05-02
fix: update api url

lgalabru:


MicaiahReid (approved):
Quick comment - do you want to upgrade txtx-cloud first to increase the timeout for the workspace query?

If so, you'd need to release the cloud crate + update the dep here before releasing surfpool

PR #85, opened 2025-05-04
feat:added fetching address lookup accounts implementation

adpthegreat:
Fixing tests (WIP)

### DIFF
--- a/crates/core/src/surfnet/mod.rs
+++ b/crates/core/src/surfnet/mod.rs
@@ -699,8 +699,23 @@ impl SurfnetSvm {
         // svm cache, fetch them from the RPC, and insert them locally
         let accounts = match &transaction.message {
             VersionedMessage::Legacy(message) => message.account_keys.clone(),
-            VersionedMessage::V0(message) => message.account_keys.clone(),
+            VersionedMessage::V0(message) => { 
+                let alts = message.address_table_lookups.clone();
+                let mut acc_keys = message.account_keys.clone();
+                  if !alts.is_empty() {
+                    let mut alt_pubkeys = 
+                    alts
+                   .iter()
+                   .map(|msg| msg.account_key)
+                   .collect::<Vec<_>>();
+                   acc_keys.append(&mut alt_pubkeys);
+                   acc_keys
+                  } else {
+                    acc_keys
+                  }
+            },
         };
+        
         let _ = self
             .get_multiple_accounts_mut(
                 &accounts,
--- a/crates/core/src/tests/integration.rs
+++ b/crates/core/src/tests/integration.rs
@@ -6,7 +6,13 @@ use jsonrpc_core::{
 use jsonrpc_core_client::transports::http;
 use solana_hash::Hash;
 use solana_keypair::Keypair;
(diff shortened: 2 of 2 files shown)


crates/core/src/surfnet/mod.rs
```diff
@@ -699,8 +699,23 @@ impl SurfnetSvm {
...
+                    let mut alt_pubkeys = 
+                    alts
+                   .iter()
+                   .map(|msg| msg.account_key)
+                   .collect::<Vec<_>>();
+                   acc_keys.append(&mut alt_pubkeys);
+                   acc_keys
+                  } else {
+                    acc_keys
+                  }
```
MicaiahReid: I think something's off with the indentation here, could that be fixed?

Also, I think there's no need for `if !alts.is_empty()`, you can just map over the empty array and we'll have nothing to append if empty.

Maybe:
```rust
let mut alt_account_keys = message
    .address_table_lookups
    .iter()
    .map(|msg| msg.account_key.clone())
    .collect::<Vec<_>>();

let mut account_keys = message.account_keys.clone();
account_keys.append(&mut alt_account_keys);
account_keys
```

Would have the same result with a bit less logic?

crates/core/src/tests/integration.rs
```diff
@@ -264,3 +270,239 @@ async fn test_simnet_some_sol_transfers() {
         "Some transfers failed"
     );
 }
+
+
+
+#[tokio::test]
+async fn test_add_alt_fetching() {
+    let mock_tx = VersionedTransaction {
```
MicaiahReid: This is awesome! Could you add in a comment or something indicating what this transaction or doing, or how the `instructions`/`account_keys`/`address_lookup_table` was derived, cause this is quite the intimidating struct 😰

MicaiahReid (changes requested):
This is awesome work, thank you @adpthegreat!!

Could you also take a look at this line and see if we can include the address table lookups: https://github.com/txtx/surfpool/blob/0888995b9a73e0d5d12320c7ed183279ebe936e2/crates/core/src/types.rs#L91

I think it should be similar to what you've already done, but perhaps keeping the alt accounts separated from the message accounts in this part: https://github.com/txtx/surfpool/blob/0888995b9a73e0d5d12320c7ed183279ebe936e2/crates/core/src/types.rs#L58-L79

adpthegreat:
@MicaiahReid 
 - so i've added the ALTs to the types 
 - removed the if statement from the account fetching part, then ran cargo fmt, 
 - but for the tests, it can't work rn because the assertion i'm writing - (to check if all the pubkeys from the tx when fetched locally do not return None)  uses surfnet_svm and its moved, so do i impl clone? want to get your feedback on everything i've done so far

lgalabru:
@MicaiahReid can we merge this PR?

MicaiahReid:
@lgalabru, not yet, I'd like to see the test working, but I'm also having trouble thinking of how to accomplish this. I can investigate tomorrow

MicaiahReid:
Great work @adpthegreat! I'll think on the tests and see what we can do, but we don't want to hold back the functionality any longer.

PR #87, opened 2025-05-05
ci: build for windows

MicaiahReid:


crates/cli/src/cli/mod.rs
```diff
@@ -36,13 +36,16 @@ pub const DEFAULT_SVM_GQL_URL: &str = "https://svm-cloud.gql.txtx.run/v1/graphql
 pub const DEFAULT_SVM_CLOUD_API_URL: &str = "https://svm-cloud-api.txtx.run/v1/surfnets";
 pub const DEFAULT_RUNBOOK: &str = "deployment";
 pub const DEFAULT_AIRDROP_AMOUNT: &str = "10000000000000";
-pub const DEFAULT_AIRDROPPED_KEYPAIR_PATH: &str = "~/.config/solana/id.json";
+#[cfg(not(windows))]
+pub const DEFAULT_SOLANA_KEYPAIR_PATH: &str = "~/.config/solana/id.json";
+#[cfg(windows)]
+pub const DEFAULT_SOLANA_KEYPAIR_PATH: &str = "~\\.config\\solana\\id.json";
```
lgalabru: feels like a slice would be better here?
```rust
["~", ".config", "solana", "id.json"]
```

MicaiahReid: The reason I kept it like this is that the value will be used as a default, but also be displayed in the CLI documentation generated by clap. So it'll look like a proper path rather than an array in the docs

lgalabru (approved):
Let's go!

PR #88, opened 2025-05-05
fix: select all programs by default

lgalabru:


crates/cli/src/cli/simnet/mod.rs
```diff
@@ -265,12 +265,18 @@ fn log_events(
...
                         if deployment_completed {
                             info!(
                                 ctx.expect_logger(),
-                                "Transaction processed {}", _meta.signature
+                                "Transaction processed {}", meta.signature
                             );
+                            for log in meta.logs {
+                                info!(
+                                    ctx.expect_logger(),
+                                    "Transaction logs {}: {}", meta.signature, log
```
MicaiahReid, line 277: I'm not sold on this format, do we want to put the signature in each line of the logs? This will get quite redundant, right?

crates/cli/src/scaffold/mod.rs
```diff
@@ -92,10 +92,14 @@ pub fn scaffold_iac_layout(
         hint_style: Style::new().cyan(),
         ..ColorfulTheme::default()
     };
-
     let selection = MultiSelect::with_theme(&theme)
-        .with_prompt("Programs to deploy:")
-        .items(&programs.iter().map(|p| p.name.as_str()).collect::<Vec<_>>())
+        .with_prompt("Programs to deploy (all by default):")
```
MicaiahReid:
```suggestion
        .with_prompt("Select the programs to deploy (all by default):")
```
I think part of the problem with the initial approach is we weren't making it clear that this is an action, maybe this wording change will help too

crates/cli/src/tui/simnet.rs
```diff
@@ -528,13 +531,13 @@ fn render_events(f: &mut Frame, app: &mut App, area: Rect) {
...
-    let symbol = ['⠈', '⠘', '⠸', '⠼', '⠾', '⠿', '⠷', '⠧', '⠇', '⠃', '⠁', '⠀'];
-    let cursor = symbol[app.slot() % 12];
+    let symbol = ["⢎ ", "⠎⠁", "⠊⠑", "⠈⠱", " ⡱", "⢀⡰", "⢄⡠", "⢆⡀"];
+    let cursor = symbol[app.slot() % symbol.len()];
     let title = Block::new()
         .padding(Padding::symmetric(4, 4))
         .borders(Borders::NONE)
         .title(Line::from(format!(
-            "{} Processing incoming transactions",
+            "{} Transaction processor unit running",
```
MicaiahReid: This new wording is quite awkward to me. I definitely prefer the previous

crates/cli/src/cli/simnet/mod.rs
MicaiahReid: Also, this is a LOT of logs on startup due to program deployment

MicaiahReid: Should this be hidden behind the debug flag?

MicaiahReid (changes requested):

crates/cli/src/cli/simnet/mod.rs
lgalabru: hmm this is a good point. On this specific code path, we would not be displaying the logs of the initial deployment (the logs are behind the `if deployment_completed` gate), but could be an issue in the TUI.

lgalabru: regarding the format, I think having the transaction signature is important in non-TUI mode. in my mind this mode is mostly for server side usage when you want to be able to grep or something.

crates/cli/src/tui/simnet.rs
lgalabru: haha I LOVE the canonical term Transaction Processing Unit, so geeky 😂

crates/cli/src/cli/simnet/mod.rs
lgalabru: I added a gate to the TUI mode

crates/cli/src/tui/simnet.rs
lgalabru:
```suggestion
            "{} Processing incoming transactions",
```

PR #90, opened 2025-05-06
fix: :bug: make building a little better

anthonyra:
## What changed?
- Updated version from 0.2.2 to 0.2.3 in Cargo.toml and Cargo.lock for all surfpool packages
- Cleaned up commented-out dependencies in Cargo.toml
- Added patch section (commented out) for local development
- Added conditional compilation for action_item_events_tx based on the "supervisor_ui" feature flag

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace.package]
-version = "0.2.2"
+version = "0.2.3"
 edition = "2021"
 description = "Surfpool is the best place to train before surfing Solana."
 license = "Apache-2.0"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace.package]
-version = "0.2.2"
+version = "0.2.3"
```
lgalabru: Can we please revert this bump? This would trigger a new release!

MicaiahReid:
Nice, thanks @anthonyra!

We'll hold off on merging until we need a new release, or you can revert that Cargo.toml bump - it will cause a release on merge

MicaiahReid:
Ha, looks like @lgalabru beat me to it!

Cargo.toml
anthonyra: I only added that bump due to adding the cfg flag for the unused code for the `supervisor-ui`

anthonyra:
It might not even be needed to merge right away and could wait until an official release

Cargo.toml
MicaiahReid:
```suggestion
version = "0.3.0"
```

PR #91, opened 2025-05-06
fix: cloud patches

lgalabru:
Should pull latest txtx-cloud update

### DIFF
--- a/crates/cli/src/cloud/mod.rs
+++ b/crates/cli/src/cloud/mod.rs
@@ -3,7 +3,9 @@ use std::fmt;
 use clap::{builder::PossibleValue, Parser, ValueEnum};
 use dialoguer::{console::Style, theme::ColorfulTheme, Input, Select};
 use surfpool_types::{BlockProductionMode, CreateNetworkRequest, CreateNetworkResponse};
-use txtx_cloud::{auth::AuthConfig, workspace::get_user_workspaces, LoginCommand};
+use txtx_cloud::{
+    auth::AuthConfig, login::pat_login, workspace::fetch_svm_workspaces, LoginCommand,
+};
 use txtx_gql::kit::{reqwest, uuid::Uuid};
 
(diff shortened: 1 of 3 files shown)


crates/cli/src/cloud/mod.rs
```diff
@@ -123,10 +125,31 @@ impl CloudStartCommand {
             ..ColorfulTheme::default()
         };
 
-        let workspaces = get_user_workspaces(&auth_config.access_token, svm_gql_url)
+        let mut workspaces = fetch_svm_workspaces(&auth_config.access_token, svm_gql_url)
             .await
             .map_err(|e| format!("failed to get available workspaces: {}", e))?;
 
+        if workspaces.is_empty() {
```
MicaiahReid, line 132: Do you think we should make this block part of the get user workspaces function?

Or at least making hard refresh a function somewhere in the cloud crate?

Also, why should this make any difference? We've just made a request with a refreshed access token, why would logging in again help?

lgalabru: I'm not sure, there seems to be a cache being evicted backend side when regenerating a new access token.

lgalabru: > Do you think we should make this block part of the get user workspaces function?

I'd love to get the cloud flow fixed asap, can we refactor later? we already took a bunch of shortcuts anyway.

MicaiahReid: So are we making this request just because something isn't working backend?

lgalabru: Hasura side.

MicaiahReid: So is this a bandaid? I'm cool with merging a bandaid, but maybe we can get a comment and an issue in linear saying that?

lgalabru: I created an entry in Linear!

PR #98, opened 2025-05-10
feat(rpc): add `surfnet_cloneProgramAccount` cheatcode

MicaiahReid:


crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -452,4 +465,103 @@ impl SvmTricksRpc for SurfnetCheatcodesRpc {
...
+    /// A program account contains a pointer to a program data account, which is a PDA derived from the program ID.
+    /// So, when cloning a program account, we need to clone the program data account as well.
+    ///
+    /// This method will:
+    ///  1. Get the program account for the source program ID.
+    ///  2. Get the program data account for the source program ID.
+    ///  3. Calculate the program data address for the destination program ID.
+    ///  4. Set the destination program account's data to point to the calculated destination program address.
+    ///  5. Copy the source program data account to the destination program data account.
+    fn clone_program_account(
```
lgalabru, line 474: Could we conform to the doc convention we have in the other RPC methods? 🙏

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -452,4 +465,103 @@ impl SvmTricksRpc for SurfnetCheatcodesRpc {
...
+
+            // create a new program account that has the `program_data` field set to the
+            // destination program data address
+            let mut new_program_account = source_program_account;
+            new_program_account.data = bincode::serialize(&UpgradeableLoaderState::Program {
+                programdata_address: destination_program_data_address,
+            })
+            .map_err(|e| {
+                Error::invalid_params(format!("Failed to serialize program data: {}", e))
+            })?;
```
lgalabru: would it make sense to introduce a new method in our `SurfnetSvm`, doing this things? Turbo charging that struct and keeping the RPC layer as skinny as possible could end up being useful + seems easier to unit test.

lgalabru:
thanks @MicaiahReid! is this addressing https://github.com/txtx/surfpool/issues/83, or is this a different cheat code?

crates/core/src/rpc/surfnet_cheatcodes.rs
MicaiahReid: This is a doc comment on the implementation, not on the trait, so these are our internal comments rather than the public RPC docs.

We don't have docs on the cheatcode RPCs yet, you cool if we follow on in another PR?

crates/core/src/rpc/surfnet_cheatcodes.rs
MicaiahReid: Does [this commit](a6cb2453a723d1de4966289fa6a060f349e697f8) do what you expected?

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -498,65 +493,9 @@ impl SvmTricksRpc for SurfnetCheatcodesRpc {
...
-
-            svm_writer.set_account(
-                &destination_program_data_address,
-                source_program_data_account,
-            )?;
-
-            svm_writer.set_account(&destination_program_id, new_program_account)?;
+            svm_writer
+                .clone_program_account(&source_program_id, &destination_program_id)
+                .await?;
```
lgalabru, line 498: Looking sharp! thanks 🙏

PR #99, opened 2025-05-10
Feat/various cleanups

MicaiahReid:


crates/cli/src/cli/simnet/mod.rs
```diff
@@ -216,10 +216,7 @@ fn log_events(
             0 => match oper.recv(&simnet_events_rx) {
                 Ok(event) => match event {
                     SimnetEvent::AccountUpdate(_dt, account) => {
-                        info!(
-                            ctx.expect_logger(),
-                            "Account {} retrieved from Mainnet", account
-                        );
+                        info!(ctx.expect_logger(), "Account {} updated", account);
```
MicaiahReid, line 219: @lgalabru curious to get your thoughts.

Maybe we can do better than just "account updated", but `from Mainnet` is misleading because:
 1. the remote may not be mainnet and, more importantly
 2. we can update accounts _outside_ a fetch from remote.

For example, if a cheat code directly writes to an account, we'd want to see this logging, but the data didn't necessarily come from the remote.

(I noticed because this confused me when working on SRS)

lgalabru (approved):
LGTM!

PR #105, opened 2025-05-13
get_token_account_balance

thewuhxyz:
`get_token_account_balance` rpc method


MicaiahReid (approved):
Awesome work, thanks @thewuhxyz

MicaiahReid (changes requested):
Actually @thewuhxyz, I see the test that isn't passing.

The actual logic of the test looks good, and it's some stuff that's lacking in our test framework that's causing this failure.

For now, can you just bypass the test and leave it as-is? We'll re-enable once we improve the test framework.

thewuhxyz (commented):
Tests is ignored!

MicaiahReid (approved):

PR #106, opened 2025-05-13
feat: draft support for blocks

lgalabru:


crates/core/src/rpc/minimal.rs
```diff
@@ -690,7 +690,7 @@ impl Minimal for SurfpoolMinimalRpc {
         let version = solana_version::Version::default();
 
         Ok(SurfpoolRpcVersionInfo {
-            surfpool_version: format!("surfpool_v{}", SURFPOOL_VERSION),
+            surfnet_version: format!("{}", SURFPOOL_VERSION),
```
MicaiahReid, line 693: Hmm, I think of this as surfpool the tool vs surfnet, an instance of the network. So isn't `surfpool_version` what we want?

Regardless, this has downstream implications in txtx - our cheatcode IaC is using this RPC method. If we want this change, can you get a corresponding change there as well?

crates/core/src/surfnet/mod.rs
```diff
@@ -327,6 +365,10 @@ impl SurfnetSvm {
         self.inner.set_sysvar(
             &solana_sdk::sysvar::recent_blockhashes::RecentBlockhashes::from_iter(entries),
         );
+        BlockIdentifier::new(
+            self.chain_tip.index + 1,
+            latest_entry.blockhash.to_string().as_str(),
```
MicaiahReid, line 370: Would storing a `Hash` for our block identifier hash prevent unnecessary conversions? I think `Hash` implements Copy as well, so it's pretty handy

MicaiahReid (changes requested):

crates/core/src/rpc/minimal.rs
lgalabru: Tracking `surfnet` here seems more adequate. I agree that as it stands today, `surfnet` =  `surfpool` from a version point of view, but this should probably be different in the future, and could help us track breaking changes vs non breaking changes. As in, this version should not be bumped if we're pathing something in the CLI, or in subgraphs, but should be updated when we're tweaking a given RPC. 
This would also help us align local surfnet vs hosted surfnet (not running surfpool directly).

crates/core/src/surfnet/mod.rs
lgalabru: At some point we'll need to render this value as a string. If we convert a hash into a string here, we do this operation once. If we keep a hash, we'll recompute the representation every time it's being rendered?

crates/core/src/rpc/minimal.rs
MicaiahReid: Okay, that's fine - can you update the downstream as well? Otherwise, there will be breakages in txtx's `setup_surfnet` action

lgalabru: https://github.com/txtx/txtx/pull/290

MicaiahReid (approved):
Looks good!

PR #108, opened 2025-05-14
feat:added alt table fetching entries logic

adpthegreat:
- added logic and tests to fetch alt entries via the readonly and writable indexes
- added custom error types for it
- fixed helpers.rs error


lgalabru (approved):
Thank you so much for jumping on this bug @adpthegreat!
Great work on this 🦾

adpthegreat:
thank you

PR #109, opened 2025-05-15
feat: add estimation compute units to the send_transaction function

BretasArthur1:


lgalabru (approved):
Looking great @BretasArthur1, thanks so much! 
I'm so excited by this new cheatcode and the avenue of features we'll be deriving from your work 🦾

PR #113, opened 2025-05-16
feat(rpc): implement tagged profile event handling and update estimation results

BretasArthur1:


crates/cli/src/tui/simnet.rs
```diff
@@ -300,6 +300,7 @@ fn run_app<B: Backend>(terminal: &mut Terminal<B>, mut app: App) -> io::Result<(
                             SimnetEvent::Shutdown => {
                                 break;
                             }
+                            &SimnetEvent::TaggedProfile { .. } => todo!(),
```
lgalabru: I think this `todo!()` codepath would crash the TUI if it's being executed. Could we have a comment instead?

BretasArthur1: Ok, sure, i didn't know abt it

lgalabru:
Thanks so much for these additions @BretasArthur1! Excited by the next steps 📊🔥

BretasArthur1:
let's goo, glad to contribute 🚀

PR #116, opened 2025-05-20
chore: rm build warnings

nkysg:


lgalabru:
Thanks so much for helping us keeping the codebase nice to work with @nkysg!
It looks like rustfmt is not happy with the latest changes, do you think you could fix this?
Could you also rebase your branch? @Aursen just added some additional guardrails :)

nkysg:
@lgalabru update it. PTAL.

lgalabru:
thanks @nkysg!

PR #121, opened 2025-05-23
feat(rpc): Implement getProgramAccounts RPC method

BretasArthur1:
This PR adds support for the getProgramAccounts RPC method, allowing clients to query all accounts owned by a specific program ID with filtering capabilities.

Key Features:
- ✅ Full support for getProgramAccounts RPC endpoint
- ✅ Account filtering by data size and memcmp
- ✅ Configurable account encoding (Base64, JsonParsed)
- ✅ Context and slot validation
- ✅ Data slicing support
- ✅ Proper error handling

Technical Details:
- Uses an account registry system to efficiently track program-owned accounts
- Implements RpcFilterType support for DataSize and Memcmp filters
- Follows Solana RPC specification for response format
- Maintains proper account encoding using encode_ui_account

Example Usage:
```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "getProgramAccounts",
  "params": [
    "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA",
    {
      "filters": [
        { "dataSize": 165 },
        {
          "memcmp": {
            "offset": 0,
            "bytes": "3N5kaPhfUGuTQZPQ3mnDZZGkUZ97rS1NVSC94QkgUzKN"
          }
        }
      ],
      "encoding": "jsonParsed"
    }
  ]
}
```

### DIFF
--- a/crates/core/src/rpc/accounts_data.rs
+++ b/crates/core/src/rpc/accounts_data.rs
@@ -443,7 +443,7 @@ impl AccountsData for SurfpoolAccountsDataRpc {
         _meta: Self::Metadata,
         _block: Slot,
     ) -> Result<RpcBlockCommitment<BlockCommitmentArray>> {
-        not_implemented_err()
+        not_implemented_err("get_block_commitment")
     }
 
     // SPL Token-specific RPC endpoints
@@ -535,7 +535,7 @@ impl AccountsData for SurfpoolAccountsDataRpc {
         _mint_str: String,
         _commitment: Option<CommitmentConfig>,
     ) -> Result<RpcResponse<UiTokenAmount>> {
-        not_implemented_err()
+        not_implemented_err("get_token_supply")
     }
 }
 
--- a/crates/core/src/rpc/accounts_scan.rs
+++ b/crates/core/src/rpc/accounts_scan.rs
@@ -551,7 +551,7 @@ impl AccountsScan for SurfpoolAccountsScanRpc {
         _meta: Self::Metadata,
         _config: Option<RpcLargestAccountsConfig>,
     ) -> BoxFuture<Result<RpcResponse<Vec<RpcAccountBalance>>>> {
-        not_implemented_err_async()
+        not_implemented_err_async("get_largest_accounts")
     }
 
     fn get_supply(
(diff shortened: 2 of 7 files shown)


MicaiahReid:
This is awesome, I love the way you code @BretasArthur1 

I'll do a more in-depth review later, but I'm wondering if this should be merged after #112. They are bound to have some conflicts, but #112 will be fixing a lot of issues with overlapping read/write lock fetches.

---

I want to think through the `account_registry` model you've set up - I've been thinking we'll need something like this for the `account_scan` RPC methods, but I want to be sure we've gotten this right and that it will work for the other `account_scan` RPCs.

---

Lastly, I think there's a piece missing here:
 1. Accounts that have been set are now added to the account registry
 2. Users fetching accounts owned by a specific program will get all accounts owned by that program in the registry

However, if the account hasn't been set locally but is owned by the specified program ID on mainnet, it won't be included in the result.

I think we will have to do the following:
 1. Run a `getProgramAccounts` to mainnet
 2. Compare results from that request with what's in our `accounts_registry` to see if any are missing from our mainnet result, and if things have been changed locally compared to the mainnet result. For example, if an account A is in both the mainnet result and local registry with different owners, we'll need to use the local registry's owner. However, if account A is not available locally, we'll need to be sure we've gotten it from mainnet.

I think this will be a common pattern for all of the `accounts_scan` RPCs.

---
So, in summary, I think we should hold off on this just a bit until we merge #112 and look holistically at the `account_scan` RPC to make sure this solution will work for all of them. Thanks for getting this conversation started!!

BretasArthur1:
First, thx @MicaiahReid about the compliment abt my code! i'm trying to introduce the best code possible in the codebase. And about `account_registry`,  when I was first introduced to the issue I first search in the LiteSVM api for a method to iterate to all the accounts passed so we can retrive the information.

From my research of the LiteSVM source and documentation, there's no public method to iterate over all accounts. The available methods are:

- get_account(&pubkey) - get a specific account
- set_account(pubkey, account) - set an account


Alternative Implementation that I think was, instead of scanning all accounts, we could:

- Keep track of program-owned accounts in our SurfnetSvm wrapper
- Maintain an index/registry of accounts by program ID
- Return that subset for getProgramAccounts



---

but taking a look into #112 I and what you said about  `account_registry`, in some cases it wouldn't support every method from account_scan just the ones with the same behavior. So let's hold it for now and then implement a more generic approach

PR #124, opened 2025-05-24
get_first_available_block

thewuhxyz:


crates/core/src/rpc/full.rs
```diff
@@ -1704,8 +1704,17 @@ impl Full for SurfpoolFullRpc {
...
-    fn get_first_available_block(&self, _meta: Self::Metadata) -> BoxFuture<Result<Slot>> {
-        Box::pin(async move { Ok(1) })
+    fn get_first_available_block(&self, meta: Self::Metadata) -> BoxFuture<Result<Slot>> {
+        let svm_locker = match meta.get_svm_locker() {
+            Ok(s) => s,
+            Err(e) => return e.into(),
+        };
+
+        Box::pin(async move {
+            let svm_reader = svm_locker.write().await;
```
lgalabru: Should be lock read, not lock write.

crates/core/src/rpc/full.rs
```diff
@@ -1704,8 +1704,17 @@ impl Full for SurfpoolFullRpc {
...
+        let svm_locker = match meta.get_svm_locker() {
+            Ok(s) => s,
+            Err(e) => return e.into(),
+        };
+
+        Box::pin(async move {
+            let svm_reader = svm_locker.read().await;
+            let earliest_slot = svm_reader.blocks.keys().min().copied();
+            Ok(earliest_slot.unwrap_or_default())
+        })
```
MicaiahReid: I wonder if this function could just return `Result<Slot>` rather than `BoxFuture<...>`? @lgalabru, what's the driver for if this is async or sync?

Also, @thewuhxyz, can this be implemented with the `with_svm_reader` as you did in #125? That's a much cleaner interface here, especially if we remove the `BoxFuture` result

MicaiahReid (changes requested):

MicaiahReid (approved):
Looks good!

PR #126, opened 2025-05-25
WIP: getBlockCommitment and getSupply implementation: untested

Ozodimgba:
is BlockCommimentCache populated on Surfpool, that will be a better way to get the commitment...either way I simulated it...would like to hear opinions @lgalabru 

the second question is in the get_supply();
```
// get bank to access capitalization (total supply)
// in agave: bank.capitalization() returns total SOL supply, simulate in surfpool? or get from remote rpc
```

### DIFF
--- a/.cursor/mcp.json
+++ b/.cursor/mcp.json
@@ -1,0 +1,8 @@
+{
+    "mcpServers": {
+      "surfpool": {
+        "command": "./target/release/surfpool",
+        "args": ["mcp"]
+      }
+    }
+  }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -775,6 +775,72 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
 
 [[package]]
+name = "axum"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "021e862c184ae977658b36c4500f7feac3221ca5da43e3f25bd04ab6c79a29b5"
+dependencies = [
+ "axum-core",
+ "axum-macros",
+ "bytes 1.10.1",
+ "form_urlencoded",
+ "futures-util",
+ "http 1.2.0",
+ "http-body 1.0.1",
+ "http-body-util",
+ "hyper 1.6.0",
+ "hyper-util",
+ "itoa",
+ "matchit",
+ "memchr",
+ "mime",
+ "percent-encoding 2.3.1",
+ "pin-project-lite",
+ "rustversion",
+ "serde",
+ "serde_json",
+ "serde_path_to_error",
+ "serde_urlencoded",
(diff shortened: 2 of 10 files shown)


MicaiahReid:
We don't currently have a strong concept of commitment levels for blocks. We do have something close for transactions.

 - A transaction is mark `processed` as soon as it's received and executed.
 - It's marked `confirmed` at every slot interval when we make a "block"
 - It's marked `finalized` every [FINALIZATION_SLOT_THRESHOLD](https://github.com/txtx/surfpool/blob/4d8359267a97ec7b2318be38dbb084f82264ff4d/crates/core/src/surfnet/mod.rs#L55) slots. (Currently this is set to every 31 blocks)

I think I like the approach you're using of just having some sensible defaults and basing this off the age of the block being requested as compared to the latest absolute slot.

Whenever the slot age reaches `FINALIZATION_SLOT_THRESHOLD` slots old, it should be 100% staked, and we can have every slot reaching up to that age slowly become more staked, similar to what you have.

Additionally, I believe we should be checking mainnet if the slot queried is before what's in our block store.

So that would mean:
 - first check the requested slot and see if it's less than the smallest slot we have in our block store
   - if so, forward the RPC request to mainnet and return that response, since we won't have data on those older slots
   - if not, implement the algorithm discussed above

---

Side note for @lgalabru: 
I do wonder if we should instead return some clearly "nonsense" data, since we don't actually have a concept of commitment, and I find it doubtful that users should be using surfpool for such consensus related RPCs.

MicaiahReid:
For the problem of calculating the total supply, here are my thoughts. I could be making some incorrect assumptions here, so feel free to push back @Ozodimgba or @lgalabru.

The supply results are composed of:
 1. Total supply - this value shouldn't be changed by surfpool, as we don't have any inflation mechanism to mint new SOL
 2. Non Circulating supply - this value _could_ be changed by surfpool, if one of the non-circulating-supply accounts is deposited into
 3. Circulating supply - just `total_supply - non_circulating_supply`
 4. Non Circulating Accounts - this value shouldn't be be changed by surfpool, I think

So I could see a solution where we: query the remote `getSupply` on startup, and store the total supply + non-circulating accounts on startup.

Whenever the `getSupply` RPC is queried, we check the balance of each of those non-circulating accounts, first checking local (if available) then falling back to remote. We use that to sum up the non circulating supply.

---

However, perhaps all of this is moot. Our `surfnet_setAccount` cheatcode allows setting an accounts lamports, completely breaking the supply. Perhaps we could increment/decrement the total supply accordingly when using this cheatcode, but I don't know if this is a level of detail we want to maintain in surfpool. I don't know the upside for the end user, and I think there's likely a sizeable cost to maintain such specificity.

So we could just return some hard-coded data here, or forward the request to the remote, or something like that.

Ozodimgba:
@MicaiahReid thanks for the breakdown and direction. I'm going to implement the extra check to on main-net if slot queried is > what is stored in blockstore.

On the ```get_supply``` side yeah I think i'm gonna go with this, not going to overthink it.

PR #130, opened 2025-05-28
feat: draft set-token-account for mcp

lgalabru:


BretasArthur1:
image


 implement set-token-account functionality with JSON-RPC integration

- Updated `set_token_account.rs` to handle token balance setting for both SOL and SPL tokens.
- Introduced JSON-RPC request and response structures for better error handling and response parsing.
- Added support for generating new wallets and funding existing ones.
- Updated dependencies in `Cargo.toml` for `reqwest`, `bs58`, and `spl-token`.
- Adjusted workspace dependencies in `crates/mcp/Cargo.toml` for better modularity.

crates/mcp/src/surfpool/set_token_account.rs
```diff
@@ -0,0 +1,317 @@
...
+    wallet_address_opt: Option<String>,
+    token_identifier_str: String, // Can be "SOL", a mint address, or a known symbol
+    token_amount_opt: Option<u64>,
+) -> SetTokenAccountResponse {
+    let client = Client::new();
+    let rpc_url = surfnet_address;
+
+    // Determine the owner's public key. If no wallet address is provided, generate a new keypair.
+    // The keypair itself is stored if generated, so the secret key can be returned.
+    let owner_keypair: Option<Keypair> = if wallet_address_opt.is_none() {
```
lgalabru: Use `match` when exhausting all variants :)

crates/mcp/src/surfpool/set_token_account.rs
```diff
@@ -0,0 +1,317 @@
...
+            Err(_) => {
+                // if not a direct mint address, try looking up as a symbol in VERIFIED_TOKENS_BY_SYMBOL (case-insensitive).
+                match VERIFIED_TOKENS_BY_SYMBOL.get(&token_identifier_str.to_uppercase()) {
+                    Some(token_info) => {
+                        actual_mint_address_str = token_info.address.clone();
+                        actual_token_symbol_opt = Some(token_info.symbol.clone());
+                    }
+                    None => {
+                        return SetTokenAccountResponse::error(format!(
+                            "Invalid token identifier: '{}'. Not SOL, a valid mint address, or a known symbol.",
```
lgalabru: I think this error should be something along the lines:
The token symbol provided '<symbol>' is not a verified token. Please provide the token address instead.

crates/mcp/src/surfpool/set_token_account.rs
BretasArthur1: sure, I was trying to simplify because was a simple if statement but match it's more robust

crates/mcp/src/surfpool/set_token_account.rs
```diff
@@ -0,0 +1,317 @@
...
+            }
+        }
+    }
+
+    // Construct the appropriate JSON-RPC request payload based on whether it's SOL or an SPL token.
+    let request_payload = if is_sol {
+        // For SOL, use the `surfnet_setAccount` RPC method.
+        // Parameters: (pubkey: String, update: AccountUpdate)
+        let update_params = AccountUpdateParams {
+            lamports: Some(amount_to_set),
```
lgalabru: I think we need to multiply this `amount_to_set` by the `decimals` field from `TokenInfo` struct.

BretasArthur1: something like this?


```rust
let lamports_to_set = match &actual_token_symbol_opt {
            Some(token_symbol) => match VERIFIED_TOKENS_BY_SYMBOL.get(&token_symbol.to_uppercase()) {
                Some(token_info) => amount_to_set
                    .checked_mul(10u64.pow(token_info.decimals as u32))
                    .unwrap_or(amount_to_set),
                None => amount_to_set,
            },
            None => amount_to_set,
        };
   ```

Comment (lgalabru) [reply crates/mcp/src/surfpool/set_token_account.rs]: Perfect thanks!

Comment (lgalabru) [reply crates/mcp/src/surfpool/set_token_account.rs]: Note: we'll need to do this for the non sol token as well.

File: crates/mcp/src/surfpool/start_surfnet.rs [hunk 4]
```diff
@@ -26,24 +25,49 @@ impl StartSurfnetResponse {
...
+        Ok::<(), String>(())
+    });
+
+    match handle {
+        Ok(_) => loop {
+            match simnet_events_rx.recv() {
+                Ok(SimnetEvent::Aborted(error)) => {
+                    return StartSurfnetResponse::error(error);
+                }
+                Ok(SimnetEvent::Connected(_)) | Ok(SimnetEvent::Ready) => {
```
lgalabru: `Connected` is not enough, can we just look at `Ready`?

crates/mcp/src/surfpool/start_surfnet.rs
```diff
@@ -26,24 +25,49 @@ impl StartSurfnetResponse {
...
+                    return StartSurfnetResponse::success(
```
lgalabru: Instead of returning, I think we should `break` from the loop, and detach another thread that will keep `simnet_events_rx.recv()`, and send notifications to the MCP - this can be part of a subsequent PR.

crates/mcp/src/surfpool/start_surfnet.rs
```diff
@@ -26,24 +25,49 @@ impl StartSurfnetResponse {
...
+                        "Surfnet started successfully at http://127.0.0.1:8899".to_string(),
```
lgalabru: Can we read the port from the config?

crates/mcp/src/surfpool/start_surfnet.rs
```diff
@@ -26,24 +25,49 @@ impl StartSurfnetResponse {
...
-            }
-            Ok::<(), String>(())
-        })
-        .map_err(|e| StartSurfnetResponse::error(format!("Failed to execute surfnet start: {}", e)))?;
+    let (simnet_commands_tx, simnet_commands_rx) = crossbeam_channel::unbounded();
+    let (subgraph_commands_tx, _subgraph_commands_rx) = crossbeam_channel::unbounded();
+
+    let simnet_events_tx = surfnet_svm.simnet_events_tx.clone();
+
+    let config = SurfpoolConfig::default();
```
lgalabru: Can we override the default port with a value coming from the arguments?

BretasArthur1: We gone to do that overriding if the user passes the argument?

BretasArthur1:
To test: Open the branch build the surfpool binary `cargo build --release` and check in the if the MCP it's avaible in your IDE or CLI. If it's like in the image. Just use the Chat in the agent mode and prompt something like: start a surfnet with X accounts, and X amount of SOL, USDC or any SPL token avaible, wait for the LLM process the request and call the MCP tools, accept the calls and see the magic happening!


Screenshot 2025-06-03 at 12 13 49

crates/mcp/src/surfpool/mod.rs
```diff
@@ -0,0 +1,147 @@
...
+    pub fn set_token_account(
+        &self,
+        #[tool(param)]
+        #[schemars(
+            description = "the RPC url of the local surfnet instance where the operation will take place. Must reference a currently running localnet."
+        )]
+        surfnet_address: String,
+        #[tool(param)]
+        #[schemars(
+            description = "the public key of the wallet to fund. If omitted, a new wallet will be generated and returned"
```
MicaiahReid:
```suggestion
            description = "The public key of the wallet to fund. If omitted, a new wallet will be generated and returned"
```

crates/mcp/src/surfpool/mod.rs
```diff
@@ -0,0 +1,147 @@
...
+    /// Sets the token balance for any account in your local Solana network. Supports SOL and any SPL token.
+    /// This method is exposed as a tool and can be invoked remotely.
+    #[tool(
+        description = "Sets the token balance for any account in your local Solana network. Supports SOL and any SPL token. If succesful, all the data should be displayed in the chat, if not, share the error message"
+    )]
+    pub fn set_token_account(
+        &self,
+        #[tool(param)]
+        #[schemars(
+            description = "the RPC url of the local surfnet instance where the operation will take place. Must reference a currently running localnet."
```
MicaiahReid:
```suggestion
            description = "The RPC url of the local surfnet instance where the operation will take place. Must reference a currently running localnet."
```

crates/mcp/src/surfpool/mod.rs
```diff
@@ -0,0 +1,147 @@
...
+        )]
+        surfnet_address: String,
+        #[tool(param)]
+        #[schemars(
+            description = "the public key of the wallet to fund. If omitted, a new wallet will be generated and returned"
+        )]
+        wallet_address: Option<String>,
+        #[tool(param)]
+        #[schemars(
+            description = "the token to set the balance for. Can be a symbol (e.g., SOL, USDC) or a full base58-encoded mint address."
```
MicaiahReid:
```suggestion
            description = "The token to set the balance for. Can be a symbol (e.g., SOL, USDC) or a full base58-encoded mint address."
```

crates/mcp/src/surfpool/mod.rs
```diff
@@ -0,0 +1,147 @@
...
+        )]
+        token: String,
+        #[tool(param)]
+        #[schemars(
+            description = "the token amount to assign to the wallet. Defaults to 100_000 if not provided"
```
MicaiahReid:
```suggestion
            description = "The token amount to assign to the wallet. Defaults to 100_000 if not provided"
```

crates/mcp/src/surfpool/mod.rs
```diff
@@ -0,0 +1,147 @@
...
+    pub fn start_surfnet_with_token_accounts(
+        &self,
+        #[tool(param)]
+        #[schemars(
+            description = "Optional. The public key of the wallet to fund. If omitted, set_token_account logic will generate a new wallet."
+        )]
+        wallet_address: Option<String>,
+        #[tool(param)]
+        #[schemars(
+            description = "the token to set the balance for. Can be a symbol (e.g., SOL, USDC) or a full base58-encoded mint address. Defaults to SOL if not provided."
```
MicaiahReid:
```suggestion
            description = "The token to set the balance for. Can be a symbol (e.g., SOL, USDC) or a full base58-encoded mint address. Defaults to SOL if not provided."
```

crates/mcp/src/surfpool/mod.rs
```diff
@@ -0,0 +1,147 @@
...
+        )]
+        token: String,
+        #[tool(param)]
+        #[schemars(
+            description = "the token amount to assign to the wallet. Defaults to 100_000 if not provided"
```
MicaiahReid:
```suggestion
            description = "The token amount to assign to the wallet. Defaults to 100_000 if not provided"
```

crates/mcp/src/surfpool/set_token_account.rs
```diff
@@ -0,0 +1,340 @@
...
+}
+
+#[derive(Serialize, Debug, Clone)]
+pub struct SetTokenAccountResponse {
+    success: Option<AccountUpdated>,
+    error: Option<String>,
+}
+
+#[derive(Serialize, Debug, Clone)]
+pub struct AccountUpdated {
```
MicaiahReid, line 56: Can we include the `owner: String` pubkey here?

crates/mcp/src/surfpool/set_token_account.rs
```diff
@@ -0,0 +1,340 @@
...
+    account: SeededAccount,
+    token_address: String,         // Mint address or "SOL"
+    token_symbol: Option<String>,  // e.g., "SOL", "USDC", or mint symbol
+    token_account_address: String, // Wallet address for SOL, ATA for SPL
```
MicaiahReid: I think `associated_token_address` may be more standard here

crates/mcp/src/surfpool/set_token_account.rs
```diff
@@ -0,0 +1,340 @@
...
+    lamports: Option<u64>,
+}
+
+#[derive(Serialize)]
+struct SetTokenAccountParams {
+    owner: String,
+    mint: String,
+    update: TokenAccountUpdateParams,
+    token_program: Option<String>,
+}
```
MicaiahReid: I believe this is unused and can be removed

crates/mcp/src/surfpool/set_token_account.rs
```diff
@@ -0,0 +1,340 @@
...
+
+#[derive(Serialize)]
+struct SetAccountParams {
+    pubkey: String,
+    update: AccountUpdateParams,
+}
+
+#[derive(Serialize)]
+struct AccountUpdateParams {
```
MicaiahReid: I wonder if we should be importing this struct from the actual RPC implementation, rather than recreating.

Maybe we don't want the mcp server to have all of surfpool-core imported, but perhaps just `surfpool-types`?

Then this wouldn't need to be recreated here

crates/mcp/src/surfpool/set_token_account.rs
```diff
@@ -0,0 +1,340 @@
...
+            }
+        }
+    }
+
+    // Construct the appropriate JSON-RPC request payload based on whether it's SOL or an SPL token.
+    let request_payload = if is_sol {
+        // For SOL, use the `surfnet_setAccount` RPC method.
+        // Parameters: (pubkey: String, update: AccountUpdate)
+        let lamports_to_set = match &actual_token_symbol_opt {
+            Some(token_symbol) => match VERIFIED_TOKENS_BY_SYMBOL.get(&token_symbol.to_uppercase())
```
MicaiahReid: If we know it's SOL, do we need to look up to get the decimals, or should we just have a const?

crates/mcp/src/surfpool/set_token_account.rs
```diff
@@ -0,0 +1,340 @@
...
+                                get_associated_token_address_with_program_id(
+                                    &owner_pubkey,
+                                    &mint_pubkey,
+                                    &spl_token::id(),
+                                )
+                                .to_string()
+                            };
+
+                            // Construct the successful response details.
+                            let account_updated = AccountUpdated {
```
MicaiahReid, line 310: In the case that the token was `SOL`, we've only done a `surfnet_setAccount` RPC.

However, we're always returning this full `AccountUpdated`, which lists the token symbol. So, the logs look like the user had `So11111111111111111111111111111111111111112` transferred, when in reality they just increased their lamports

crates/mcp/src/surfpool/set_token_account.rs
```diff
@@ -0,0 +1,340 @@
...
+        Self {
+            success: None,
+            error: Some(message),
+        }
+    }
+}
+
+const DEFAULT_TOKEN_AMOUNT: u64 = 100_000_000_000; // 100 SOL or 100 of a token with 9 decimals
+const SOL_SYMBOL: &str = "SOL";
+const SOL_MINT_PLACEHOLDER: &str = "So11111111111111111111111111111111111111112"; // Official SOL mint placeholder
```
MicaiahReid: When the user says "SOL", you're doing a `surfnet_setAccount`, which just increases the lamports. However, you're marking the mint as `So11111111111111111111111111111111111111112`, which is the token for wrapped sol

crates/mcp/src/surfpool/mod.rs
```diff
@@ -0,0 +1,147 @@
...
+        #[schemars(
+            description = "the token to set the balance for. Can be a symbol (e.g., SOL, USDC) or a full base58-encoded mint address."
+        )]
+        token: String,
+        #[tool(param)]
+        #[schemars(
+            description = "the token amount to assign to the wallet. Defaults to 100_000 if not provided"
+        )]
+        token_amount: Option<u64>,
+    ) -> Json<SetTokenAccountResponse> {
```
MicaiahReid, line 108: We will need another optional parameter for the token program ID. We should use the SPL Token library as default, but users may need to specify Token2022.

crates/types/src/types.rs
```diff
@@ -315,6 +315,7 @@ pub struct SimnetConfig {
     pub block_production_mode: BlockProductionMode,
     pub airdrop_addresses: Vec<Pubkey>,
     pub airdrop_token_amount: u64,
+    pub expiry: Option<u64>, // 15 minutes
```
```suggestion
    pub expiry: Option<u64>,
```

crates/types/src/verified_tokens.json
MicaiahReid, line 1: Can we format this doc?

crates/core/src/runloops/mod.rs
```diff
@@ -125,6 +130,13 @@ pub async fn start_block_production_runloop(
                         if block_production_mode.eq(&BlockProductionMode::Clock) {
                             do_produce_block = true;
                         }
+
+                        if let Some(expiry_ms) = expiry_duration_ms {
+                            let svm = svm_locker.0.read().await;
```
MicaiahReid: I don't love that we're getting the read lock every 400ms here.

Maybe my logic is wrong here, but at a _maximum_, we need to check this value every `expiry_duration_ms`. There's a chance we will _extend_ this, if the user has made some updates, but it will never be shorter than the original `expiry_duration_ms`.

If that makes sense to you, just before this loop we should declare an
```rust
let mut check_expiry_at = Utc::now().timestamp_millis() + expiry_ms
```

Then, in the clock event loop, we rarely need to check if expired:
```rust
if let Some(expiry_ms) = expiry_duration_ms {
    let now = Utc::now().timestamp_millis() as u64;
    if now >= check_expiry_at {
        let svm = svm_locker.0.read().await;
        if svm.updated_at + expiry_ms <  now {
            let _ = simnet_commands_tx.send(SimnetCommand::Terminate(None));
        }
        else {
            check_expiry_at = svm.updated_at + expiry_ms;
        }
    }
}
```

crates/core/src/surfnet/svm.rs
```diff
@@ -110,6 +111,7 @@ impl SurfnetSvm {
                 transactions_queued_for_finalization: VecDeque::new(),
                 signature_subscriptions: HashMap::new(),
                 tagged_profiling_results: HashMap::new(),
+                updated_at: Utc::now().timestamp_millis() as u64,
```
MicaiahReid, line 118: Currently, we're updating this value when we set an account or send a transaction. So technically, not all RPC requests will reset this value.

If I do a bunch of `getBalance` RPC requests to my network, my network will still shut down, even though I'm using it.

However, we can only set this value on functions that are `&mut self`, and we don't want all of these functions to have to write to the `SurfnetSvm`.

For now, I'm okay with this, with one exception. Can you set the `updated_at` on every fn in this file that is `&mut self`?


In the future, I think we can add some middleware to the RPC which will reset this value at every request. And in the future I think this `updated_at` field won't live on the `SurfnetSvm`

MicaiahReid (changes requested):

crates/mcp/src/surfpool/set_token_account.rs
BretasArthur1: so something like this:
```rust
 token_address: if is_sol { SOL_SYMBOL.to_string() } else { actual_mint_address_str.clone() }
```

crates/mcp/src/surfpool/set_token_account.rs
```diff
@@ -0,0 +1,344 @@
...
+}
+
+#[derive(Deserialize, Debug)]
+struct JsonRpcError {
+    code: i64,
+    message: String,
+}
+
+#[derive(Serialize)]
+struct TokenAccountUpdateParams {
```
MicaiahReid, line 36: This type is also available in the cheatcodes file - can we move it out to types and import it as well?

crates/mcp/src/surfpool/set_token_account.rs
```diff
@@ -0,0 +1,344 @@
...
+#[derive(Serialize, Debug, Clone)]
+pub struct SetTokenAccountsResponse {
+    pub success: Option<Vec<AccountUpdated>>,
+    pub error: Option<String>,
+}
+
+#[derive(Serialize, Debug, Clone)]
+pub struct AccountUpdated {
+    account: SeededAccount,
+    token_address: String,            // Mint address or "SOL"
```
MicaiahReid, line 58: Here's an example of a change I'd like to see across the whole PR:
```rust
token_address: String,            // Mint address or "SOL"
```

I have a feeling this is AI generated 😅

But if we're going to annotate a struct's fields, let's:
 1. Do it as an RS-doc comment and
 2. Make the language clear and specific
For example:
```rust
/// The mint address of the token that was updated, or the "SOL" symbol
token_address: String
```

The RS doc comments give us proper annotation with the rust analyzer

crates/mcp/src/surfpool/set_token_account.rs
MicaiahReid: I think we need to have two possible return values:
 - if we did `surfnet_setAccount`, we should return one struct
 - if we did `surfnet_setTokenAccount`, we should return this `AccountUpdated` struct

crates/mcp/src/surfpool/set_token_account.rs
```diff
@@ -0,0 +1,344 @@
...
+    jsonrpc: String,
+    id: u64,
+    method: String,
+    params: serde_json::Value,
+}
+
+#[derive(Deserialize, Debug)]
+struct JsonRpcResponse<T> {
+    // jsonrpc: String,
+    // id: u64,
```
MicaiahReid, line 23: remove if not used

crates/mcp/src/surfpool/set_token_account.rs
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    pub fn error(message: String) -> Self {
+        Self {
+            success: None,
+            error: Some(message),
+        }
+    }
+}
+const SOL_DECIMALS: u32 = 9; // Define SOL decimals constant
```
MicaiahReid, line 130: This comment seems unnecessary - the name of the variable tells us what it is

crates/mcp/src/surfpool/set_token_account.rs
```diff
@@ -0,0 +1,344 @@
...
+const DEFAULT_TOKEN_AMOUNT: u64 = 100_000_000_000; // 100 SOL or 100 of a token with 9 decimals
+const SOL_SYMBOL: &str = "SOL";
```
MicaiahReid, line 132: Let's move constants to be under the imports but above any code.

crates/mcp/src/surfpool/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+        }
+        Json(res)
+    }
+
+    /// Sets the token balance for any account in your local Solana network. Supports SOL and any SPL token.
+    /// This method is exposed as a tool and can be invoked remotely.
+    #[tool(
+        description = "Sets the token balance for any account in your local Solana network. Supports SOL and any SPL token. If succesful, all the data should be displayed in the chat, if not, share the error message"
+    )]
+    pub fn set_token_account(
```
MicaiahReid, line 81: Currently, this only allows setting one account, and is pretty limiting.

If I first prompt the MCP to "Start a surfnet", then later tell it "create a new account and give it 100 SOL + 100 USDC", this endpoint isn't particularly efficient - I'll have to approve it running twice.

Additionally, it'll likely have the same problem we had with `start_surfnet_and_set_token_accounts` - it'll end up creating two new wallets, one with 100 SOL and one with 100 USDC.

---

**The solution**

This endpoint should be `set_token_accounts`, and should mirror what we're already doing for `start_surfnet_with_token_accounts`.

It should receive a `token_params_with_owner: Vec<CreateTokenAccountForOwnerParams>`, so a user can say "create 3 accounts, each with 100 SOL & 100 USDC", and the user will only have to approve one request.

crates/mcp/src/surfpool/mod.rs
MicaiahReid, line 1: On this doc, can you please revisit all of the `schemars` description annotations? We want all fields to be documented in a consistent way.

For example, the `start_surfnet` action has "Start a new local Solana network, also called surfnet or localnet.", but the `start_surfnet_and_set_token_accounts` action has `Starts a new local Solana network,`. They do the same thing (partially) but have vastly different docs.

crates/mcp/src/surfpool/start_surfnet.rs
```diff
@@ -0,0 +1,151 @@
...
+    let (surfnet_svm, simnet_events_rx, geyser_events_rx) = SurfnetSvm::new();
+
+    let (simnet_commands_tx, simnet_commands_rx) = crossbeam_channel::unbounded();
+    let (subgraph_commands_tx, _subgraph_commands_rx) = crossbeam_channel::unbounded();
+
+    let simnet_events_tx = surfnet_svm.simnet_events_tx.clone();
+
+    let mut config = SurfpoolConfig::default();
+    let rpc_port = 8899 + 10000 * surfnet_id;
+    let ws_port = 8890 + 10000 * surfnet_id;
```
MicaiahReid, line 43: Using only the `surfnet_id` to choose the port is insufficient. If the user has started a surfnet _without_ the MCP, then starts another with the MCP, they will have a port collision.

Add a helper function to ping `*8899` ports to find the next one that's open. Call this function _before_ entering this `run` function. Then in our `pub surfnets: Arc<RwLock<Vec<u16>>>` field, we should instead store a mapping of id to port.

So whenever a surfnet is going to be started we:
 1. Get the next ID
 2. Get the next open port
 3. Call `start_surfnet::run`, passing in the ports it should run on
 4. Store the id<>port mapping in the `surfnets` lock

MicaiahReid (commented):
One last comment. Currently we have some inconsistent language on how we refer to tokens/token program id/ATAs, etc. Across the whole PR, let's be consistent:
 - `token_mint` - the mint address of a token
 - `token_program_id` - the token program id (TOKEN/TOKEN2022/etc)
 - `owner` - the address of the owner of a token
 - `associated_token_address` - the associated token address for the `owner`/`token_mint`/`token_program_id` combo
 - `token_symbol` - USDC/JUP/etc
 - `token_amount` - the smallest value of a token (so lamports for SOL)

Let's be consistent with these, and avoid a field being more than one thing. For example, `token: String` being a mint address _or_ a symbol is not ideal, we want a field for each and to be consistent.

PR #131, opened 2025-05-29
feat(core): implement `getSupply` RPC endpoint

Ozodimgba:


crates/core/src/rpc/accounts_scan.rs
```diff
@@ -485,13 +486,64 @@ impl AccountsScan for SurfpoolAccountsScanRpc {
...
+                            "9i93DT71Ac82dSrqGfE5Y8Ko81copbsjcQbjnc4h337a".to_string(),
+                            "FdRWzEd6i4oQuzxfmZ4fLB7Bd8v9XnuD6AARg7jXotJn".to_string(),
+                            "AFya6aUWmzT14CaXPiFTz3xUWV4uoznv5thwKymYNB4y".to_string(),
+                            "9JoChhEVmxCtMq6XQ7BfeQZuiZaz4Sq5Ym97R9B43uy3".to_string(),
+                            "ESWL7qERyLmzbG1kgmoZb3RAF66Gz42WAThdvdqWmZAq".to_string(),
+                            "J3d7MDV2DBN5PMmY4MTvfeaDKb1vnbd1Wvb2KjXQTwun".to_string(),
+                            "E46i1Cs7rBWjmF2gYm8tRKYcoe9RJ71PWstVM7Xdh4UE".to_string(),
+                            "6Vkp8GRrgxWr81ph98Njtwuq6DybJbL7qFfv1RVF9EkY".to_string(),
+                        ]
+                    },
```
lgalabru, line 546: Could we move these values to static constants / `lazy_static` slice?

MicaiahReid: I don't think we want these values here in the first place... @lgalabru let's discuss then give clear direction to @Ozodimgba

Ozodimgba: hey guys @Ozodimgba and @lgalabru, bumping this

Ozodimgba: hey guys @lgalabru and @MicaiahReid haven't received much direction on this so will just go on to remove the hardcode values...will just fetch from remote and fallback with error

MicaiahReid: @Ozodimgba, please hold for now, we will let you know when we have more direction

MicaiahReid: Hey @Ozodimgba, sorry for the wait! Here's our plan for this endpoint.

1. The `total`, `circulating`, `non_circulating`, and `non_circulating_accounts` fields should all be stored on the `SurfnetSvm`, and they should have empty defaults (0 for numbers, empty vec for non circulating accounts)
2. This RPC method should just read directly from the value on the `SurfnetSvm`.
3. These values should be configurable via a cheatcode RPC:
   1. Implement `surfnet_setSupply` RPC endpoint, which takes in an optional total supply, optional circulating supply, optional non circulating supply, and optional vec of non circulating accounts, and updates the `SurvnetSvm` with whatever is passed in.

Let me know if you need any further clarification!

PR #132, opened 2025-05-29
feat(core): implement `getBlockHeight` RPC method

Ozodimgba:


crates/core/src/rpc/minimal.rs
```diff
@@ -668,12 +668,40 @@ impl Minimal for SurfpoolMinimalRpc {
...
+
+        // commitment adjustments to determine which slot's data to use
+        let _target_slot = match config.commitment.unwrap_or_default().commitment {
+            CommitmentLevel::Processed => current_slot,
+            CommitmentLevel::Confirmed => current_slot.saturating_sub(1),
+            CommitmentLevel::Finalized => current_slot.saturating_sub(FINALIZATION_SLOT_THRESHOLD),
+        };
+
+
+        // return the current block height since we don't have historical block height data in the current implementation
```
MicaiahReid: The `SurfnetSvm` struct has `pub blocks: HashMap<Slot, BlockHeader>`, which can be used to get the block height associated with a slot

crates/core/src/rpc/minimal.rs
```diff
@@ -668,12 +668,40 @@ impl Minimal for SurfpoolMinimalRpc {
     fn get_block_height(
         &self,
         meta: Self::Metadata,
-        _config: Option<RpcContextConfig>,
+        config: Option<RpcContextConfig>,
     ) -> Result<u64> {
+        let config = config.unwrap_or_default();
+
+        //get the current slot and check min_context_slot constraint
```
MicaiahReid:
```suggestion
```
A good practice for comments that I've been taught by my seniors is:
 - DO write comments if it explains _why_ we're doing something, and it's not immediately clear
 - Don't write comments if it's explaining _what_ you're doing - this should be clear from the code

crates/core/src/rpc/minimal.rs
```diff
@@ -668,12 +668,40 @@ impl Minimal for SurfpoolMinimalRpc {
...
+        let current_slot = meta
+            .with_svm_reader(|svm_reader| svm_reader.get_latest_absolute_slot())
+            .map_err(Into::<jsonrpc_core::Error>::into)?;
+
+
+        // check minimum context slot whn specified -> against current slot, not adjusted slot
```
MicaiahReid:
```suggestion
        // check minimum context slot when specified -> against current slot, not adjusted slot
```
Good comment!

crates/core/src/rpc/minimal.rs
```diff
@@ -668,12 +668,40 @@ impl Minimal for SurfpoolMinimalRpc {
...
+            if current_slot < min_context_slot {
+                return Err(RpcCustomError::MinContextSlotNotReached {
+                    context_slot: min_context_slot,
+                }
+                    .into());
+            }
+        }
+
+        // commitment adjustments to determine which slot's data to use
+        let _target_slot = match config.commitment.unwrap_or_default().commitment {
```
MicaiahReid: You're not currently using the target slot - I think you'll be able to if you access the `SurfnetSvm`'s `blocks` property

crates/core/src/rpc/minimal.rs
```diff
@@ -668,12 +668,40 @@ impl Minimal for SurfpoolMinimalRpc {
...
+            CommitmentLevel::Processed => current_slot,
+            CommitmentLevel::Confirmed => current_slot.saturating_sub(1),
+            CommitmentLevel::Finalized => current_slot.saturating_sub(FINALIZATION_SLOT_THRESHOLD),
+        };
+
+
+        // return the current block height since we don't have historical block height data in the current implementation
         meta.with_svm_reader(|svm_reader| svm_reader.latest_epoch_info.block_height)
```
MicaiahReid: Currently, we're calling `with_svm_reader` twice for this function.

Each call to this function is waiting to acquire a read lock - we'd like to do this as few times as possible.

Can we adjust this RPC method to only call the function once?

crates/core/src/rpc/minimal.rs
```diff
@@ -752,6 +780,100 @@ impl Minimal for SurfpoolMinimalRpc {
...
+    fn test_get_block_height_basic() {
+        let setup = TestSetup::new(SurfpoolMinimalRpc);
+        
+        let result = setup.rpc.get_block_height(Some(setup.context.clone()), None);
+        assert!(result.is_ok(), "get_block_height should not fail");
+
+        let block_height = result.unwrap();
+
+        println!("Block height returned: {:?}", block_height);
+    }
```
MicaiahReid:
```suggestion
```
This test seems unnecessary - the `test_get_block_height_matches_svm_state` is doing the same test and more

crates/core/src/rpc/minimal.rs
```diff
@@ -752,6 +780,100 @@ impl Minimal for SurfpoolMinimalRpc {
...
+        match &result {
+            Ok(height) => println!("Success: Block height = {}", height),
+            Err(e) => println!("Error: {:?}", e),
+        }
+
+        assert!(result.is_ok(), "Should succeed when min_context_slot is less than current slot");
+    }
+
+    #[test]
+    fn test_get_block_height_debug_slots() {
```
MicaiahReid: This test has no assertions, and I'm not sure from the name what the purpose of the test is

crates/core/src/rpc/minimal.rs
```diff
@@ -752,6 +780,100 @@ impl Minimal for SurfpoolMinimalRpc {
...
+
+        let block_height = result.unwrap();
+        let svm_block_height = setup.context.surfnet_svm.blocking_read().latest_epoch_info.block_height;
+
+        assert_eq!(block_height, svm_block_height,
+                   "Implementation should return the same block height as SVM state");
+    }
+
+    #[test]
+    fn test_get_block_height_with_processed_commitment() {
```
MicaiahReid: Good test!

Would be nice to include tests for `confirmed` + `finalized` commitment

crates/core/src/rpc/minimal.rs
```diff
@@ -752,6 +780,100 @@ impl Minimal for SurfpoolMinimalRpc {
...
+        let result = setup.rpc.get_block_height(Some(setup.context.clone()), config);
+        assert!(result.is_ok());
+
+        let block_height = result.unwrap();
+        let expected_height = setup.context.surfnet_svm.blocking_read().latest_epoch_info.block_height;
+        assert_eq!(block_height, expected_height);
+    }
+
+    #[test]
+    fn test_get_block_height_with_valid_min_context_slot() {
```
MicaiahReid: Good test! Can we also get one for the failure case - asserting the RPC returns an error if the min_context_slot is beyond the current slot?

MicaiahReid (changes requested):

crates/core/src/rpc/minimal.rs
Ozodimgba: Okay noted

crates/core/src/rpc/minimal.rs
Ozodimgba: yeah I can do that

crates/core/src/rpc/minimal.rs
Ozodimgba: I think that was my personal test that I didnt remove

Ozodimgba:
@MicaiahReid refactored blocks, take a look

crates/core/src/rpc/minimal.rs
```diff
@@ -754,6 +784,106 @@ impl Minimal for SurfpoolMinimalRpc {
...
+
+        // defaults to finalized, finalized should be 31 lesser than latest height
+        let expected_height = setup.rpc.get_block_height(Some(setup.context.clone()), None);
+        
+        let latest_epoch_block_height = setup.context.svm_locker.with_svm_reader(|svm_reader| {
+            svm_reader.latest_epoch_info.block_height
+        });
+        
+        assert_eq!(latest_epoch_block_height, expected_height.unwrap() + FINALIZATION_SLOT_THRESHOLD);
+    }
```
MicaiahReid:
```suggestion
```
I think you said this is your personal test that needs to be removed?

crates/core/src/rpc/minimal.rs
```diff
@@ -754,6 +784,106 @@ impl Minimal for SurfpoolMinimalRpc {
...
+        
+        let latest_epoch_block_height = setup.context.svm_locker.with_svm_reader(|svm_reader| {
+            svm_reader.latest_epoch_info.block_height
+        });
+        
+        assert_eq!(expected_height.unwrap(), latest_epoch_block_height - 1);
+    }
+
+    #[test]
+    fn test_get_block_height_with_min_context_slot() {
```
MicaiahReid, line 839: This test still isn't checking the error case, where we hit:
```rust
            if !block_exists {
                return Err(jsonrpc_core::Error::invalid_params(format!(
                    "Block not found for slot: {}", target_slot
                )));
            }
```

Can you add a test for that?

MicaiahReid (changes requested):
Looking great! Two minor changes then we're ready to merge!

Ozodimgba:
@MicaiahReid this should do it

MicaiahReid (approved):
Great work!!

MicaiahReid:
aggh, sorry, looks like CI isn't passing cause of some formatting. Should be an easy fix!

Run `cargo +nightly fmt --all -- --check` and implement the changes

Ozodimgba:
@MicaiahReid  done

MicaiahReid:
Hey @Ozodimgba there are still more issues that need to be fixed from `cargo +nightly fmt --all -- --check`

PR #133, opened 2025-05-29
feat(core): implement `getBlocks` RPC endpoint

Ozodimgba:
Pushed both of them because they have very similar implementation.


crates/core/src/rpc/full.rs
```diff
@@ -1642,22 +1642,146 @@ impl Full for SurfpoolFullRpc {
...
+        config: Option<RpcContextConfig>,
     ) -> BoxFuture<Result<Vec<Slot>>> {
-        not_implemented_err_async()
+        let svm_locker = match meta.get_svm_locker() {
+            Ok(locker) => locker,
+            Err(e) => return e.into(),
+        };
+
+        Box::pin(async move {
+            let svm_reader = svm_locker.read().await;
```
MicaiahReid: In order to minimize how often we acquire the read lock, all possible input value checks should be done _before_ getting the lock.

Here, I believe you should be able to check ahead of time if: 
 - the user provided an end slot and it is more than 500k greater than start slot
 - the user provided a "processed" commitment

crates/core/src/rpc/full.rs
```diff
@@ -1642,22 +1642,146 @@ impl Full for SurfpoolFullRpc {
...
+            let slot_threshold = match commitment {
+                solana_commitment_config::CommitmentLevel::Confirmed => current_slot.saturating_sub(1),
+                solana_commitment_config::CommitmentLevel::Finalized => {
+                    current_slot.saturating_sub(FINALIZATION_SLOT_THRESHOLD)
+                }
+                _ => current_slot,
+            };
+
+            // get all available slots from the blocks map
+            let mut available_slots: Vec<Slot> = svm_reader
```
MicaiahReid: I think we'll also need to fetch missing blocks from the remote.

If the user requests blocks 10 - 20, and the lowest block available locally to the surfnet is block 15, we should:
 - get blocks 15 - 20 locally from the surfnet
 - fetch blocks 10-14 from the remote
 - return the combined list

crates/core/src/rpc/full.rs
```diff
@@ -1642,22 +1642,146 @@ impl Full for SurfpoolFullRpc {
...
+        config: Option<RpcContextConfig>,
     ) -> BoxFuture<Result<Vec<Slot>>> {
-        not_implemented_err_async()
+        let svm_locker = match meta.get_svm_locker() {
+            Ok(locker) => locker,
+            Err(e) => return e.into(),
+        };
+
+        Box::pin(async move {
+            let svm_reader = svm_locker.read().await;
```
MicaiahReid: Same comment as above on accessing the reader and on getting blocks from the remote

crates/core/src/rpc/full.rs
```diff
@@ -2296,4 +2420,214 @@ mod tests {
                 .to_string()
         );
     }
+
+    #[tokio::test]
+    async fn test_get_blocks_with_commitment() {
```
MicaiahReid: Good test!

Ozodimgba:
@MicaiahReid please review

MicaiahReid (approved):
Very good work, thanks @Ozodimgba!

MicaiahReid (changes requested):
Ahh wait, I realized we still aren't getting remote blocks here.

If locally we have blocks 50-100, and the user requests blocks 1-100, we should get all of the local blocks, then fetch 1-50 from the remote, and combine the list.

Ozodimgba:
> Ahh wait, I realized we still aren't getting remote blocks here.
> 
> If locally we have blocks 50-100, and the user requests blocks 1-100, we should get all of the local blocks, then fetch 1-50 from the remote, and combine the list.

okay this should apply too when the user fetches 50-150 blocks too right?

MicaiahReid:
> > Ahh wait, I realized we still aren't getting remote blocks here.
> > If locally we have blocks 50-100, and the user requests blocks 1-100, we should get all of the local blocks, then fetch 1-50 from the remote, and combine the list.
> 
> okay this should apply too when the user fetches 50-150 blocks too right?

~Sorry, I'm not sure if I understand the question.~

~If the user requests some blocks that are outside the range of blocks that we have locally, we should fetch that range from the remote.~

Edited:

Oooh I think I get what you're asking! If the user requests a block that is _greater than_ the highest slot we have, should that be fetched from the remote? No.

If the user requests some blocks that are _before_ our local records - fetch from remote. Anything after our local block store's lowest slot should be retrieved only from the local block store.

Ozodimgba:
> > > Ahh wait, I realized we still aren't getting remote blocks here.
> > > If locally we have blocks 50-100, and the user requests blocks 1-100, we should get all of the local blocks, then fetch 1-50 from the remote, and combine the list.
> > 
> > 
> > okay this should apply too when the user fetches 50-150 blocks too right?
> 
> ~Sorry, I'm not sure if I understand the question.~
> 
> ~If the user requests some blocks that are outside the range of blocks that we have locally, we should fetch that range from the remote.~
> 
> Edited:
> 
> Oooh I think I get what you're asking! If the user requests a block that is _greater than_ the highest slot we have, should that be fetched from the remote? No.
> 
> If the user requests some blocks that are _before_ our local records - fetch from remote. Anything after our local block store's lowest slot should be retrieved only from the local block store.

okay noted…should they be cached too?

MicaiahReid:
> > > > Ahh wait, I realized we still aren't getting remote blocks here.
> > > > If locally we have blocks 50-100, and the user requests blocks 1-100, we should get all of the local blocks, then fetch 1-50 from the remote, and combine the list.
> > > 
> > > 
> > > okay this should apply too when the user fetches 50-150 blocks too right?
> > 
> > 
> > ~Sorry, I'm not sure if I understand the question.~
> > ~If the user requests some blocks that are outside the range of blocks that we have locally, we should fetch that range from the remote.~
> > Edited:
> > Oooh I think I get what you're asking! If the user requests a block that is _greater than_ the highest slot we have, should that be fetched from the remote? No.
> > If the user requests some blocks that are _before_ our local records - fetch from remote. Anything after our local block store's lowest slot should be retrieved only from the local block store.
> 
> okay noted…should they be cached too?

Nope! Our block store is not including remote blocks, just those that we've made since startup

crates/core/src/rpc/full.rs
```diff
@@ -1661,20 +1661,155 @@ impl Full for SurfpoolFullRpc {
...
+
+                    let local_min_slot = svm_reader.blocks.keys().min().copied();
+
+                    let local_slots: Vec<Slot> = svm_reader
+                        .blocks
+                        .keys()
+                        .filter(|&&slot| {
+                            let in_range = slot >= start_slot
+                                && slot <= effective_end_slot
+                                && slot <= committed_latest_slot;
```
MicaiahReid, line 1732:
```suggestion
                                && slot <= effective_end_slot;
```
I could be wrong, but isn't that last check redundant? Because `effective_end_slot` will always be less than or equal to `committed_latest_slot`?

PR #136, opened 2025-05-30
feat(rpc): enhance profile result structure with pre and post executi…

BretasArthur1:
…on state tracking

### DIFF
--- a/crates/core/src/rpc/surfnet_cheatcodes.rs
+++ b/crates/core/src/rpc/surfnet_cheatcodes.rs
@@ -1,4 +1,4 @@
-use std::fmt;
+use std::{collections::BTreeMap, fmt};
 
 use base64::{engine::general_purpose::STANDARD, Engine as _};
 use jsonrpc_core::{futures::future, BoxFuture, Error, Result};
@@ -16,7 +16,7 @@ use solana_sdk::{
 use spl_associated_token_account::get_associated_token_address_with_program_id;
 use spl_token::state::{Account as TokenAccount, AccountState};
 use surfpool_types::{
-    types::{ComputeUnitsEstimationResult, ProfileResult},
+    types::{ComputeUnitsEstimationResult, ProfileResult, ProfileState},
     SimnetEvent,
 };
 
@@ -599,6 +599,7 @@ impl SvmTricksRpc for SurfnetCheatcodesRpc {
                     context: RpcResponseContext::new(0),
                     value: ProfileResult {
                         compute_units: error_cu_result,
+                        state: ProfileState::new(BTreeMap::new(), BTreeMap::new()),
                     },
                 }));
             }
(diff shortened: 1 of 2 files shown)


crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -617,24 +618,89 @@ impl SvmTricksRpc for SurfnetCheatcodesRpc {
...
         Box::pin(async move {
+            let account_keys_to_profile = transaction.message.static_account_keys().to_vec();
+            // In a future step, we could extend account_keys_to_profile with an optional RPC parameter `accounts_to_profile`
+
+            let mut pre_execution_capture = BTreeMap::new();
+            for key in &account_keys_to_profile {
+                let SvmAccessContext {
+                    inner: account_result,
+                    ..
+                } = svm_locker.get_account_local(key);
```
MicaiahReid: Here' we're acquiring the read lock in a loop, meaning each iteration of this loop will await the read lock being available. Is it possible to switch this to `get_multiple_accounts_local` instead?

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -617,24 +618,89 @@ impl SvmTricksRpc for SurfnetCheatcodesRpc {
...
+                    GetAccountResult::None(_) => {
+                        pre_execution_capture.insert(*key, None);
+                    }
+                }
+            }
+
             let SvmAccessContext {
                 slot,
-                inner: estimation_result,
+                inner: estimation_result, // This is ComputeUnitsEstimationResult
```
MicaiahReid:
```suggestion
                inner: estimation_result,
```
This comment feels unnecessary, the Rust Analyzer will tell us the type 🙂

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -617,24 +618,89 @@ impl SvmTricksRpc for SurfnetCheatcodesRpc {
...
+            // TODO: This currently fetches the original account state again, as
+            // svm_locker.estimate_compute_units does not (yet) return the post-simulation state changes.
+            // The underlying LiteSVM simulate_transaction or its wrapper in SurfnetSvm
+            // needs to be modified to provide the state diffs for accurate post-execution capture.
+            let mut post_execution_capture = BTreeMap::new();
+            for key in &account_keys_to_profile {
+                let SvmAccessContext {
+                    inner: account_result,
+                    ..
+                } = svm_locker.get_account_local(key);
```
MicaiahReid: Same thing about using `get_multiple_accounts_local`

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -617,24 +618,89 @@ impl SvmTricksRpc for SurfnetCheatcodesRpc {
...
+            }
+
             let SvmAccessContext {
                 slot,
-                inner: estimation_result,
+                inner: estimation_result, // This is ComputeUnitsEstimationResult
                 ..
             } = svm_locker.estimate_compute_units(&transaction);
 
+            // TODO: This currently fetches the original account state again, as
```
MicaiahReid: I want to be sure I understand this TODO correctly - does our `estimate_compute_units` not actually update the pre/post state, so the result of fetching the accounts again won't actually be shown in the profile?

MicaiahReid (changes requested):

crates/core/src/rpc/surfnet_cheatcodes.rs
MicaiahReid: @BretasArthur1, I never saw an answer to this question. I know the TODO is gone, but I'm still curious of the answer - can we actually see a difference in the before/after state here?

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -580,9 +580,23 @@ impl SvmTricksRpc for SurfnetCheatcodesRpc {
         transaction_data_b64: String,
         tag: Option<String>,
     ) -> BoxFuture<Result<RpcResponse<ProfileResult>>> {
-        let svm_locker = match meta.get_svm_locker() {
-            Ok(locker) => locker,
-            Err(e) => return e.into(),
+        let rpc_context = match meta.get_rpc_context(CommitmentConfig::confirmed()) {
+            Ok(ctx) => ctx,
+            Err(e) => {
```
MicaiahReid: Hmm, I think we should return an error here `Err(e) => return e.into()`.

If we can't get the context we need, the node is in a bad state and the user should be informed of this.

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -596,9 +610,12 @@ impl SvmTricksRpc for SurfnetCheatcodesRpc {
                     error_message: Some(format!("Invalid base64 for transaction data: {}", e)),
                 };
                 return Box::pin(future::ok(RpcResponse {
-                    context: RpcResponseContext::new(0),
+                    context: RpcResponseContext::new(
+                        rpc_context.svm_locker.get_latest_absolute_slot(),
+                    ),
                     value: ProfileResult {
                         compute_units: error_cu_result,
```
MicaiahReid: I'm also questioning if we should be returning this result with data - why not just return an error? If you create a `SurfpoolError`, you can `return surfpool_error.into()` and it will automatically be cast to the `Box::pin(future::ok(...))` error that you need!

crates/core/src/surfnet/svm.rs
```diff
@@ -156,31 +157,34 @@ impl SurfnetSvm {
     /// A `TransactionResult` indicating success or failure.
     pub fn airdrop(&mut self, pubkey: &Pubkey, lamports: u64) -> TransactionResult {
         let res = self.inner.airdrop(pubkey, lamports);
-        if let Ok(ref tx_result) = res {
```
MicaiahReid: I'm not understanding the purpose of this change

crates/core/src/surfnet/svm.rs
```diff
@@ -290,7 +294,7 @@ impl SurfnetSvm {
         let _ = self
             .simnet_events_tx
             .send(SimnetEvent::account_update(*pubkey));
-        Ok(())
+        std::result::Result::Ok(())
```
MicaiahReid: We are doing this a lot in this file - why?

crates/core/src/surfnet/svm.rs
```diff
@@ -414,6 +422,96 @@ impl SurfnetSvm {
         }
     }
 
+    pub fn simulate_with_account_snapshots(
+        &self,
+        _remote_ctx: &Option<SurfnetRemoteClient>,
```
MicaiahReid: This is close!

However, you'll see here that we're not using the `remote_ctx` - this is how we get data from the mainnet data source.

So, if the user sends a transaction to be simulated that doesn't have all of the accounts available in the local state, the transaction will fail.

You can see down below, you're doing `svm_clone.inner.get_account` - this means you're getting just the local account.

If you do `let svm_locker_clone = SurfnetSvmLocker::new(svm_clone);`, then you should be able to properly process transactions:

`svm_locker.process_transactions(...)`

This will require you to pass in this remote context, and will pull the accounts from the remote as needed

crates/core/src/surfnet/svm.rs
```diff
@@ -443,10 +541,11 @@ impl SurfnetSvm {
         let slot = self.latest_epoch_info.slot_index;
 
         while let Some((tx, status_tx)) = self.transactions_queued_for_confirmation.pop_front() {
-            let _ = status_tx.try_send(TransactionStatusEvent::Success(
-                TransactionConfirmationStatus::Confirmed,
-            ));
-            // .map_err(Into::into)?;
+            status_tx
```
MicaiahReid: I think in the future, if we're going to refactor other parts of the code, let's keep it in a separate commit!

That way it's easy to back out the change if we don't want it. For this one, I'm a bit concerned there will be unintended consequences and would not want it on this PR. Was this change necessary to implement the RPC endpoint you're working on?

MicaiahReid (changes requested):
Getting closer!

crates/core/src/surfnet/svm.rs
MicaiahReid: Reverted

crates/core/src/surfnet/svm.rs
MicaiahReid: Reverted

lgalabru (approved):
Looking sharp @BretasArthur1 @MicaiahReid, team work makes the dream work! 🚀

PR #138, opened 2025-05-31
fix simulate_transaction alts

daog1:
The code modification is insufficient; it still throws an error when simulating the call to alt. During simulation, calling the same method to retrieve alt account data should suffice.

### DIFF
--- a/crates/core/src/rpc/full.rs
+++ b/crates/core/src/rpc/full.rs
@@ -1535,7 +1535,6 @@ impl Full for SurfpoolFullRpc {
         let (_, unsanitized_tx) =
             decode_and_deserialize::<VersionedTransaction>(data, binary_encoding).unwrap();
 
-
         let SurfnetRpcContext {
             svm_locker,
             remote_ctx,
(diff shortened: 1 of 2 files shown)


lgalabru:
Thanks @daog1 🙏
Would you be able to demonstrate this fix through a unit test? That'd be super helpful to avoid future regressions.

daog1:
> Thanks @daog1 🙏 Would you be able to demonstrate this fix through a unit test? That'd be super helpful to avoid future regressions.

This is a bit complex as it involves a call on the blockchain. The call won't succeed without accessing on-chain data, and it requires a private key address that exists on the blockchain to be successful. I tested it on my own computer, and it works fine.

I made this transaction (https://solscan.io/tx/4Y4y5KG547ctTKiQoV9wRRxP9TSCy8jpkiiZtDzihHSXSfeS6pChAXY9DZsoCMa8pZasR7LF6b9bNevh12h7gmhG), and the second instruction (scope: refreshPriceList) is the same thing.

Here is a demonstration (https://github.com/daog1/PyGenIDL001/blob/main/notebooks/scope.ipynb) of calling the refreshPriceList instruction using a scope interface generated by my in-development codama Python, connected to a local surfpool (my fixed version).

PR #141, opened 2025-06-03
Feat/rpc/get recent prioritization fees

MicaiahReid:
This endpoint was requested by the poof team

### DIFF
--- a/crates/core/src/rpc/full.rs
+++ b/crates/core/src/rpc/full.rs
@@ -1941,13 +1941,14 @@ mod tests {
         EncodedTransaction, EncodedTransactionWithStatusMeta, UiCompiledInstruction, UiMessage,
         UiRawMessage, UiTransaction,
     };
-    use surfpool_types::{SimnetCommand, TransactionConfirmationStatus};
+    use surfpool_types::{SimnetCommand, TransactionConfirmationStatus, TransactionMetadata};
     use test_case::test_case;
 
     use super::*;
     use crate::{
         surfnet::{BlockHeader, BlockIdentifier},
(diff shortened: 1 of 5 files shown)


lgalabru (approved):
LGTM!
Could we backlog the improvement we touched on? 🙏

crates/core/src/rpc/full.rs
```diff
@@ -1785,10 +1793,123 @@ impl Full for SurfpoolFullRpc {
...
+        let SurfnetRpcContext {
+            svm_locker,
+            remote_ctx,
+        } = match meta.get_rpc_context(CommitmentConfig::confirmed()) {
+            Ok(res) => res,
+            Err(e) => return e.into(),
+        };
+
+        Box::pin(async move {
+            let (blocks, transactions) = svm_locker.with_svm_reader(|svm_reader| {
```
BretasArthur1, line 1820: If we are cloning blocks and transactions every call to `get_recent_prioritization_fees` maybe if these data structures grow large, repeated cloning may impact memory usage and RPC latency under heavy load.

BretasArthur1:
This PR it's so clean! The code feels smooth to read I could predict the next lines when I was reading lol. even with my comment It lgtm!

BretasArthur1 (approved):

crates/core/src/rpc/full.rs
MicaiahReid: Agreed, good call! I think this would also be solved by @lgalabru's suggestion. I've created [this issue](https://github.com/txtx/surfpool/issues/143) and am assigning it to you!! 😈

PR #144, opened 2025-06-05
feat(core): implement `getBlocksWithLimit` RPC endpoint

Ozodimgba:


crates/core/src/rpc/full.rs
```diff
@@ -1671,12 +1671,73 @@ impl Full for SurfpoolFullRpc {
...
+                    return Err(Error::invalid_params("Limit must be greater than 0"));
+                }
+    
+                const MAX_LIMIT: usize = 500_000;
+                if limit > MAX_LIMIT {
+                    return Err(Error::invalid_params(format!(
+                        "Limit too large. Maximum limit allowed: {}",
+                        MAX_LIMIT
+                    )));
+                }
```
MicaiahReid: Let's move these checks to be before we acquire the read lock via `svm_locker.with_svm_reader`, since they only rely on the user's input and not on any svm data.

MicaiahReid (changes requested):
Good stuff! This one should be have similar to `getBlocks` in terms of how it fetches blocks from the remote!

crates/core/src/rpc/full.rs
Ozodimgba: okay will do!

Ozodimgba:
@MicaiahReid forgot to tag you for update. I've put the checks before the locks acquisition

MicaiahReid:
Nice @Ozodimgba! Still waiting on this fetching blocks from the remote.

Ozodimgba:
> Nice @Ozodimgba! Still waiting on this fetching blocks from the remote.

oh missed that

MicaiahReid (approved):
🙌

PR #146, opened 2025-06-05
feat(core): implement `getBlockProduction` RPC endpoint

Ozodimgba:
implementation of get block production following @MicaiahReid directions


MicaiahReid (approved):
Thanks @Ozodimgba!

lgalabru (approved):

PR #148, opened 2025-06-06
feat(core): implement `getTokenSupply` RPC endpoint

Ozodimgba:
@MicaiahReid and @lgalabru  please review

### DIFF
--- a/crates/core/src/rpc/accounts_data.rs
+++ b/crates/core/src/rpc/accounts_data.rs
@@ -588,7 +588,7 @@ impl AccountsData for SurfpoolAccountsDataRpc {
             {
                 return Err(SurfpoolError::invalid_account_data(
                     mint_pubkey,
-                    "Account is not owned by the SPL Token program",
+                    "Account is not a token mint account",
                     None::<String>,
                 )
                 .into());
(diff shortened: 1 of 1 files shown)


crates/core/src/rpc/accounts_data.rs
```diff
@@ -531,11 +531,82 @@ impl AccountsData for SurfpoolAccountsDataRpc {
...
+                ..
+            } = svm_locker
+                .get_account(&remote_ctx, &mint_pubkey, None)
+                .await?;
+    
+            svm_locker.write_account_update(mint_account_result.clone());
+    
+            let mint_account = mint_account_result.map_account()?;
+    
+            if mint_account.owner != spl_token::id() {
```
MicaiahReid: Hmm, what if it's owned by the token2022 program?

crates/core/src/rpc/accounts_data.rs
```diff
@@ -531,11 +531,82 @@ impl AccountsData for SurfpoolAccountsDataRpc {
...
+                            decimals: mint_data.decimals,
+                            ..Default::default()
+                        }),
+                    )
+                    .ok()
+                    .and_then(|t| match t {
+                        TokenAccountType::Mint(mint) => Some(UiTokenAmount {
+                            amount: mint.supply.clone(),
+                            decimals: mint.decimals,
+                            ui_amount: None, // let client calculate if needed?
```
MicaiahReid: We have the supply and decimals, so we can calculate for them, right?

crates/core/src/rpc/accounts_data.rs
```diff
@@ -637,4 +708,254 @@ mod tests {
             }
         );
     }
+
+    #[ignore = "requires-network"]
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_get_token_supply_with_real_mint() {
```
MicaiahReid, line 831: For this test and `test_get_token_supply_caches_remote_account`, can we instead write some dummy account data to the accounts store locally, then verify the RPC returns that data? It would be great to have a baseline version of this test runnable in CI, then we can add tests to fetch from the remote in the integration tests down the road.

MicaiahReid (changes requested):
Mostly there, good work!

crates/core/src/rpc/accounts_data.rs
Ozodimgba: oh yeah that's true

crates/core/src/rpc/accounts_data.rs
Ozodimgba: yeah we can, should I?

MicaiahReid: Yes, please!

crates/core/src/rpc/accounts_data.rs
```diff
@@ -554,23 +554,109 @@ impl AccountsData for SurfpoolAccountsDataRpc {
...
+
+            svm_locker.write_account_update(mint_account_result.clone());
+
+            let mint_account = mint_account_result.map_account()?;
+
+            if !matches!(mint_account.owner, owner if owner == spl_token::id() || owner == spl_token_2022::id())
+            {
+                return Err(SurfpoolError::invalid_account_data(
+                    mint_pubkey,
+                    "Account is not owned by the SPL Token program",
```
MicaiahReid:
```suggestion
                    "Account is not a token mint account",
```

crates/core/src/rpc/accounts_data.rs
```diff
@@ -740,4 +826,366 @@ mod tests {
...
+            .rpc
+            .get_token_supply(
+                None, // no context
+                valid_mint.to_string(),
+                Some(CommitmentConfig::confirmed()),
+            )
+            .await;
+
+        assert!(res.is_err(), "Should fail when no context provided");
+    }
```
MicaiahReid:
```suggestion
```
This test is unnecessary tbh

crates/core/src/rpc/accounts_data.rs
```diff
@@ -740,4 +826,366 @@ mod tests {
...
+            second_call_duration
+        );
+
+        println!("Local account test passed!");
+        println!("Both calls used local data");
+        println!(
+            "First call: {:?}, Second call: {:?}",
+            first_call_duration, second_call_duration
+        );
+    }
```
MicaiahReid:
```suggestion
```
I don't think this test is bringing much value, and timing things like this in CI can sometimes lead to flakiness

MicaiahReid (changes requested):

BretasArthur1 (approved):
looking good, down the road we could maybe cache the responses if it get multiple requests

PR #149, opened 2025-06-08
feat(core): implement `getHighestSnapshotSlot` RPC endpoint

Ozodimgba:
@MicaiahReid surfpool doesn't implement incremental snapshots?


Ozodimgba:
> We currently don't have any concept of snapshots, and I'm not sure if we'll need it or now. It looks like if a node doesn't have a snapshot, it should return
> 
> ```json
> {
>   "jsonrpc": "2.0",
>   "error": {
>     "code": -32008,
>     "message": "No snapshot"
>   },
>   "id": 1
> }
> ```
> 
> based off these docs: https://solana.com/docs/rpc/http/gethighestsnapshotslot
> 
> Can we go with this?

sure will do

MicaiahReid:
Ah @Ozodimgba, just needs `cargo +nightly fmt --all -- --check`

lgalabru (approved):

PR #150, opened 2025-06-08
feat(core): implement `getBlockCommitment` RPC endpoint

Ozodimgba:


crates/core/src/rpc/accounts_data.rs
```diff
@@ -440,10 +440,39 @@ impl AccountsData for SurfpoolAccountsDataRpc {
...
+            // 1. blocks that are explicitly in the blocks HashMap (confirmed blocks)
+            // 2. current and past slots that might not be in blocks HashMap yet
+            if block_exists || block <= current_slot {
+                commitment_array[30] = 100;
+                commitment_array[31] = 50;
+            }
+            
+            Ok(RpcBlockCommitment {
+                commitment: Some(commitment_array),
+                total_stake: 150,
```
MicaiahReid: Is there any significance to the 150 number we've chosen here? If we're going to hardcode something, I'd prefer 0 so it's more clearly indicating there are no commitments/stakes.

crates/core/src/rpc/accounts_data.rs
```diff
@@ -440,10 +440,39 @@ impl AccountsData for SurfpoolAccountsDataRpc {
...
+            // this handles both:
+            // 1. blocks that are explicitly in the blocks HashMap (confirmed blocks)
+            // 2. current and past slots that might not be in blocks HashMap yet
+            if block_exists || block <= current_slot {
+                commitment_array[30] = 100;
+                commitment_array[31] = 50;
+            }
+            
+            Ok(RpcBlockCommitment {
+                commitment: Some(commitment_array),
```
MicaiahReid: We are always returning `Some(...)`, but if I query an old block on the RPC, I'll see a `null` result.

crates/core/src/rpc/accounts_data.rs
```diff
@@ -637,4 +666,106 @@ mod tests {
             }
         );
     }
+
+    #[test]
+    fn test_get_block_commitment_basic() {
```
MicaiahReid: This test isn't asserting anything - can we remove if the other tests are covering its cases?

MicaiahReid (changes requested):

crates/core/src/rpc/accounts_data.rs
Ozodimgba: no significance really only that we are mocking to show that confirmed blocks should have some stake

crates/core/src/rpc/accounts_data.rs
Ozodimgba: @MicaiahReid there is blockCommitmentCache on the RPC client that holds data about the block commitment (surfpool doesn't have this cache or need it)...the thing about the cache is that old finalized blocks with lesser chance of being rollback leave the cache...returning null when you query their commitment array. Can implement null if you want but this was the idea

crates/core/src/rpc/accounts_data.rs
Ozodimgba: yeah that was just to show the return value will remove it after review and linting checks

MicaiahReid: Let's get these sorts of things and lints fixed before review to save the back and forth! 👍

crates/core/src/rpc/accounts_data.rs
MicaiahReid: I think let's zero out, since for surfpool there isn't really a concept of stake

crates/core/src/rpc/accounts_data.rs
MicaiahReid: Gotcha, then I'm okay with always returning `Some([0u64; 32])` for valid requests

MicaiahReid (approved):

PR #151, opened 2025-06-08
WIP: account indexs

Ozodimgba:


LevBeta:
The optimization idea is solid, but I'm not a big fan of introducing multiple new hash maps that store owned Pubkey values both as keys and within their entries—especially since many of these values will end up duplicated or even triplicated.

For example the 

```
pub token_accounts: HashMap<Pubkey, TokenAccount>,                 // pubkey -> parsed_token_data
pub token_accounts_by_owner: HashMap<Pubkey, Vec<Pubkey>>,         // owner -> token_account_pubkeys
pub token_accounts_by_delegate: HashMap<Pubkey, Vec<Pubkey>>,      // delegate -> token_account_pubkeys
pub token_accounts_by_mint: HashMap<Pubkey, Vec<Pubkey>>,          // mint -> token_account_pubkeys
```

Assuming `N` token accounts, we're storing roughly `N * 3` extra repeated owned Pubkey values just for indexing. This leads to unnecessary memory usage and allocation overhead.

MicaiahReid:
> The optimization idea is solid, but I'm not a big fan of introducing multiple new hash maps that store owned Pubkey values both as keys and within their entries—especially since many of these values will end up duplicated or even triplicated.
> 
> For example the
> 
> ```
> pub token_accounts: HashMap<Pubkey, TokenAccount>,                 // pubkey -> parsed_token_data
> pub token_accounts_by_owner: HashMap<Pubkey, Vec<Pubkey>>,         // owner -> token_account_pubkeys
> pub token_accounts_by_delegate: HashMap<Pubkey, Vec<Pubkey>>,      // delegate -> token_account_pubkeys
> pub token_accounts_by_mint: HashMap<Pubkey, Vec<Pubkey>>,          // mint -> token_account_pubkeys
> ```
> 
> Assuming `N` token accounts, we're storing roughly `N * 3` extra repeated owned Pubkey values just for indexing. This leads to unnecessary memory usage and allocation overhead.

Agreed on the downsides, @LevBeta, but this will work for now.

Any suggestions for future optimizations on this approach?

Ozodimgba:
@MicaiahReid 
1: added new indexs following your directions. The PR is a major dependency for the endpoints => getTokenAccountsByDelegate, getTokenAccountsByOwner as this impact their implementation.

2: if we going to be frequently checking for duplicates, we can use HashSet<Pubkey> instead of Vec<Pubkey> for the indexes. This makes duplicate checking O(1) instead of O(n). (not a blocker at all just a suggestion)

Ozodimgba:
@MicaiahReid resolved here on github...it should build sucessfully. If it doesnt will just sort it out first thing tomorrow before I commit the other endpoints on WIP

Ozodimgba:
@lgalabru need you to review this so it can get merged...I need it for the token endpoints

PR #152, opened 2025-06-09
fix(core/rpc): use `noSigVerify` in `simulateTransaction` endpoint

daog1:
Add config.sig_verify to take effect when calling simulate_transaction, and add a test case test_simulate_transaction_no_signers.

https://github.com/txtx/surfpool/issues/117

### DIFF
--- a/crates/core/src/rpc/full.rs
+++ b/crates/core/src/rpc/full.rs
@@ -1552,6 +1552,8 @@ impl Full for SurfpoolFullRpc {
         };
 
         Box::pin(async move {
+            svm_locker
+                .with_svm_writer(|svm_writer| svm_writer.inner.set_sigverify(config.sig_verify));
             let pubkeys = svm_locker
                 .get_pubkeys_from_message(&remote_ctx, &unsanitized_tx.message)
                 .await?;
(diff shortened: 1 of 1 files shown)


daog1:
> Sadly, this PR won't work for this.
> 
> When you call
> 
> ```rust
> svm_locker
>   .with_svm_writer(|svm_writer| svm_writer.inner.set_sigverify(config.sig_verify));
> ```
> 
> We are altering the internal VM behavior for this and all future transaction simulations and executions.
> 
> I think a more robust solution would mean:
> 
> 1. We permanently disable signature verification with LiteSvm (where we have `LiteSVM::new()`, we add `.with_sigverify(false)`). With this, the internal SVM will never verify signatures
> 2. On all transaction processing and simulating that we do (probably in the `locker.rs` file, we manually to verification _if_ the user requests it:
>    For example, in our current `process_instructions`, we have:
> 
> ```rust
>        // verify valid signatures on the transaction
>        {
>            if transaction
>                .verify_with_results()
>                .iter()
>                .any(|valid| !*valid)
>            {
>                return Ok(self.with_contextualized_svm_reader(|svm_reader| {
>                    svm_reader
>                        .notify_invalid_transaction(transaction.signatures[0], status_tx.clone());
>                }));
>            }
>        }
> ```
> 
> We could add a similar block to `simulate_transaction`, but where this block is only called if the user passes `config.sig_verify = true`.
> 
> Let me know if you want to implement this, @daog1, or if you'd like me to take over!

Yes, I also think changing it to global is not ideal. To achieve individual control, it would probably require modifying the litesvm code, right?

MicaiahReid:
> > Sadly, this PR won't work for this.
> > When you call
> > ```rust
> > svm_locker
> >   .with_svm_writer(|svm_writer| svm_writer.inner.set_sigverify(config.sig_verify));
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We are altering the internal VM behavior for this and all future transaction simulations and executions.
> > I think a more robust solution would mean:
> > 
> > 1. We permanently disable signature verification with LiteSvm (where we have `LiteSVM::new()`, we add `.with_sigverify(false)`). With this, the internal SVM will never verify signatures
> > 2. On all transaction processing and simulating that we do (probably in the `locker.rs` file, we manually to verification _if_ the user requests it:
> >    For example, in our current `process_instructions`, we have:
> > 
> > ```rust
> >        // verify valid signatures on the transaction
> >        {
> >            if transaction
> >                .verify_with_results()
> >                .iter()
> >                .any(|valid| !*valid)
> >            {
> >                return Ok(self.with_contextualized_svm_reader(|svm_reader| {
> >                    svm_reader
> >                        .notify_invalid_transaction(transaction.signatures[0], status_tx.clone());
> >                }));
> >            }
> >        }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We could add a similar block to `simulate_transaction`, but where this block is only called if the user passes `config.sig_verify = true`.
> > Let me know if you want to implement this, @daog1, or if you'd like me to take over!
> 
> Yes, I also think changing it to global is not ideal. To achieve individual control, it would probably require modifying the litesvm code, right?

Exactly, that's why I'm proposing disabling this permanently at the LiteSVM level, then implementing the verification checks manually on our wrapper

daog1:
> > > Sadly, this PR won't work for this.
> > > When you call
> > > ```rust
> > > svm_locker
> > >   .with_svm_writer(|svm_writer| svm_writer.inner.set_sigverify(config.sig_verify));
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > We are altering the internal VM behavior for this and all future transaction simulations and executions.
> > > I think a more robust solution would mean:
> > > 
> > > 1. We permanently disable signature verification with LiteSvm (where we have `LiteSVM::new()`, we add `.with_sigverify(false)`). With this, the internal SVM will never verify signatures
> > > 2. On all transaction processing and simulating that we do (probably in the `locker.rs` file, we manually to verification _if_ the user requests it:
> > >    For example, in our current `process_instructions`, we have:
> > > 
> > > ```rust
> > >        // verify valid signatures on the transaction
> > >        {
> > >            if transaction
> > >                .verify_with_results()
> > >                .iter()
> > >                .any(|valid| !*valid)
> > >            {
> > >                return Ok(self.with_contextualized_svm_reader(|svm_reader| {
> > >                    svm_reader
> > >                        .notify_invalid_transaction(transaction.signatures[0], status_tx.clone());
> > >                }));
> > >            }
> > >        }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > We could add a similar block to `simulate_transaction`, but where this block is only called if the user passes `config.sig_verify = true`.
> > > Let me know if you want to implement this, @daog1, or if you'd like me to take over!
> > 
> > 
> > Yes, I also think changing it to global is not ideal. To achieve individual control, it would probably require modifying the litesvm code, right?
> 
> Exactly, that's why I'm proposing disabling this permanently at the LiteSVM level, then implementing the verification checks manually on our wrapper

Got it, I'll take a look tomorrow first.

MicaiahReid (approved):
Thanks @daog1

PR #154, opened 2025-06-09
feat(mcp): `call_surfnet_rpc` method

BretasArthur1:


crates/mcp/src/surfpool/mod.rs
```diff
@@ -259,6 +290,89 @@ impl Surfpool {
             },
         ))
     }
+
+    /// Calls any RPC method on a running surfnet instance.
+    /// This generic method allows calling any of the available surfnet cheatcode RPC methods.
+    /// The LLM will interpret user requests and determine which method to call with appropriate parameters.
+    #[tool(
+        description = "Calls any RPC method on a running surfnet instance. This is a generic method that can invoke any surfnet cheatcode RPC method. The LLM should interpret user requests and determine the appropriate method and parameters to call. The retrive the list of RPC endpoints available check the resource str:///rpc_endpoint_list"
```
MicaiahReid:
```suggestion
        description = "Calls any RPC method on a running surfnet instance. This is a generic method that can invoke any surfnet RPC method. The LLM should interpret user requests and determine the appropriate method and parameters to call. To retrieve the list of RPC endpoints available check the resource str:///rpc_endpoint_list"
```

MicaiahReid (changes requested):
Looking good!

PR #159, opened 2025-06-13
feat(rpc) : add account subscription to the WS rpc methods

BretasArthur1:


crates/core/src/rpc/ws.rs
```diff
@@ -193,4 +228,120 @@ impl Rpc for SurfpoolWsRpc {
...
+                    if let Some(sink) = account_active.read().unwrap().get(&sub_id) {
+                        let _ = sink.notify(Ok(RpcResponse {
+                            context: RpcResponseContext::new(
+                                svm_locker.with_svm_reader(|svm| svm.get_latest_absolute_slot()),
+                            ),
+                            value: ui_account,
+                        }));
+                    }
+                }
+            }
```
MicaiahReid: It's unclear to me from the docs - should this actually send the account when the socket is first initiated? Or should it only send updates? If only updates, this can be removed. Can you test by connecting to the mainnet WS RPC, doing a `accountSubscribe`, and checking if it sends an account when there's no updates?

crates/core/src/surfnet/svm.rs
```diff
@@ -323,7 +326,10 @@ impl SurfnetSvm {
         self.account_registry
             .entry(account.owner)
             .or_insert_with(Vec::new)
-            .push((*pubkey, account));
+            .push((*pubkey, account.clone()));
+
+        // Notify account subscribers
+        self.notify_account_subscribers(pubkey, &account);
```
MicaiahReid: Teeny tiny optimization, not that it really matters.

But would it work if you moved this notify to _before_ the `self.account_registry` to remove the need to do a clone when pushing to the account registry?

crates/core/src/surfnet/svm.rs
```diff
@@ -66,6 +67,7 @@ pub struct SurfnetSvm {
     pub simnet_events_tx: Sender<SimnetEvent>,
     pub geyser_events_tx: Sender<GeyserEvent>,
     pub signature_subscriptions: HashMap<Signature, Vec<SignatureSubscriptionData>>,
+    pub account_subscriptions: HashMap<Pubkey, Vec<(Option<UiAccountEncoding>, Sender<Account>)>>,
```
MicaiahReid: From what I can tell, you don't actually end up using this `UiAccountEncoding`, do you? We can remove from this map if so

crates/core/src/rpc/ws.rs
```diff
@@ -193,4 +228,120 @@ impl Rpc for SurfpoolWsRpc {
...
+
+            // subscribe to account updates
+            let rx = svm_locker.subscribe_for_account_updates(&pubkey, config.encoding);
+
+            loop {
```
MicaiahReid, line 614: We need to be able to break out of this loop. Currently, if you never receive an account update, you'll never break out of the loop, right?

I think we need something like:
```rust
if let Some(sink) = account_active.read().unwrap().get(&sub_id).is_none() {
     break;
}
```
after each `rx.try_recv()`

crates/core/src/rpc/ws.rs
BretasArthur1: So, for I found in the documentation and doing some research/testing the behavior in the endpoint is:

1. Gets the RPC context
2. Gets the current account from the SVM
3. Sends the initial account state if the account exists
4. Then subscribes to future account updates

And what I did was:

```rust


//  Get current account state
let result = svm_locker.get_account(&remote_ctx, &pubkey, None).await;

// If account exists, send initial notification 
if let GetAccountResult::FoundAccount(_, account, _) = account_result {
    let ui_account = encode_ui_account(...);
    if let Some(sink) = account_active.read().unwrap().get(&sub_id) {
        let _ = sink.notify(Ok(RpcResponse {
            context: RpcResponseContext::new(/* current slot */),
            value: ui_account,
        }));
    }
}

```
So first it checks if the account exists and if it does we subscribe and keep receiving updates when the account changes

crates/core/src/surfnet/svm.rs
BretasArthur1: Actually I'm gone to change the approach to use it so users can decide the encoding type instead the raw account

crates/core/src/rpc/ws.rs
MicaiahReid: Hmm, can you point to some docs for this? I just tested on ws://api.devnet.solana.com and it didn't send the initial account.

crates/core/src/rpc/ws.rs
```diff
@@ -193,4 +228,88 @@ impl Rpc for SurfpoolWsRpc {
...
+        let meta = meta.clone();
+
+        self.tokio_handle.spawn(async move {
+            account_active.write().unwrap().insert(sub_id.clone(), sink);
+
+            let SurfnetRpcContext { svm_locker, .. } =
+                match meta.get_rpc_context(config.commitment.unwrap_or_default()) {
+                    Ok(res) => res,
+                    Err(_) => panic!(),
+                };
```
MicaiahReid: Since you don't need the rpc context, you should be able to do `meta.get_svm_locker`

crates/core/src/rpc/ws.rs
```diff
@@ -193,4 +228,88 @@ impl Rpc for SurfpoolWsRpc {
...
+                    if let Some(sink) = account_active.read().unwrap().get(&sub_id) {
+                        let _ = sink.notify(Ok(RpcResponse {
+                            context: RpcResponseContext::new(
+                                svm_locker.with_svm_reader(|svm| svm.get_latest_absolute_slot()),
+                            ),
+                            value: ui_account,
+                        }));
+                    }
+                }
+                tokio::time::sleep(tokio::time::Duration::from_millis(50)).await;
```
MicaiahReid:
```suggestion
```

crates/core/src/surfnet/svm.rs
```diff
@@ -66,6 +67,7 @@ pub struct SurfnetSvm {
     pub simnet_events_tx: Sender<SimnetEvent>,
     pub geyser_events_tx: Sender<GeyserEvent>,
     pub signature_subscriptions: HashMap<Signature, Vec<SignatureSubscriptionData>>,
+    pub account_subscriptions: HashMap<Pubkey, Vec<(Option<UiAccountEncoding>, Sender<UiAccount>)>>,
```
MicaiahReid: This is a pretty big type, can we create a named type `type AccountSubscriptionData = (Option<UiAccountEncoding>, Sender<UiAccount>)` like we do for the signatures?

crates/core/src/surfnet/svm.rs
```diff
@@ -707,6 +727,36 @@ impl SurfnetSvm {
         }
     }
 
+    pub fn notify_account_subscribers(
+        &mut self,
+        account_updated_pubkey: &Pubkey,
+        account: &Account,
+    ) {
+        //TODO should we notify the subscribers in a separate thread?
```
```suggestion
```
Maybe I'm missing something, but I don't think there's a reason to put this in a separate thread. What're your thoughts on why we would?

MicaiahReid (changes requested):
Pretty much there, just a few small items!

crates/core/src/surfnet/svm.rs
BretasArthur1: mutiple signatures

BretasArthur1: but with mutiple surfnets we can have mutiple signatures

crates/core/src/rpc/ws.rs
```diff
@@ -193,4 +228,82 @@ impl Rpc for SurfpoolWsRpc {
...
+        if removed.is_some() {
+            Ok(true)
+        } else {
+            Err(Error {
+                code: ErrorCode::InvalidParams,
+                message: "Invalid subscription.".into(),
+                data: None,
+            })
+        }
+    }
```
lgalabru, line 664: I see a few `self.account_active.write()` / `self.account_active.read()` being unwrapped, are these safe?

crates/core/src/rpc/ws.rs
```diff
@@ -51,11 +61,36 @@ pub trait Rpc {
...
+    #[pubsub(
+        subscription = "accountNotification",
+        unsubscribe,
+        name = "accountUnsubscribe"
+    )]
+    fn account_unsubscribe(
+        &self,
+        meta: Option<Self::Metadata>,
+        subscription: SubscriptionId,
+    ) -> Result<bool>;
```
lgalabru, line 339: Could we add some documentation for this 2 functions? 🙏

lgalabru (approved):
Looking sharp, thanks @BretasArthur1 !
Some non blocking feedback, but would appreciate if we could address!

PR #161, opened 2025-06-17
feat(rpc): add slot subscription methods to WS RPC interface

BretasArthur1:


crates/core/src/rpc/ws.rs
```diff
@@ -58,6 +59,13 @@ pub struct RpcAccountSubscribeConfig {
     pub encoding: Option<UiAccountEncoding>,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RpcSlotSubscribeConfig {
+    #[serde(flatten)]
+    pub commitment: Option<CommitmentConfig>,
+}
+
```
MicaiahReid:
```suggestion
```
unused struct

crates/types/src/types.rs
```diff
@@ -522,3 +522,11 @@ pub struct SupplyUpdate {
     pub non_circulating: Option<u64>,
     pub non_circulating_accounts: Option<Vec<String>>,
 }
+
+#[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RpcSlotResult {
```
MicaiahReid: There's a `SlotInfo` struct exported by `solana-rpc-client-api` that we should use instead

crates/core/src/rpc/ws.rs
```diff
@@ -375,6 +397,7 @@ pub struct SurfpoolWsRpc {
     pub uid: atomic::AtomicUsize,
     pub active: Arc<RwLock<HashMap<SubscriptionId, Sink<RpcResponse<RpcSignatureResult>>>>>,
     pub account_active: Arc<RwLock<HashMap<SubscriptionId, Sink<RpcResponse<UiAccount>>>>>,
+    pub slot_active: Arc<RwLock<HashMap<SubscriptionId, Sink<RpcSlotResult>>>>,
```
MicaiahReid: I've not liked the names of these variables, can we update? `subscription_signature_result_map`, `signature_account_result_map`, etc

crates/core/src/rpc/ws.rs
```diff
@@ -337,6 +345,19 @@ pub trait Rpc {
...
+    #[pubsub(
+        subscription = "slotNotification",
+        unsubscribe,
+        name = "slotUnsubscribe"
+    )]
+    fn slot_unsubscribe(
+        &self,
+        meta: Option<Self::Metadata>,
+        subscription: SubscriptionId,
+    ) -> Result<bool>;
```
MicaiahReid: Can we get docs?

crates/core/src/surfnet/svm.rs
```diff
@@ -738,6 +740,12 @@ impl SurfnetSvm {
             self.latest_epoch_info.slot_index = 0;
             self.latest_epoch_info.epoch += 1;
         }
+
+        let parent_slot = self.latest_epoch_info.absolute_slot - 1;
+        let new_slot = self.latest_epoch_info.absolute_slot;
+        let root = parent_slot;
```
MicaiahReid: From looking up what the root is, it's the most recent finalized slot.

We have a const to dictate when a block is finalized
```suggestion
        let root = new_slot - FINALIZATION_SLOT_THRESHOLD;
```

BretasArthur1: yep, nice, I didn't noticed that

crates/core/src/surfnet/svm.rs
```diff
@@ -743,7 +744,7 @@ impl SurfnetSvm {
 
         let parent_slot = self.latest_epoch_info.absolute_slot - 1;
         let new_slot = self.latest_epoch_info.absolute_slot;
-        let root = parent_slot;
+        let root = new_slot - FINALIZATION_SLOT_THRESHOLD;
```
lgalabru: Let's, generally speaking, favor `saturating_sub`.

crates/core/src/rpc/ws.rs
```diff
@@ -555,7 +649,11 @@ impl Rpc for SurfpoolWsRpc {
         _meta: Option<Self::Metadata>,
         subscription: SubscriptionId,
     ) -> Result<bool> {
-        let removed = self.active.write().unwrap().remove(&subscription);
+        let removed = self
+            .signature_subscription_map
+            .write()
+            .unwrap()
```
lgalabru: Is this `unwrap` under control?

lgalabru: Same question for the 2 other occurrences in this file :)

MicaiahReid (changes requested):
To Ludo's point about the unwraps in the ws.rs file - I started this, that's my bad.

I first implemented the WS stuff under the gun, and didn't clean up my unwraps. If we hit one of those unwraps, that WS thread will die, and the user will be unexpectedly disconnected.

In general, for those error cases, we should properly close out the connection. Could you fix that across the whole WS file?

You can do this in a follow-on PR to prevent this one from being held up, if you prefer.

BretasArthur1:
that's also my bad, I know the issues with unwrap but didn't come over with a better approach, but yes I will change the approach on it to be more safety

MicaiahReid (approved):
Looks good! Would love to see https://github.com/txtx/surfpool/pull/161/files#r2153025951 addressed real quick, but merge after that!

PR #162, opened 2025-06-18
feat(rpc): implement `getTokenAccountsbyDelegate` endpoint

Ozodimgba:
@MicaiahReid | @lgalabru added getTokenAccountsByDelegate, please skim through if you'd like to change anything because its a big one

### DIFF
--- a/crates/core/src/surfnet/remote.rs
+++ b/crates/core/src/surfnet/remote.rs
@@ -190,7 +190,7 @@ impl SurfnetRemoteClient {
         filter: &TokenAccountsFilter,
         config: &RpcAccountInfoConfig,
     ) -> SurfpoolResult<Vec<RpcKeyedAccount>> {
-        // Validate that the program is supported if using ProgramId filter
+        // validate that the program is supported if using ProgramId filter
         if let TokenAccountsFilter::ProgramId(program_id) = &filter {
             if !is_supported_token_program(program_id) {
                 return Err(SurfpoolError::unsupported_token_program(*program_id));
@@ -204,7 +204,6 @@ impl SurfnetRemoteClient {
             }
         };
 
-        // Use the RPC client's send method for consistency with get_token_accounts_by_owner
         let res: RpcResult<Vec<RpcKeyedAccount>> = self
             .client
             .send(


crates/core/src/surfnet/locker.rs
```diff
@@ -596,6 +597,72 @@ impl SurfnetSvmLocker {
...
+            // for local-only mode, get from SVM indexes using your new method
+            self.with_svm_reader(|svm_reader| {
+                let local_accounts = svm_reader.get_token_accounts_by_delegate(&delegate);
+                local_accounts
+                    .into_iter()
+                    .filter_map(|(pubkey, token_account)| {
+                        // filter
+                        let include = match &token_account_filter {
+                            TokenAccountsFilter::Mint(mint) => token_account.mint == *mint,
+                            TokenAccountsFilter::ProgramId(_) => true, // all token accounts are under token program
```
MicaiahReid: This is not true, there are different potential token programs (SPL Token, Token 2022, or a custom lib)

crates/core/src/surfnet/locker.rs
```diff
@@ -596,6 +597,72 @@ impl SurfnetSvmLocker {
 
         Ok(local_accounts.with_new_value((keyed_accounts, missing_pubkeys)))
     }
+
+    pub async fn get_all_token_accounts_by_delegate(
```
MicaiahReid: I think this one should be organized more like `get_program_accounts`.

We'll need to implement:
 - `get_token_accounts_by_delegate_local` which:
   - calls `svm_reader. get_token_accounts_by_delegate`, then filters with the token accounts by `token_account_filter`
 - `get_token_accounts_by_delegate_local_then_remote` which:
   - gets local accounts via `get_token_accounts_by_delegate_local`
   - gets remote accounts via `remote_accounts = client.get_token_accounts_by_delegate`
   - combines the account in one list. if the account is in both the local and remote list, we should add the local one and not the remote
 - `get_token_accounts_by_delegate` which:
   - calls `get_token_accounts_by_delegate_local_then_remote` if remote ctx is provided
   - calls `get_token_accounts_by_delegate_local` if remote ctx is not provided

MicaiahReid (changes requested):

crates/core/src/surfnet/locker.rs
Ozodimgba: Oh ok, custom lib means…we shouldn’t even set a program id filter at all or there should be a param to pass in custom id?

MicaiahReid: SPL and Token 2022 are token program standards, but technically a user can create their own token program. From why I understand, a user should be able to pass a custom token program and get results.

However, for this to be possible, the user's token program format would have to be derserializable to the token account type.

For example, Token 2022 has all of the same fields as an SPL token, then further extends it. Because of this, the `data` field of an account owned by token 2022 can be deserialized to the Token Account type. However, those "extended" fields that the Token 2022 adds won't be included in the parsed account.

---

So all that to say, this is getting a bit complex. For now, let's stick to supporting SPL and Token 2022 out of the box, and adding custom token programs if users request it down the line.

Ozodimgba: noted

crates/core/src/rpc/accounts_scan.rs
```diff
@@ -635,12 +635,73 @@ impl AccountsScan for SurfpoolAccountsScanRpc {
...
+                            &config,
+                        )
+                        .await?;
+
+                    Ok(RpcResponse {
+                        context: RpcResponseContext::new(slot),
+                        value: keyed_accounts,
+                    })
+                }
+            }
```
MicaiahReid, line 703:
```suggestion
  let filter = match token_account_filter {
            RpcTokenAccountsFilter::Mint(mint_str) => TokenAccountsFilter::Mint(verify_pubkey(&mint_str)?),
            RpcTokenAccountsFilter::ProgramId(program_id_str) => TokenAccountsFilter::ProgramId(verify_pubkey(&program_id_str)?)
        };
        let remote_ctx = remote_ctx.map(|(r, _)| r);
        let SvmAccessContext {
            slot,
            inner: (keyed_accounts, _missing_pubkeys),
            ..
        } = svm_locker
            .get_all_token_accounts_by_delegate(
                &remote_ctx,
                delegate,
                filter,
            )
            .await?;
                    
                    
        Ok(RpcResponse {
            context: RpcResponseContext::new(slot),
            value: keyed_accounts,
        })
```

Something like this would clean things up

MicaiahReid: I'll actually take care of this later, let's ship!

MicaiahReid (approved):
Ship it!

PR #163, opened 2025-06-18
 added minimumLedgerSlot

Ozodimgba:


crates/core/src/rpc/full.rs
```diff
@@ -1631,8 +1631,16 @@ impl Full for SurfpoolFullRpc {
         })
     }
 
-    fn minimum_ledger_slot(&self, _meta: Self::Metadata) -> Result<Slot> {
-        not_implemented_err("minimum_ledger_slot")
+    fn minimum_ledger_slot(&self, meta: Self::Metadata) -> Result<Slot> {
+        let svm_locker = meta.get_svm_locker()
+            .map_err(|_| Error::internal_error())?;
```
MicaiahReid:
```suggestion
        let svm_locker = meta.get_svm_locker()?;
```
This already returns an error that implements `Into<Result>`, so no need to create a new error

MicaiahReid:
> I'm kind of torn on if this should return the earliest slot that surfpool has, or if it should forward to the upstream.
> 
> Thoughts, @lgalabru?

@Ozodimgba, we discussed - let's forward to the remote

Ozodimgba:
Ok will do

Ozodimgba:
@MicaiahReid added a fallback...let me know if you want to implement without it

MicaiahReid (approved):

crates/core/src/rpc/full.rs
```diff
@@ -1631,8 +1631,31 @@ impl Full for SurfpoolFullRpc {
...
+            if let Some((remote_client, _)) = remote_ctx {
+                remote_client
+                    .client
+                    .minimum_ledger_slot()
+                    .await
+                    .map_err(|e| SurfpoolError::client_error(e).into())
+            } else {
+                let min_slot = svm_locker.with_svm_reader(|svm_reader| {
+                    svm_reader.blocks.keys().min().copied().unwrap_or(0)
+                });
```
BretasArthur1, line 1654: unwrap_or(0) is a design choice just ensure that is this expected by the users 👍

BretasArthur1: but this is fine in my pov

BretasArthur1 (approved):

PR #167, opened 2025-06-18
feat(core): implement `getStakeMinimumDelegation` RPC endpoint

Ozodimgba:
@MicaiahReid  went with the assumption that surfpool doesnt have minimum stake

### DIFF
--- a/crates/core/src/rpc/full.rs
+++ b/crates/core/src/rpc/full.rs
@@ -2023,10 +2023,31 @@ impl Full for SurfpoolFullRpc {
 
     fn get_stake_minimum_delegation(
         &self,
-        _meta: Self::Metadata,
-        _config: Option<RpcContextConfig>,
+        meta: Self::Metadata,
+        config: Option<RpcContextConfig>,
     ) -> Result<RpcResponse<u64>> {
-        not_implemented_err("get_stake_minimum_delegation")
+        let config = config.unwrap_or_default();
+        let commitment_config = config.commitment.unwrap_or(CommitmentConfig {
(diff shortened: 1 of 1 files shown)


crates/core/src/rpc/full.rs
```diff
@@ -2023,10 +2023,31 @@ impl Full for SurfpoolFullRpc {
...
-        _meta: Self::Metadata,
-        _config: Option<RpcContextConfig>,
+        meta: Self::Metadata,
+        config: Option<RpcContextConfig>,
     ) -> Result<RpcResponse<u64>> {
-        not_implemented_err("get_stake_minimum_delegation")
+        let config = config.unwrap_or_default();
+        let commitment_config = config.commitment.unwrap_or(CommitmentConfig {
+            commitment: CommitmentLevel::Processed,
+        });
```
MicaiahReid, line 2096:
```suggestion
        let commitment_config = config.commitment.unwrap_or_default();
```
isn't there a default here you can use? Could be wrong

MicaiahReid (approved):
Minor nit, but looks good!

crates/core/src/rpc/full.rs
```diff
@@ -3716,4 +3737,49 @@ mod tests {
...
+            }),
+            min_context_slot: None,
+        });
+
+        let result = setup
+            .rpc
+            .get_stake_minimum_delegation(Some(setup.context.clone()), config)
+            .unwrap();
+
+        // Sshould return finalized slot
```
BretasArthur1, line 3839: Should

BretasArthur1 (approved):
LGTM!

PR #170, opened 2025-06-18
feat(rpc): implement `getGenesisHash` RPC endpoint

thewuhxyz:


crates/core/src/rpc/minimal.rs
```diff
@@ -631,8 +631,9 @@ impl Minimal for SurfpoolMinimalRpc {
             .map_err(Into::into)
     }
 
-    fn get_genesis_hash(&self, _meta: Self::Metadata) -> Result<String> {
-        not_implemented_err("get_genesis_hash")
+    fn get_genesis_hash(&self, meta: Self::Metadata) -> Result<String> {
+        meta.with_svm_reader(|svm_reader| svm_reader.genesis_config.hash().to_string())
```
MicaiahReid: I'm actually thinking that this should come from the remote, if the remote ctx is available.

If not I think this config default is appropriate.

thewuhxyz: hmm so we make this async? or we load genesis from remote on startup?

MicaiahReid: We can make async!

crates/core/src/surfnet/locker.rs
```diff
@@ -980,6 +980,33 @@ impl SurfnetSvmLocker {
...
+    }
+
+    pub async fn get_genesis_hash_local_then_remote(
+        &self,
+        client: &SurfnetRemoteClient,
+    ) -> SurfpoolContextualizedResult<Hash> {
+        let local_hash = self.get_genesis_hash_local();
+        let remote_hash = client.get_genesis_hash().await?;
+
+        Ok(local_hash.with_new_value(remote_hash))
```
MicaiahReid, line 1364: Note for myself: clean up this fn to remove some unnecessary stuffs

PR #172, opened 2025-06-18
feat(rpc): Implement `getInflationGovernor` RPC endpoint

thewuhxyz:


MicaiahReid (approved):
TBH, I'm not sure if we should just return empty data here, or if we should have it stored on the SVM like you have.

With it on the SVM, in the future we could allow users to configure it manually via cheatcode, if they want.

However, surfpool has no concept of inflation, so maybe it should just return an empty object.

Thoughts, @lgalabru?

BretasArthur1 (approved):

PR #173, opened 2025-06-18
rpc get_max_retransmit_slot

thewuhxyz:


MicaiahReid (changes requested):
I definitely don't love adding new imports for this.

Surfpool doesn't retransmit blocks, can we just use our latest slot number instead?

MicaiahReid (approved):

PR #174, opened 2025-06-18
feat(rpc): implement `getInflationRate` RPC endpoint

thewuhxyz:


MicaiahReid (approved):
I'll give a reluctant approval - I think we should likely rethink this method and the other inflation related endpoints

PR #175, opened 2025-06-18
feat(rpc): Implement `getLargestAccounts` RPC endpoint

thewuhxyz:


crates/core/src/rpc/accounts_scan.rs
```diff
@@ -525,10 +525,47 @@ impl AccountsScan for SurfpoolAccountsScanRpc {
...
+                        })
+                        .collect();
+
+                    Ok(RpcResponse {
+                        context: RpcResponseContext::new(current_slot),
+                        value: largest_accounts,
+                    })
+                })
+            }
+        })
```
MicaiahReid: This logic should be on the `locker.rs` file as `svm_locker.get_largest_accounts(remote_ctx)`

You can follow the pattern set by `svm_locker.get_accounts`

crates/core/src/rpc/accounts_scan.rs
```diff
@@ -525,10 +525,47 @@ impl AccountsScan for SurfpoolAccountsScanRpc {
...
+            svm_locker,
+            remote_ctx,
+        } = match meta.get_rpc_context(()) {
+            Ok(res) => res,
+            Err(e) => return e.into(),
+        };
+
+        Box::pin(async move {
+            let current_slot = svm_locker.get_latest_absolute_slot();
+            if let Some((remote_client, _)) = remote_ctx {
```
MicaiahReid: If the network is configured to have a remote, this means we need to _combine_ the remote and local data, not that we skip the local data completely.

For example, on the `svm_locker` we implement:
 - `get_accounts_local`
 - `get_accounts_local_then_remote` - this calls `get_accounts_local`, then fetches from remote, then merges the results
 - `get_accounts` - this just routes to either `get_accounts_local` or `get_accounts_local_then_remote` depending on if we have a remote_ctx.

We should follow the same pattern for this function

crates/core/src/error.rs
```diff
@@ -148,6 +148,18 @@ impl SurfpoolError {
...
+        T: ToString,
+    {
+        let mut error = Error::internal_error();
+        error.data = Some(json!(format!(
+            "Failed to fetch accounts from remote: {}",
+            e.to_string()
+        )));
+        Self(error)
+    }
+
```
MicaiahReid: This is returning the same error as `get_multiple_accounts` above - we can either just use `get_multiple_accounts` and remove this error, or change the message here to `Failed to get largest accounts from remote`

crates/core/src/surfnet/locker.rs
```diff
@@ -311,6 +314,57 @@ impl SurfnetSvmLocker {
...
+                    lamports: account.lamports,
+                })
+                .collect()
+        })
+    }
+
+    pub async fn get_largest_accounts_local_then_remote(
+        &self,
+        client: &SurfnetRemoteClient,
+        config: Option<RpcLargestAccountsConfig>,
```
MicaiahReid: We're passing this config to the remote client, which will filter results by circulating/non-circulating accounts.

However, we aren't using this filter on the local results. The svm has `non_circulating_accounts` on it that can be used for the filter.

This is a bit confusing but I think we should be:
 1. filtering the results of `get_largest_accounts_local`
 2. when we do `remote_client.get_largest_accounts`, we should always _remove_ the filter (so it returns all largest accounts) _then_ we manually apply the filter based on our local svm's `non_circulating_accounts` setting

MicaiahReid (changes requested):
Looking much better! Just one last thing I realized about the filtering

crates/core/src/surfnet/locker.rs
MicaiahReid: To add some clarity, here is what I think we should do for this method:
 1. Make two requests to the remote - one for the largest circulating accounts, one for the largest non-circulating accounts
 2. If those accounts don't yet exist locally, insert them
 3. Pull from our local cache to find the largest accounts, with the filter applied. If the filter is provided, it should be looking at the `SurfnetSvm`'s `non_circulating_accounts` to see what accounts are circulating/non-circulating

PR #176, opened 2025-06-18
feat(rpc): Implement `getMaxShredInsertSlot` RPC endpoint

thewuhxyz:


MicaiahReid (changes requested):
I think we should treat this like `getMaxRetransmitSlot` and just return the latest slot to prevent importing a new crate and this extra logic

MicaiahReid (approved):

BretasArthur1 (approved):

PR #178, opened 2025-06-18
feat(rpc): Implement `getSignaturesForAddress` RPC endpoint

thewuhxyz:


crates/core/src/rpc/full.rs
```diff
@@ -1956,11 +1956,61 @@ impl Full for SurfpoolFullRpc {
...
+                            .unwrap_or_default()
+                    })
+                    .map(|(sig, status)| (sig.to_owned(), status.to_owned()))
+                    .collect()
+            });
+
+            let mut responses = Vec::with_capacity(sigs.len());
+
+            for (signature, status) in sigs.into_iter() {
+                let res = svm_locker.get_transaction(&remote_ctx, &signature).await;
```
MicaiahReid: Looks like we only needed to fetch the transaction to find the confirmation status. I think we can skip this fetch, we've been using the following to get the status:
 - `slot == svm_locker.get_latest_block_hash()` - Processed
 - `slot == svm_locker.get_latest_block_hash().saturating_sub(1)` - Confirmed
 - `slot == svm_locker.get_latest_block_hash().saturating_sub(FINALIZATION_SLOT_THRESHOLD)` - Finalized

crates/core/src/rpc/full.rs
```diff
@@ -1956,11 +1956,61 @@ impl Full for SurfpoolFullRpc {
...
+                        tx.message
+                            .static_account_keys()
+                            .iter()
+                            .find_position(|pubkey| pubkey.to_bytes() == address.to_bytes())
+                            .map(|(idx, _)| tx.message.is_signer(idx))
+                            .unwrap_or_default()
+                    })
+                    .map(|(sig, status)| (sig.to_owned(), status.to_owned()))
+                    .collect()
+            });
```
MicaiahReid: Let's make this a function in the `locker.rs` file on `svm_locker`. On the locker, we'll need to implement:
 - `get_signatures_for_address_local`
 - `get_signatures_for_address_local_then_remote`: gets from local, gets from remote, then merges them
 - `get_signatures_for_address` - routes to either `get_signatures_for_address_local` or `get_signatures_for_address_local_then_remote` depending on if there is a `remote_ctx` or not

This is similar to the `get_accounts` implementation on the locker.

The reason is - the requested address could have some transactions on mainnet _before_ the surfnet was created. The account could also have been involved with new transactions locally. We'll want to include both in this list.

crates/core/src/surfnet/locker.rs
```diff
@@ -313,6 +320,87 @@ impl SurfnetSvmLocker {
...
+                })
+                .collect()
+        })
+    }
+
+    pub async fn get_signatures_for_address_local_then_remote(
+        &self,
+        client: &SurfnetRemoteClient,
+        pubkey: &Pubkey,
+        config: Option<RpcSignaturesForAddressConfig>,
```
MicaiahReid, line 451: We aren't using this config currently for the local. Specifically the `before`, `limit`, and `until` fields need to be used here.

For example, if the limit is 1, and locally we found 1 signature, we don't need to call the remote ctx. 

If locally we have 10 signatures and the limit is 1000, we'll need to truncate the combined list to the limit.

If the use provides the `until` field, and that signature is found locally, no need to go to remote. etc. etc.

MicaiahReid: We're also not checking the `min_context_slot` field. I think FX has already used this config a few times, so you should be able to copy what he did

MicaiahReid (changes requested):

crates/core/src/surfnet/locker.rs
```diff
@@ -313,6 +321,129 @@ impl SurfnetSvmLocker {
...
+    ) -> SurfpoolContextualizedResult<Vec<RpcConfirmedTransactionStatusWithSignature>> {
+        let results = self.get_signatures_for_address_local(pubkey, config.clone());
+        let limit = config.clone().and_then(|c| c.limit).unwrap_or(1000);
+
+        let mut combined_results = results.inner.clone();
+        if combined_results.len() < limit {
+            let mut remote_results = client.get_signatures_for_address(pubkey, config).await?;
+            combined_results.append(&mut remote_results);
```
MicaiahReid, line 459: I think we'll need to clamp this combined result to `limit` after appending

crates/core/src/rpc/full.rs
```diff
@@ -3716,4 +3738,48 @@ mod tests {
         let expected_local: Vec<Slot> = (100..=120).collect();
         assert_eq!(result, expected_local, "Should return local blocks 100-120");
     }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_get_signatures_for_address() {
```
MicaiahReid: The logic for before/until is a bit intense, could we get some tests for it? 🙏

MicaiahReid (changes requested):

MicaiahReid:
This is looking great! This seems like one of the trickier ones, good work

BretasArthur1 (approved):

PR #179, opened 2025-06-20
fix(core): allow any slot in epoch for `getSlotLeaders` RPC

MicaiahReid:


crates/core/src/rpc/bank_data.rs
```diff
@@ -450,12 +450,17 @@ impl BankData for SurfpoolBankDataRpc {
         }
 
         let svm_locker = meta.get_svm_locker()?;
-        let latest_slot = svm_locker.get_latest_absolute_slot();
-        if start_slot >= latest_slot {
+        let epoch_info = svm_locker.get_epoch_info();
+
+        let first_slot_in_epoch = epoch_info.absolute_slot - epoch_info.slot_index;
+        let last_slot_in_epoch = first_slot_in_epoch + epoch_info.slots_in_epoch - 1;
```
BretasArthur1: saturating sub here? if the first slot is 0 the epoch info will be 0 also and then -1 will cause the underflow

BretasArthur1 (approved):

crates/core/src/rpc/bank_data.rs
MicaiahReid: Good call, thanks @BretasArthur1!!

PR #180, opened 2025-06-23
feat(rpc): Implement `getInflationReward` RPC endpoint

thewuhxyz:


MicaiahReid (changes requested):
I'm good with the approach on the return values - can we just get some basic validation of the inputs? Verify the addresses are valid pubkeys, that the config.epoch is not after the current epoch, etc

BretasArthur1 (approved):

crates/core/src/rpc/full.rs
```diff
@@ -1348,11 +1348,52 @@ impl Full for SurfpoolFullRpc {
...
+                        "Invalid epoch. Epoch is larger that current epoch",
+                    ));
+                }
+            };
+
+            let current_slot = svm_locker.get_epoch_info().absolute_slot;
+            if let Some(slot) = config.as_ref().and_then(|config| config.min_context_slot) {
+                if slot > current_slot {
+                    return Err(Error::invalid_params(
+                        "Invalid epoch. min_context_slot is larger that current slot",
```
MicaiahReid:
```suggestion
                        "Minimum context slot has not been reached",
```

MicaiahReid (approved):
Thanks, @thewuhxyz!

PR #181, opened 2025-06-23
refactor(mcp): introduce spawning a terminal as the default behavior of the start surfnet tool

BretasArthur1:


crates/mcp/src/surfpool/start_surfnet.rs
```diff
@@ -10,10 +12,28 @@ pub struct StartSurfnetResponse {
 
 #[derive(Serialize)]
 pub struct StartSurfnetSuccess {
-    pub surfnet_url: String,
+    pub kind: String,
```
MicaiahReid: Instead of an string with hard-coded values, can we use an enum with `Command` and `SubProcess` options?

crates/mcp/src/surfpool/start_surfnet.rs
```diff
@@ -30,7 +50,27 @@ impl StartSurfnetResponse {
     }
 }
 
-pub fn run(surfnet_id: u16, rpc_port: u16, ws_port: u16) -> StartSurfnetResponse {
+pub fn generate_command(_surfnet_id: u16, rpc_port: u16, ws_port: u16) -> Command {
+    let mut cmd = Command::new("surfpool");
```
MicaiahReid, line 60: I'm curious if a clap command (the options we use in starting the CLI) can be created and serialized. This would give us easy support for other CLI options without having to add to the serializer here. We'd just create an instance of the struct and serialize it. Try to experiment with the `Command` enum exported in `crates/cli/src/cli/mod.rs`

MicaiahReid (changes requested):
Tomorrow I want to test to fully grok this, but here's some comments for now

crates/mcp/src/surfpool/mod.rs
```diff
@@ -176,21 +176,41 @@ impl SurfnetRpcCallResponse {
...
+        This is the default behavior (`run_as_subprocess: false`).
+        If the user explicitly asks to run surfnet as a background process, 
+        set `run_as_subprocess: true` to start it directly. When in doubt, return the command for execution.")]
+    pub fn start_surfnet(
+        &self,
+        #[tool(param)]
+        #[schemars(
+            description = "If `false` (default), returns a command for the AI to execute. If `true`, starts surfnet directly as a background process."
+        )]
+        run_as_subprocess: Option<bool>,
```
MicaiahReid: Does this need to be optional? Seems like we're requiring the field

BretasArthur1: I will the tests we discussed and I bring a answer to this

crates/mcp/src/surfpool/start_surfnet.rs
BretasArthur1: discussed

crates/mcp/src/surfpool/mod.rs
```diff
@@ -176,21 +176,41 @@ impl SurfnetRpcCallResponse {
...
     /// This method is exposed as a tool and can be invoked remotely.
-    #[tool(
-        description = "Starts a new local Solana network (surfnet). Returns the RPC server's binding address and the surfnet ID."
-    )]
-    pub fn start_surfnet(&self) -> Json<StartSurfnetResponse> {
+    #[tool(description = "Starts a new local Solana network (surfnet).
+        This tool returns a shell command that the AI must execute in a new terminal.
+        This is the default behavior (`run_as_subprocess: false`).
+        If the user explicitly asks to run surfnet as a background process, 
+        set `run_as_subprocess: true` to start it directly. When in doubt, return the command for execution.")]
```
MicaiahReid, line 185: I'm not understanding what this part means:
`When in doubt, return the command for execution.`

crates/mcp/src/surfpool/mod.rs
```diff
@@ -176,21 +176,41 @@ impl SurfnetRpcCallResponse {
...
+            Some(false) => start_surfnet::run_command(surfnet_id, port, port.saturating_sub(9)),
+            _ => {
+                return Json(StartSurfnetResponse::error(
+                    "run_as_subprocess is required".to_string(),
+                ));
+            }
+        };
+
+        // Only track the surfnet in the registry if we've actually started it as a subprocess
+        if run_as_subprocess.unwrap_or(false) && res.success.is_some() {
```
MicaiahReid: Let's store the surfnet id for commands as well as subprocesses

crates/mcp/src/surfpool/start_surfnet.rs
```diff
@@ -30,7 +50,27 @@ impl StartSurfnetResponse {
     }
 }
 
-pub fn run(surfnet_id: u16, rpc_port: u16, ws_port: u16) -> StartSurfnetResponse {
+pub fn generate_command(_surfnet_id: u16, rpc_port: u16, ws_port: u16) -> Command {
```
MicaiahReid: Can we remove the unused id field?

crates/mcp/src/surfpool/start_surfnet.rs
MicaiahReid: Also, since the `command` is an optional that is only used when the `kind` is "command", your enum should contain that data:
```rust
enum StartSurfnetSuccessKind {
    Command(SerializedCommand),
    SubProcess
}
```

MicaiahReid (changes requested):

crates/mcp/src/surfpool/mod.rs
BretasArthur1: just for the llm context

MicaiahReid: Yeah, but what does it mean? I just want to understand

crates/mcp/src/surfpool/start_surfnet.rs
```diff
@@ -10,10 +12,34 @@ pub struct StartSurfnetResponse {
 
 #[derive(Serialize)]
 pub struct StartSurfnetSuccess {
-    pub surfnet_url: String,
+    pub kind: StartSurfnetKind,
+    pub surfnet_url: Option<String>,
```
MicaiahReid: Is there a reason for this to be optional? Seems like we always provide it when creating this struct

lgalabru (approved):
Great work, thank @BretasArthur1 🙏

PR #183, opened 2025-06-24
feat(core): implement `getVoteAccounts` RPC endpoint

Ozodimgba:


crates/core/src/rpc/minimal.rs
```diff
@@ -739,9 +739,29 @@ impl Minimal for SurfpoolMinimalRpc {
...
+        meta: Self::Metadata,
+        config: Option<RpcGetVoteAccountsConfig>,
     ) -> Result<RpcVoteAccountStatus> {
+        // validate inputs if provided
+        if let Some(config) = config {
+            // validate vote_pubkey if provided
+            if let Some(vote_pubkey_str) = config.vote_pubkey {
+                verify_pubkey(&vote_pubkey_str).map_err(|e| {
+                    jsonrpc_core::Error::invalid_params(format!("Invalid vote pubkey: {}", e))
+                })?;
```
MicaiahReid:
```suggestion
                verify_pubkey(&vote_pubkey_str)?;
```
verify pubkey should already be returning an error that should work. If you get a complier error here, you may just need
`verify_pubkey(&vote_pubkey_str).map_err(Into::into)?`

MicaiahReid (changes requested):

crates/core/src/rpc/minimal.rs
```diff
@@ -743,9 +743,27 @@ impl Minimal for SurfpoolMinimalRpc {
...
+        // validate inputs if provided
+        if let Some(config) = config {
+            // validate vote_pubkey if provided
+            if let Some(vote_pubkey_str) = config.vote_pubkey {
+                verify_pubkey(&vote_pubkey_str).map_err(jsonrpc_core::Error::from)?;
+            }
+
+            // validate delinquent_slot_distance if provided
+            if let Some(distance) = config.delinquent_slot_distance {
+                if distance > 1000 {
```
MicaiahReid: I am not seeing this requirement documented anywhere, and I'm able to run the query to devnet with a huge distance; where is this requirement coming from?

Ozodimgba: pulling from the type on solana rpc config. https://docs.rs/solana-client/latest/solana_client/rpc_config/struct.RpcGetVoteAccountsConfig.html

Ozodimgba: can remove if not necessary tho

MicaiahReid: @Ozodimgba, I'm saying where are you seeing it required for that `delinquent_slot_distance` to be `< 1000>`? I don't see that in the docs your provided either.

Ozodimgba: That is just arbitrary…I just guessed the number sinece we are mocking the original impl

MicaiahReid: From what I cal tell, the original impl doesn't have a limit here at all. Did you find any evidence that this number is limited? If not, this check needs to be removed.

Ozodimgba: ok got it

MicaiahReid (approved):

PR #184, opened 2025-06-24
feat(core): implement ```getSlotLeader``` RPC endpoint

Ozodimgba:


crates/core/src/rpc/bank_data.rs
```diff
@@ -429,10 +431,48 @@ impl BankData for SurfpoolBankDataRpc {
...
+                    return Err(RpcCustomError::MinContextSlotNotReached {
+                        context_slot: min_context_slot,
+                    }
+                    .into());
+                }
+            }
+        }
+
+        // return empty string similar to getSlotLeaders returning empty vec
+        Ok(String::new())
```
MicaiahReid: Can we return the following instead?
```
Pubkey::from_str_const("SUrFPooLSUrFPooLSUrFPooLSUrFPooLSUrFPooLSUr")
                .to_string(),
```
The reason being - this string represents a pubkey, and if a consuming library tries to take the result can convert an empty string to a pubkey, they'll get an error. We need to be sure our "fake" data doesn't break API consumers

MicaiahReid (changes requested):

MicaiahReid (approved):
Thanks @Ozodimgba!

PR #185, opened 2025-06-24
fix(core): get local accounts for `getTokenAccountsByOwner` RPC; fix …

MicaiahReid:
…`mint` filter

### DIFF
--- a/crates/core/src/error.rs
+++ b/crates/core/src/error.rs
@@ -4,6 +4,7 @@ use crossbeam_channel::TrySendError;
 use jsonrpc_core::{Error, Result};
 use serde::Serialize;
 use serde_json::json;
+use solana_client::rpc_request::TokenAccountsFilter;
 use solana_pubkey::Pubkey;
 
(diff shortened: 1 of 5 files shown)


BretasArthur1 (approved):
Looking good

BretasArthur1:
just one concern we are defining the filter logic multiple times in the code, and if we create a helper function that handles all the filter cases and use it in the specific places, it would ensure the dry principals

MicaiahReid:
> just one concern we are defining the filter logic multiple times in the code, and if we create a helper function that handles all the filter cases and use it in the specific places, it would ensure the dry principals


Yeah, it's a bummer, but the logic looks the same but is different, and we don't manage those types to make easy helpers. But we have:
 1. convert `RpcTokenAccountsFilter` to `TokenAccountsFilter`
 2.  convert `TokenAccountsFilter` to `RpcTokenAccountsFilter`
 3.  actually check the `TokenAccountsFilter` against an account

So they aren't actually doing the same things :/

PR #190, opened 2025-06-26
fix(core): index token2022 additional_data from account data rather t…

MicaiahReid:
…han ix

This PR aims to expand upon #188.

I could be missing something, but I was trying to understand our initial fix on the `additional_data` field that is sometimes used, and I think I found an important gap in the current approach.

Currently, when a transaction is successfully executed, we check the instructions to see if it contained a `InitializeMint2` instruction. If so, we'd use that data to index part of the additional data field.

The problems with this approach:
 1. We're only aware of the `additional_data` if the transaction to initialize the mint was sent to surfpool, which means that any token2022 accounts pulled from mainnet won't work
 2. We're only using the `decimals` field

The additional data we want to return with the token accounts here, from what I understand, can be retrieved via the mint account. So, if instead of checking the instructions for this additional data, whenever we are indexing our accounts (as we already do after successful transaction processing _and_ when inserting a new account from the remote) we check if is owned by the token2022 program and is able to unpacked to a mint account with state, then we can index that data.

If there's ever a `InitializeMint2` instruction, the transaction will be creating this mint account, so we'll still hit this code path. But additionally, if we ever _reference_ one of these mint accounts from mainnet, whenever we pull it from mainnet we'll be able to index that additional data, too.

### Caveat

There is still one hole in this approach: if there's a token 2022 account that the user fetches, but surfpool hasn't had to pull the mint account yet, the index doesn't have the additional data for that account yet, so we'll still fail as before. This doesn't really seem like an edge case to me - it could be relatively common.

A more robust fix might require us to, whenever we fetch any account from the remote, check if it's a token 2022 account and, if so, also fetch the mint account. This is similar to how we handle executable accounts - always fetching the data account alongside it.

### Concerns

From what I can tell, the new tests that we've added pass _before_ the original fix, #188, was merged. So the `test_transfer_token` test that was added is not actually a proper test for this.


BretasArthur1 (approved):
BRAVO!

BretasArthur1:
about 
"There is still one hole in this approach: if there's a token 2022 account that the user fetches, but surfpool hasn't had to pull the mint account yet, the index doesn't have the additional data for that account yet, so we'll still fail as before. This doesn't really seem like an edge case to me - it could be relatively common."

Should we add and pre state that verify if the token exists locally and if doesn't we unpack it?

lgalabru (approved):
I think it looks good to me. Before merging, could we please fix the test vector? cc @BretasArthur1 
🙏

PR #194, opened 2025-06-30
fix(core): set program data accounts before program accounts; other fixes

MicaiahReid:


crates/core/src/surfnet/svm.rs
```diff
@@ -1275,4 +1296,199 @@ mod tests {
...
+    ) -> bool {
+        match events_rx.recv() {
+            Ok(event) => match event {
+                SimnetEvent::AccountUpdate(_, account_pubkey) => {
+                    assert_eq!(pubkey, &account_pubkey);
+                    assert_eq!(svm.accounts_registry.get(&pubkey), Some(expected_account));
+                    true
+                }
+                event => {
+                    println!("unexpected simnet event: {:?}", event);
```
lgalabru, line 1314: no print please :)

MicaiahReid: This is a helper in the test mod, does that still count?

PR #211, opened 2025-07-08
Restore support for geyser plugins

lgalabru:


crates/core/src/runloops/mod.rs
```diff
@@ -226,51 +233,51 @@ pub fn start_clock_runloop(mut slot_time: u64) -> (Receiver<ClockEvent>, Sender<
...
+            let (plugin, lib) = unsafe {
+                let lib = match Library::new(&plugin_dylib_location.to_string()) {
+                    Ok(lib) => lib,
+                    Err(e) => {
+                        let _ = simnet_events_tx.send(SimnetEvent::ErrorLog(Local::now(), format!("Unable to load plugin {}: {}", plugin_name, e.to_string())));
+                        continue;
+                    }
+                };
+                let constructor: Symbol<PluginConstructor> = lib
+                    .get(b"_create_plugin")
```
sonicfromnewyoke, line 271: won't work for cases like [metaplex has](https://github.com/metaplex-foundation/digital-asset-validator-plugin/blob/main/plerkle/src/geyser_plugin_nft.rs#L828)

TBH, not so many guys are using the default naming like `_create_plugin`

would recommend uploading them in any other way

sonicfromnewyoke (changes requested):

crates/core/src/runloops/mod.rs
lgalabru: Ah interesting, thanks for the pointer!
I created an issue we can use for keeping track of improvements that can be made in subsequent PR https://github.com/txtx/surfpool/issues/212.
I'd love to get this first iteration merged, this PR is already touching a lot of code and can diverge quickly.

crates/core/src/runloops/mod.rs
```diff
@@ -226,51 +233,51 @@ pub fn start_clock_runloop(mut slot_time: u64) -> (Receiver<ClockEvent>, Sender<
...
 ) -> Result<JoinHandle<Result<(), String>>, String> {
     let handle = hiro_system_kit::thread_named("Geyser Plugins Handler").spawn(move || {
-        let mut plugin_manager = vec![];
+        let mut plugin_manager = GeyserPluginManager::new();
+        let mut surfpool_plugin = None;
+
+        for plugin_config_path in plugin_config_paths.into_iter() {
+            let plugin_manifest_location = FileLocation::from_path(plugin_config_path);
+            let contents = plugin_manifest_location.read_content_as_utf8()?;
+            let result: serde_json::Value = json5::from_str(&contents).unwrap();
```
MicaiahReid, line 249: Could `.map_err(...)?;` here? Seems quite likely that a user accidentally points to a non-json file, and a helpful error would be great for that

crates/core/src/runloops/mod.rs
```diff
@@ -226,51 +233,51 @@ pub fn start_clock_runloop(mut slot_time: u64) -> (Receiver<ClockEvent>, Sender<
...
+            let (plugin_name, plugin_dylib_path) = match (result.get("name").map(|p| p.as_str()), result.get("libpath").map(|p| p.as_str())) {
+                (Some(Some(name)), Some(Some(libpath))) => (name, libpath),
+                _ => {
+                    let error = format!("Unable to retrieve dylib: {}", plugin_manifest_location);
+                    let _ = simnet_events_tx.send(SimnetEvent::ErrorLog(Local::now(), error.clone()));
```
MicaiahReid, line 254: Note to me for future cleanup: we can convert these `SimnetEvent::ErrorLog`s to `SimnetEvent::log_error(msg)`, to prevent  importing Local and just tidy things up a bit

examples/hello-geyser/hello-geyser.json
MicaiahReid, line 1: Curious - is a json file like this standard for a geyser plugin? Couldn't we pull this info from the Cargo.toml and prevent the user from needing an extra json file?

crates/core/src/runloops/mod.rs
```diff
@@ -226,51 +233,51 @@ pub fn start_clock_runloop(mut slot_time: u64) -> (Receiver<ClockEvent>, Sender<
...
+                    return Err(error)
+                }
+            };
+
+            let mut plugin_dylib_location = plugin_manifest_location.get_parent_location().expect("path invalid");
+            plugin_dylib_location.append_path(&plugin_dylib_path).expect("path invalid");
+
+            let (plugin, lib) = unsafe {
+                let lib = match Library::new(&plugin_dylib_location.to_string()) {
```
MicaiahReid, line 263: As we're moving into unsafe code here - should we validate that a file exists at this location before creating the library?

MicaiahReid:
Only blocking to get this one question answered: https://github.com/txtx/surfpool/pull/211#discussion_r2192984721

Because if we just used a Cargo.toml it would change the example + how the user interacts with it, I'd like this settled before merging. Other than that, we can take care of all comments in a follow-on PR

examples/hello-geyser/hello-geyser.json
lgalabru: From what I saw, this is standard yeah. Maybe @sonicfromnewyoke could confirm.

crates/core/src/runloops/mod.rs
lgalabru: This is happening line `266` ?

crates/core/src/runloops/mod.rs
lgalabru: Addressed in #212

crates/core/src/runloops/mod.rs
lgalabru: Addressed in #212

crates/core/src/runloops/mod.rs
MicaiahReid: I don't know what is "unsafe" about the code. But we're passing a string path in, not knowing if it exists or not, then handle errors. With this unsafe code, is there any risk of side effects of that `Libray::new` fn?

Sorry, not super familiar with unsafe rust at this point

examples/hello-geyser/hello-geyser.json
sonicfromnewyoke: @MicaiahReid 
this is how it's done in `agave` and `jito` validators.

The main point here is  somehow "tell" validator where is plugin binary  (via `libpath` field, if I'm not mistaken).

So, there is no strict rules on it. But, I guess it will be simpler for the devs to have the same `config.json` file for production and development

PR #217, opened 2025-07-09
feat: migrate subgraph storage to in-memory sqlite + additional constant cleanup

lgalabru:


crates/cli/src/cli/simnet/mod.rs
```diff
@@ -66,7 +67,7 @@ pub async fn handle_start_local_surfnet_command(
     let remote_rpc_url = config.simnets[0].remote_rpc_url.clone();
     let local_rpc_url = config.rpc.get_socket_address();
 
-    let network_binding = format!("{}:{}", cmd.network_host, DEFAULT_EXPLORER_PORT);
+    let network_binding = format!("{}:{}", cmd.network_host, DEFAULT_TXTX_PORT);
```
MicaiahReid: Is this right? The explorer & subgraphs endpoint will now run on the same port as a supervised runbook?

crates/gql/src/types/filters.rs
```diff
@@ -99,7 +98,8 @@ impl GraphQLType<DefaultScalarValue> for NumericFilter {
         DefaultScalarValue: 'r,
     {
         let args = [
-            build_number_filter_argument(&spec.name, "isEqual"),
+            build_number_filter_argument(&spec.name, "equals"),
+            build_number_filter_argument(&spec.name, "notEquals"),
             build_number_filter_argument(&spec.name, "greaterThan"),
             build_number_filter_argument(&spec.name, "greaterOrEqual"),
             build_number_filter_argument(&spec.name, "lowerThan"),
```
MicaiahReid, line 105: I think the GQL way to do this is `_eq`, `_neq`, etc, can we use those instead?

(I'm good with converting this comment to an issue and resolving later)

MicaiahReid (changes requested):
I tried testing an got an error:
```
{
  "data": null,
  "errors": [
    {
      "message": "No subgraph found for name splitTransferEvent: no such table: collections",
      "locations": [
        {
          "line": 32,
          "column": 3
        }
      ],
      "path": [
        "splitTransferEvent"
      ]
    }
  ]
}
```

Are we expecting this to be stable right now? Also, is the new db stuff still just in memory - no db files are currently introduced?

MicaiahReid:
Ooh, another time I got the request to work, but we're not properly decoding txtx addon data:
```
{
  "data": {
    "splitTransferEvent": [
      {
        "recipient_1": "AddonData { bytes: \"0x1af07435743932886236717396e6f3b01999ed74178b3ae123d78ff796fdd111\", id: \"svm::pubkey\" }"
      }
    ]
  }
}
```

crates/gql/src/types/filters.rs
lgalabru: hmm do you have a source? I've definitely seen different conventions.

crates/cli/src/cli/simnet/mod.rs
lgalabru: yes, this is where we're headed with the studio

MicaiahReid: Cool, I like it. But right now if you run the supervisor then try to reach the gql console, you'll be redirected to the supervisor, making the console unreachable

MicaiahReid (approved):

PR #218, opened 2025-07-09
feat(core): Enhance block retrieval with configuration support

### LINKED ISSUE #206: Solscan homepage: Blocks list is always empty
Image

BretasArthur1:
Closes #206

### DIFF
--- a/crates/core/src/error.rs
+++ b/crates/core/src/error.rs
@@ -4,7 +4,7 @@ use crossbeam_channel::TrySendError;
 use jsonrpc_core::{Error, Result};
 use serde::Serialize;
 use serde_json::json;
-use solana_client::rpc_request::TokenAccountsFilter;
+use solana_client::{client_error::ClientError, rpc_request::TokenAccountsFilter};
 use solana_pubkey::Pubkey;
 use solana_sdk::slot_history::Slot;
 
@@ -375,9 +375,11 @@ impl SurfpoolError {
         )))
     }
 
-    pub fn get_block(e: solana_client::client_error::ClientError) -> Self {
+    pub fn get_block(e: ClientError, block: Slot) -> Self {
         let mut error = Error::internal_error();
-        error.data = Some(json!(format!("Failed to get block: {e}")));
+        error.data = Some(json!(format!(
+            "Failed to get block {block} from remote: {e}"
+        )));
         Self(error)
     }
 }
--- a/crates/core/src/error.rs
+++ b/crates/core/src/error.rs
@@ -6,6 +6,7 @@ use serde::Serialize;
 use serde_json::json;
 use solana_client::rpc_request::TokenAccountsFilter;
 use solana_pubkey::Pubkey;
+use solana_sdk::slot_history::Slot;
 
 pub type SurfpoolResult<T> = std::result::Result<T, SurfpoolError>;
 
@@ -366,5 +367,17 @@ impl SurfpoolError {
         Self(Error::invalid_params(
             "sigVerify may not be used with replaceRecentBlockhash",
         ))
+    }
+
+    pub fn slot_too_old(slot: Slot) -> Self {
+        Self(Error::invalid_params(format!(
+            "Requested {slot} is before the first local slot, and no remote RPC was provided."
+        )))
+    }
+
+    pub fn get_block(e: solana_client::client_error::ClientError) -> Self {
+        let mut error = Error::internal_error();
+        error.data = Some(json!(format!("Failed to get block: {e}")));
+        Self(error)
     }
 }
--- a/crates/core/src/rpc/full.rs
+++ b/crates/core/src/rpc/full.rs
@@ -1737,16 +1737,16 @@ impl Full for SurfpoolFullRpc {
         &self,
         meta: Self::Metadata,
         slot: Slot,
-        _config: Option<RpcEncodingConfigWrapper<RpcBlockConfig>>,
+        config: Option<RpcEncodingConfigWrapper<RpcBlockConfig>>,
     ) -> BoxFuture<Result<Option<UiConfirmedBlock>>> {
+        let config = config.map(|c| c.convert_to_current()).unwrap_or_default();
+
         let svm_locker = match meta.get_svm_locker() {
             Ok(locker) => locker,
             Err(e) => return e.into(),
         };
 
-        Box::pin(async move {
-            Ok(svm_locker.with_svm_reader(|svm_reader| svm_reader.get_block_at_slot(slot)))
-        })
+        Box::pin(async move { Ok(svm_locker.get_block_local(&slot, &config).inner) })
     }
 
     fn get_block_time(
@@ -1819,7 +1819,6 @@ impl Full for SurfpoolFullRpc {
                 .map(|end| end.min(committed_latest_slot))
                 .unwrap_or(committed_latest_slot);
 
-            println!("effective_end_slot: {}", effective_end_slot);
             if let Some(min_context_slot) = config.min_context_slot {
                 if committed_latest_slot < min_context_slot {
                     return Err(RpcCustomError::MinContextSlotNotReached {
@@ -1837,18 +1836,8 @@ impl Full for SurfpoolFullRpc {
                         .get_blocks(start_slot, end_slot)
                         .await
                         .unwrap_or_else(|_| vec![]);
-                    println!(
-                        "[getBlocks] FORWARDED to remote: start_slot={}, end_slot={:?}, got {} slots",
-                        start_slot,
-                        end_slot,
-                        remote_result.len()
-                    );
                     return Ok(remote_result);
                 } else {
-                    println!(
-                        "[getBlocks] EMPTY: start_slot={}, end_slot={:?}, committed_latest_slot={}, effective_end_slot={}",
-                        start_slot, end_slot, committed_latest_slot, effective_end_slot
-                    );
                     return Ok(vec![]);
                 }
             }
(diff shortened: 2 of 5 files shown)


crates/core/src/rpc/full.rs
```diff
@@ -1771,7 +1784,9 @@ impl Full for SurfpoolFullRpc {
         };
 
         let config = config.unwrap_or_default();
-        let commitment = config.commitment.unwrap_or_default();
+        let commitment = config.commitment.unwrap_or(CommitmentConfig {
```
MicaiahReid, line 1795:
```suggestion
        // get blocks should default to processed rather than finalized to default to the most recent
        let commitment = config.commitment.unwrap_or(CommitmentConfig {
```

I want a comment here so in the future we know this was intentional

crates/core/src/rpc/full.rs
```diff
@@ -1843,47 +1849,8 @@ impl Full for SurfpoolFullRpc {
...
-            combined_slots.sort_unstable();
-            combined_slots.dedup();
-
-            if combined_slots.len() > MAX_SLOT_RANGE as usize {
-                combined_slots.truncate(MAX_SLOT_RANGE as usize);
-            }
-
-            Ok(combined_slots)
+            let slots: Vec<Slot> = (start_slot..=effective_end_slot).collect();
+            Ok(slots)
```
MicaiahReid: I'm pretty sure all of these changes to get_blocks should be reverted except the changes we made to the commitment config

crates/core/src/rpc/full.rs
```diff
@@ -1737,24 +1737,37 @@ impl Full for SurfpoolFullRpc {
...
+
         let svm_locker = match meta.get_svm_locker() {
             Ok(locker) => locker,
             Err(e) => return e.into(),
         };
 
-        Box::pin(async move {
-            Ok(svm_locker.with_svm_reader(|svm_reader| svm_reader.get_block_at_slot(slot)))
-        })
+        Box::pin(async move { Ok(svm_locker.get_block_local(&slot, &config).inner) })
```
MicaiahReid: This shouldn't be `get_block_local`, we need to `svm_locker.get_block`

crates/core/src/surfnet/locker.rs
```diff
@@ -1752,6 +1753,55 @@ impl SurfnetSvmLocker {
...
+        let first_local_slot = self.get_first_local_slot();
+
+        let result = match first_local_slot > *slot {
+            true => match remote_ctx {
+                Some(remote_client) => remote_client.get_block(slot, *config).await?,
+                None => return Err(SurfpoolError::slot_too_old(*slot)),
+            },
+            false => match self.get_block_local(slot, config).inner {
+                Some(block) => block,
+                None => return Err(SurfpoolError::slot_too_old(*slot)),
```
MicaiahReid, line 1771: If we don't have a block locally, that does not mean the slot provided to the user is too old. It means they're requesting a slot that doesn't exist yet. This shouldn't return an error, this should just return `None`

crates/core/src/error.rs
```diff
@@ -367,4 +368,16 @@ impl SurfpoolError {
...
         ))
     }
+
+    pub fn slot_too_old(slot: Slot) -> Self {
+        Self(Error::invalid_params(format!(
+            "Requested {slot} is before the first local slot, and no remote RPC was provided."
+        )))
+    }
+
+    pub fn get_block(e: solana_client::client_error::ClientError) -> Self {
```
MicaiahReid: Can we improve this error to say `"Failed to get block {} from remote: {}"`, passing in the block?

crates/core/src/surfnet/svm.rs
```diff
@@ -792,12 +799,12 @@ impl SurfnetSvm {
                 signatures: confirmed_signatures,
             },
         );
-
         if self.perf_samples.len() > 30 {
             self.perf_samples.pop_back();
         }
+        // Always add a perf_sample for the new block slot
```
```suggestion
```
don't think this comment is really adding anything helpful

crates/core/src/surfnet/svm.rs
```diff
@@ -937,91 +944,177 @@ impl SurfnetSvm {
...
-        let mut transactions = vec![];
-        for signature in &block.signatures {
-            let Some(TransactionWithStatusMeta(_slot, tx, _meta, _err)) = self
-                .transactions
-                .get(signature)
-                .map(|t| t.expect_processed().clone())
-            else {
-                continue;
-            };
+        let num_reward_partitions = config.rewards.unwrap_or(false);
```
MicaiahReid: This is a misleading variable name - it's not a number it's a bool. This should be used to decide if rewards are returned or not, not represent the number of reward partitions

crates/core/src/surfnet/svm.rs
```diff
@@ -937,91 +944,177 @@ impl SurfnetSvm {
...
+        let transaction_details = config
+            .transaction_details
+            .unwrap_or(TransactionDetails::Full);
+
+        let transactions = match transaction_details {
+            TransactionDetails::Full => {
+                let mut txs = vec![];
+                for signature in &block.signatures {
+                    let Some(tx_with_meta) = self
```
```suggestion
                    let Some(TransactionWithStatusMeta(_slot, tx, meta, err)) = self
```
We can revert this to be more concise, since we didn't actually end up using `tx_with_meta` for anything

crates/core/src/surfnet/svm.rs
```diff
@@ -937,91 +944,177 @@ impl SurfnetSvm {
...
+                            compute_units_consumed: OptionSerializer::Some(
+                                meta.compute_units_consumed,
+                            ),
+                            pre_token_balances: OptionSerializer::None, // todo
+                            post_token_balances: OptionSerializer::None, // todo
+                            rewards: OptionSerializer::None,            // todo
+                            loaded_addresses: OptionSerializer::None,   // todo
+                            pre_balances: vec![],
+                            post_balances: vec![],
+                            fee: 0,
```
MicaiahReid, line 1002: Can you create an issue to track our need to start storing TransactionStatusMeta data with every block?

crates/core/src/surfnet/svm.rs
```diff
@@ -937,91 +944,177 @@ impl SurfnetSvm {
...
+                                instructions,
+                                address_table_lookups,
+                            }),
+                        }),
+                        meta: Some(UiTransactionStatusMeta {
+                            err: err.clone(),
+                            status: err.map(|e| Err(e)).unwrap_or(Ok(())),
+                            inner_instructions,
+                            log_messages: OptionSerializer::Some(meta.logs.clone()),
+                            return_data: OptionSerializer::None,
```
MicaiahReid: The `meta` field has `return_data` that you should be converting and returning here

crates/core/src/surfnet/svm.rs
```diff
@@ -937,91 +944,177 @@ impl SurfnetSvm {
...
+                        .map(|t| EncodedTransactionWithStatusMeta {
+                            transaction: EncodedTransaction::LegacyBinary(t.to_string()),
+                            meta: None,
+                            version: None,
+                        })
+                        .collect(),
+                )
+            }
+            TransactionDetails::None => None,
+            _ => None,
```
MicaiahReid: Why are we ignoring the accounts case?

MicaiahReid (changes requested):

crates/core/src/surfnet/svm.rs
```diff
@@ -937,91 +944,245 @@ impl SurfnetSvm {
...
-        let mut transactions = vec![];
-        for signature in &block.signatures {
-            let Some(TransactionWithStatusMeta(_slot, tx, _meta, _err)) = self
-                .transactions
-                .get(signature)
-                .map(|t| t.expect_processed().clone())
-            else {
-                continue;
-            };
+        let rewards_partitions = config.rewards.unwrap_or(false);
```
MicaiahReid: I think my previous comment here was misunderstood, apologies if it was unclear.

This `config.rewards` field should not be used to represent the number of rewards partitions. Why are you converting this bool to a u64? That's a pretty good signal something is going wrong.

This `config.rewards` should be used to determing if the `rewards` field should be `Some(vec![])` or `None`

crates/core/src/surfnet/svm.rs
```diff
@@ -937,91 +944,245 @@ impl SurfnetSvm {
...
+                            })
+                            .collect(),
+                    );
+
+                    let return_data = match meta.return_data.data.is_empty() {
+                        true => OptionSerializer::None,
+                        false => OptionSerializer::Some(UiTransactionReturnData {
+                            program_id: meta.return_data.program_id.to_string(),
+                            data: (
+                                base64::encode(&meta.return_data.data),
```
MicaiahReid: Looks like this `encode` function is deprecated, can you use the recommended `Engine::decode` instead?

crates/core/src/surfnet/svm.rs
```diff
@@ -937,91 +944,245 @@ impl SurfnetSvm {
     ///
     /// # Arguments
     /// * `slot` - The slot number to retrieve the block for.
+    /// * `config` - The configuration for the block retrieval.
     ///
     /// # Returns
     /// `Some(UiConfirmedBlock)` if found, or `None` if not present.
-    pub fn get_block_at_slot(&self, slot: Slot) -> Option<UiConfirmedBlock> {
-        // Retrieve block
+    pub fn get_block_at_slot(
```
MicaiahReid, line 953: This function is quite huge and hard to read, with a good bit of duplicate code. Can we do a quick pass to send some of this logic to smaller functions? `to_ui_inner_instructions`, `to_ui_return_data`, `to_ui_message`, etc etc

crates/core/src/surfnet/svm.rs
```diff
@@ -937,91 +944,245 @@ impl SurfnetSvm {
...
+                        let message = UiMessage::Raw(UiRawMessage {
+                            header,
+                            account_keys,
+                            recent_blockhash,
+                            instructions: vec![],
+                            address_table_lookups,
+                        });
+
+                        let transaction = EncodedTransactionWithStatusMeta {
+                            transaction: EncodedTransaction::Json(UiTransaction {
```
MicaiahReid: This should return the `EncodedTransaction::Accounts` variant

lgalabru (approved):
thanks @BretasArthur1!

PR #223, opened 2025-07-15
fix(core): Fix metadata fields parsing

BretasArthur1:


crates/core/src/rpc/full.rs
```diff
@@ -3711,6 +3711,12 @@ mod tests {
 
         assert!(result.is_err());
 
+        {
+            let mut svm_writer = setup.context.svm_locker.0.write().await;
+            svm_writer.latest_epoch_info.absolute_slot = 110;
+            drop(svm_writer);
+        }
+
```
MicaiahReid: what's the purpose of this addition? seems unnecessary?

MicaiahReid (approved):

crates/core/src/rpc/utils.rs
```diff
@@ -206,18 +206,25 @@ pub fn transform_tx_metadata_to_ui_accounts(
     meta.inner_instructions
```
lgalabru, line 204: Should this method take ownership of `meta` and transform the data, instead of cloning?
So
```rust
pub fn transform_tx_metadata_to_ui_accounts(
    meta: TransactionMetadata,
) 
```
and then use `into_iter()`
?

lgalabru (approved):
Looking good, thanks @BretasArthur1 !

PR #224, opened 2025-07-16
fix(core): add transaction status with meta to transaction processing result

MicaiahReid:


crates/core/src/rpc/full.rs
```diff
@@ -1630,9 +1630,11 @@ impl Full for SurfpoolFullRpc {
         };
 
         Box::pin(async move {
-            let pubkeys = svm_locker
-                .get_pubkeys_from_message(&remote_ctx, &unsanitized_tx.message)
+            let loaded_addresses = svm_locker
+                .get_loaded_address_from_message(&remote_ctx, &unsanitized_tx.message)
```
lgalabru: `get_loaded_addresses` ? 💅

crates/core/src/tests/helpers.rs
```diff
@@ -106,28 +105,14 @@ where
...
-                        tx,
-                        convert_transaction_metadata_from_canonical(&e.meta),
-                        Some(e.err),
-                    ))),
-                ),
-            };
+            let _ = self
+                .context
+                .svm_locker
+                .process_transaction(&None, tx.clone(), status_tx.clone(), true)
```
lgalabru, line 113: Is this clone required?

crates/core/src/types.rs
```diff
@@ -31,34 +41,364 @@ impl SurfnetTransactionStatus {
...
+                        .map(|(((i, a), mint), token_program)| TransactionTokenBalance {
+                            account_index: *i as u8,
+                            mint: a.mint.to_string(),
+                            ui_token_amount: UiTokenAmount {
+                                ui_amount: Some(a.amount as f64 / 10f64.powi(mint.decimals as i32)),
+                                decimals: mint.decimals,
+                                amount: a.amount.to_string(),
+                                ui_amount_string: format!(
+                                    "{:.2}",
+                                    a.amount as f64 / 10f64.powi(mint.decimals as i32)
```
lgalabru: Could we maybe have a little helper for those?

crates/core/src/rpc/full.rs
BretasArthur1: 😂

MicaiahReid: Done! bcc2dc13c77f8bb9d20815e409b627bf3976bfc8

crates/core/src/tests/helpers.rs
MicaiahReid: Sir this is a test

MicaiahReid: But I did remove one clone in a different call to `process_transaction` just for you 😘 
fadfbe798f02782832769b7516678f626204deeb

crates/core/src/types.rs
MicaiahReid: Done: 2b9c4af796318fb7c5f9e9793cd13af926551a2c

crates/core/src/surfnet/svm.rs
```diff
@@ -218,30 +209,55 @@ impl SurfnetSvm {
...
-                ))),
+                    transaction: tx.clone(),
+                    meta: TransactionStatusMeta {
+                        status: Ok(()),
+                        fee: 5000,
+                        pre_balances: vec![account.lamports.saturating_sub(lamports)],
+                        post_balances: vec![account.lamports],
+                        inner_instructions: None,
+                        log_messages: Some(tx_result.logs.clone()),
+                        pre_token_balances: None,
```
BretasArthur1: the default behavior of these fields when empty are different for e.g pre_token_balances: Some(vec![]), lets work with these conventional so we make things standardized

BretasArthur1 (approved):

crates/core/src/surfnet/svm.rs
MicaiahReid: Hmm, it's always `Some(vec![])`? Even when there are no token accounts involved in the tx?

crates/core/src/surfnet/locker.rs
```diff
@@ -2002,3 +2075,26 @@ fn update_programdata_account(
         ));
     }
 }
+
+pub fn format_ui_amount_string(amount: u64, decimals: u8) -> String {
+    let formatted_amount = if decimals > 0 {
+        let divisor = 10u64.pow(decimals as u32);
```
BretasArthur1, line 2081: I seems that this logic is duplicated on the function below why not calling it here and remove the duplicate logic

crates/core/src/surfnet/svm.rs
MicaiahReid: Done: f9e89296180e7038611c5d7da13634b7e3841655

crates/core/src/surfnet/locker.rs
MicaiahReid: The logic is a bit different to account for string vs number, doesn't seem worth the effort

PR #225, opened 2025-07-16
fix(core): index token accounts to always include additional data

MicaiahReid:


crates/core/src/surfnet/locker.rs
```diff
@@ -504,6 +469,18 @@ impl SurfnetSvmLocker {
...
         Ok(results)
     }
+
+    pub fn account_to_rpc_keyed_account<T: ReadableAccount + Send + Sync>(
+        &self,
+        pubkey: &Pubkey,
+        account: &T,
+        config: &RpcAccountInfoConfig,
+        token_mint: Option<Pubkey>,
+    ) -> RpcKeyedAccount {
```
lgalabru, line 479: I'm not a huge fan of having / re-inforcing inter-dependencies between surfnets and the RPC layer

lgalabru (approved):
feels like a whole lot of fixes, thanks for this upgrade @MicaiahReid 🙏

PR #227, opened 2025-07-17
fix(core): fix types mismatch to expected on explorer

BretasArthur1:


BretasArthur1:
before this changes we're seeing this searching to a airdrop tx on explorer 

image
 
 
 
 after
 
image

crates/core/src/surfnet/svm.rs
```diff
@@ -245,13 +251,17 @@ impl SurfnetSvm {
                     meta: TransactionStatusMeta {
                         status: Ok(()),
                         fee: 5000,
-                        pre_balances: vec![account.lamports.saturating_sub(lamports)],
-                        post_balances: vec![account.lamports],
-                        inner_instructions: None,
+                        pre_balances: vec![
+                            account.lamports.saturating_sub(lamports),
+                            0,
```
MicaiahReid: What is this account that starts with 0 and ends with `lamports`?

BretasArthur1: In this cause I'm assuming that is the first airdrop of the account

BretasArthur1: adjust it with a5dba99a301e6990fc68dd408ab0a2c614b81745

MicaiahReid: So we have three accounts involved: the system program, the sender, and the recipient. I don't know for sure what the order is. But sadly, the airdrop keypair isn't exposed via litesvm, so we don't know what account to query here.

Maybe that will change in the future, I will make an issue for this

MicaiahReid (approved):

PR #231, opened 2025-07-19
fix(core): index spl 2022 tokens with state with other token mints

MicaiahReid:


MicaiahReid:
> Should we be matching to exhaust that variant?

Sorry I'm not following, what variant?

PR #241, opened 2025-07-22
fix: improve TUI & logs Devex

lgalabru:


crates/cli/src/runbook/mod.rs
```diff
@@ -280,6 +277,7 @@ pub async fn execute_runbook(
             &simnet_events_tx,
             cmd.output_json,
         );
+        let _ = simnet_events_tx.send(SimnetEvent::RunbookCompleted(runbook_id.clone()));
```
BretasArthur1: if we're cloning twice here, should be only the first one no? is there any problem if the second one can take ownership of it?

crates/cli/src/tui/simnet.rs
```diff
@@ -215,16 +216,23 @@ fn run_app<B: Backend>(terminal: &mut Terminal<B>, mut app: App) -> io::Result<(
...
-            let blockhash = client.get_latest_blockhash().unwrap();
-            let transaction = Transaction::new(&[keypair], message, blockhash);
-            let _ = client.send_transaction(&transaction).unwrap();
+            match client.get_latest_blockhash() {
+                Ok(blockhash) => {
+                    let transaction = Transaction::new(&[keypair], message, blockhash);
+                    let _ = client.send_transaction(&transaction).unwrap();
+                }
+                Err(_e) => {
+                    // print error
```
BretasArthur1: we should add this print no?

crates/cli/src/runbook/mod.rs
lgalabru: Addressed with latest

crates/cli/src/tui/simnet.rs
lgalabru: We're actually already getting logs from another code path!

crates/cli/src/tui/simnet.rs
```diff
@@ -422,6 +465,12 @@ fn run_app<B: Backend>(terminal: &mut Terminal<B>, mut app: App) -> io::Result<(
     Ok(())
 }
 
+// fn add_event(app: &mut App, event: (EventType, DateTime<Local>, String)) {
+//     app.events
+//         .push_back(event);
+//     app.scroll_state.position(app.events.len() + 1)
+// }
+
```
MicaiahReid:
```suggestion
```

crates/cli/src/tui/simnet.rs
```diff
@@ -506,15 +555,15 @@ fn render_stats(f: &mut Frame, app: &mut App, area: Rect) {
             vec![
                 Line::from(vec![
                     Span::styled("۬", app.colors.white),
-                    Span::styled("Surfnet   ", app.colors.gray),
+                    Span::styled("Surfnet   ", app.colors.light_gray),
```
MicaiahReid, line 552: I'm confused, cause in my TUI I'm not seeing the Surfnet or Provider url show up 🤔 

Image

BretasArthur1 (approved):

MicaiahReid (approved):
Love it, looks and feels so much better!

PR #243, opened 2025-07-23
feat: implements logs subscribe / unsubscribe websocket endpoint

lgalabru:


crates/core/src/surfnet/svm.rs
```diff
@@ -1207,6 +1202,38 @@ impl SurfnetSvm {
         self.slot_subscriptions
             .retain(|tx| tx.send(SlotInfo { slot, parent, root }).is_ok());
     }
+
+    pub fn subscribe_for_logs_updates(
```
MicaiahReid, line 1217: We may have already set a precedence for this with the other WS methods, idk, but are we ever removing from the list of `log_subscriptions` here? I know a user can unsubscribe and the WS channel is disconnected, but do we clean up this list when that happens?

crates/core/src/surfnet/svm.rs
```diff
@@ -1207,6 +1202,38 @@ impl SurfnetSvm {
...
+
+    pub fn notify_logs_subscribers(
+        &mut self,
+        signature: &Signature,
+        err: Option<TransactionError>,
+        logs: Vec<String>,
+        commitment_level: CommitmentLevel,
+    ) {
+        self.updated_at = Utc::now().timestamp_millis() as u64;
+        for (expected_level, filter, tx) in self.logs_subscriptions.iter() {
```
MicaiahReid, line 1237: Can we implement this filter? (specifically the `mentions` field)

crates/core/src/surfnet/svm.rs
```diff
@@ -712,6 +715,17 @@ impl SurfnetSvm {
...
+                self.transactions.get(&signature)
+            else {
+                continue;
+            };
+            self.notify_logs_subscribers(
+                &signature,
+                None,
+                tx_data.meta.log_messages.clone().unwrap_or(vec![]),
+                CommitmentLevel::Confirmed,
+            );
```
MicaiahReid, line 728: Do we need to notify when Finalized as well? (below in the `finalize_transactions` fn)

crates/core/src/surfnet/svm.rs
```diff
@@ -1207,6 +1202,38 @@ impl SurfnetSvm {
...
+    ) {
+        self.updated_at = Utc::now().timestamp_millis() as u64;
+        for (expected_level, filter, tx) in self.logs_subscriptions.iter() {
+            if expected_level.eq(&commitment_level) {
+                let message = RpcLogsResponse {
+                    signature: signature.to_string(),
+                    err: err.clone(),
+                    logs: logs.clone(),
+                };
+                let _ = tx.send((self.get_latest_absolute_slot(), message));
```
MicaiahReid, line 1244: Maybe a simple solution for my above question about removing from the `logs_subscriptions` vec is if this `tx.send` fails, we consider the `rx` disconnected and remove? 

Would be easy to implement, but could potentially yield some hard to track "what happened to my subscription" bugs ¯\_(ツ)_/¯

MicaiahReid (changes requested):

crates/core/src/surfnet/svm.rs
lgalabru: tracking this here: https://github.com/txtx/surfpool/issues/242
I need to spend some time understanding how that filtering is being done in the canonical implementation.

crates/core/src/surfnet/svm.rs
lgalabru: Good idea. Let's do this in a subsequent PR and prioritize - starting with `signatureSubscribe` is probably more important (heavily used and leaking in cloud services).

lgalabru: https://github.com/txtx/surfpool/issues/247

crates/core/src/surfnet/svm.rs
lgalabru: Good catch, I had it in a local stash!

MicaiahReid (approved):
Thanks for making those issues and fixes! looks good!

PR #245, opened 2025-07-23
feat(core): add transaction profile retrieval by UUID or signature

BretasArthur1:
- Introduce `get_transaction_profile` RPC to:
  - Support lookup by both transaction signature and simulation UUID.
  - Reconstruct and store profile results for executed transactions if missing.
  - Cleanly handle UUID lookups without reconstruction.
- All state mutation and UUID management is now handled in the locker, not the RPC.
- Ensured all responses include the UUID in the `ProfileResult` when available.


crates/core/src/surfnet/locker.rs
```diff
@@ -1506,14 +1506,40 @@ impl SurfnetSvmLocker {
             capture
         };
 
+        // Determine if this is a simulation (no signature) or execution (has signature)
+        // For now, always generate a UUID and store in simulated_transaction_profiles
+        let uuid = uuid::Uuid::new_v4();
+        let mut is_execution = false;
```
lgalabru: I think I'd call this boolean `is_simulation` instead (and invert the logic of the test below)

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -592,6 +602,7 @@ impl SvmTricksRpc for SurfnetCheatcodesRpc {
     }
 
     fn profile_transaction(
+        //TODO: surfnet_getTransactionProfilesByTag
```
MicaiahReid:
```suggestion
```

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -743,4 +754,82 @@ impl SvmTricksRpc for SurfnetCheatcodesRpc {
             })
         })
     }
+
+    fn get_transaction_profile(
+        &self,
+        meta: Self::Metadata,
+        transaction_signature: UuidOrSignature,
```
MicaiahReid: Can we rename this guy to `signature_or_uuid`?

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -743,4 +754,82 @@ impl SvmTricksRpc for SurfnetCheatcodesRpc {
...
+
+            match profile_result {
+                Some(profile_result) => {
+                    let context_slot = if profile_result.slot >= first_slot
+                        && profile_result.slot <= latest_slot
+                    {
+                        profile_result.slot
+                    } else {
+                        latest_slot
+                    };
```
MicaiahReid: This logic seems unnecessary - how would our profile result be after the latest slot or before the first slot? Seems like we could always just do `profile_result.slot`?

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -743,4 +754,82 @@ impl SvmTricksRpc for SurfnetCheatcodesRpc {
...
+                            {
+                                profile_result = Some(prof);
+                            }
+                        }
+                    }
+                    UuidOrSignature::Uuid(_) => {
+                        // No reconstruction for UUID, error will be returned below if not found
+                    }
+                }
+            }
```
MicaiahReid: This is interesting, but I'm not sure if we should have it. If I'm getting a profile result for a signature and we don't have results, I'd just expect it to return `None`, not simulate.

MicaiahReid: Thoughts @lgalabru?

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -743,4 +754,82 @@ impl SvmTricksRpc for SurfnetCheatcodesRpc {
...
+            Err(e) => return e.into(),
+        };
+        let first_slot = svm_locker
+            .get_first_local_slot()
+            .unwrap_or_else(|| svm_locker.with_svm_reader(|svm| svm.get_latest_absolute_slot()));
+        let latest_slot = svm_locker.with_svm_reader(|svm| svm.get_latest_absolute_slot());
+
+        Box::pin(async move {
+            let mut profile_result =
+                svm_locker.with_svm_reader(|svm| match &transaction_signature {
```
MicaiahReid: Rather than `.with_svm_reader`, can we add a function to the locker to `get_transaction_profile`?

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -743,4 +754,82 @@ impl SvmTricksRpc for SurfnetCheatcodesRpc {
...
+                        latest_slot
+                    };
+                    Ok(RpcResponse {
+                        context: RpcResponseContext::new(context_slot),
+                        value: profile_result,
+                    })
+                }
+                None => match &transaction_signature {
+                    UuidOrSignature::Signature(signature) => {
+                        Err(SurfpoolError::transaction_not_found_in_svm(signature).into())
```
MicaiahReid: I don't think we should be returning an error, just a `None`. This is the same as how https://solana.com/docs/rpc/http/getsignaturestatuses returns null if the signature isn't found

crates/core/src/surfnet/locker.rs
MicaiahReid: Agreed

MicaiahReid: Or just a one-liner `let is_simulation = signature == solana_signature::Signature::default()`

crates/types/src/types.rs
```diff
@@ -104,6 +104,8 @@ pub struct ComputeUnitsEstimationResult {
 pub struct ProfileResult {
     pub compute_units: ComputeUnitsEstimationResult,
     pub state: ProfileState,
+    pub slot: u64,
+    pub uuid: Option<Uuid>,
```
MicaiahReid, line 110: Is this field giving us anything useful? Seems unnecessary. Or if it is necessary, I'd imagine it should be `uuid_or_signature: UuidOrSignature`

crates/core/src/surfnet/locker.rs
```diff
@@ -1506,14 +1506,40 @@ impl SurfnetSvmLocker {
...
+            svm.simulated_transaction_profiles
+                .insert(uuid, profile_result.clone());
+            if is_execution {
+                svm.executed_transaction_profiles
+                    .insert(signature, profile_result.clone());
+                svm.profile_tag_map
+                    .entry(signature.to_string())
+                    .or_default()
+                    .push(UuidOrSignature::Uuid(uuid));
+            }
```
MicaiahReid: This logic seems off. I thought it should be:
1. If simulation we insert into `simulated_transaction_profiles` and _not_ `executed_transaction_profiles`.
2. If not simulation, we insert into `executed_transaction_profiles` and _not_ `simulated_transaction_profiles`.
3. We _always_ insert into `profile_tag_map`

crates/core/src/surfnet/svm.rs
```diff
@@ -83,6 +84,9 @@ pub struct SurfnetSvm {
     pub signature_subscriptions: HashMap<Signature, Vec<SignatureSubscriptionData>>,
     pub account_subscriptions: AccountSubscriptionData,
     pub slot_subscriptions: Vec<Sender<SlotInfo>>,
+    pub profile_tag_map: HashMap<String, Vec<UuidOrSignature>>,
+    pub simulated_transaction_profiles: HashMap<Uuid, ProfileResult>,
+    pub executed_transaction_profiles: HashMap<Signature, ProfileResult>,
     pub tagged_profiling_results: HashMap<String, Vec<ProfileResult>>,
```
MicaiahReid: Shouldn't we be removing `tagged_profiling_results` and retrieving via `profile_tag_map` instead?

MicaiahReid (changes requested):
My understanding of this was that we want to be profiling transactions whenever `send_transaction` is called, but we're not doing that here. Is that intended for a future PR?

PR #254, opened 2025-07-24
feat: add event to display both rpc/ws endpoints and datasource provi…

BretasArthur1:
…der on surfpool startup:

### DIFF
--- a/crates/cli/src/tui/simnet.rs
+++ b/crates/cli/src/tui/simnet.rs
@@ -120,7 +120,27 @@ impl App {
                 transaction_count: None,
             },
             successful_transactions: 0,
-            events: vec![],
+            events: {
+                let mut events = vec![];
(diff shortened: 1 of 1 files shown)


crates/cli/src/tui/simnet.rs
```diff
@@ -120,7 +120,27 @@ impl App {
...
+                    DisplayedUrl::Datasource(config) | DisplayedUrl::Studio(config) => (
+                        config.rpc_url.clone(),
+                        config.ws_url.clone(),
+                        config.rpc_datasource_url.clone(),
+                    ),
+                };
+                events.push((
+                    EventType::Success,
+                    Local::now(),
+                    format!("Started localvalidator [RPC: {rpc_url}, WS: {ws_url}]"),
```
lgalabru:
```suggestion
                    format!("Started local validator [RPC: {rpc_url}, WS: {ws_url}]"),
```

PR #255, opened 2025-07-25
feat(core): add IDL registration and retrieval RPC methods

BretasArthur1:


crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -919,15 +979,17 @@ impl SvmTricksRpc for SurfnetCheatcodesRpc {
...
         program_id: String,
+        slot: Option<Slot>,
     ) -> BoxFuture<Result<RpcResponse<Option<Idl>>>> {
         let svm_locker = match meta.get_svm_locker() {
             Ok(locker) => locker,
             Err(e) => return e.into(),
         };
         Box::pin(async move {
-            let idl = svm_locker.get_idl(&program_id)?;
+            let idl = svm_locker.get_idl(&Pubkey::from_str_const(&program_id), slot);
```
MicaiahReid: Whenever we recieve a pubkey as a string in an RPC request, we have a function to check if it's valid - can you use that?

`Pubkey::from_str_const` will panic if the user passes an invalid pubkey, so we can only use that for data we know will be a valid pubkey

crates/core/src/surfnet/locker.rs
```diff
@@ -1638,15 +1638,19 @@ impl SurfnetSvmLocker {
...
-            Ok::<(), SurfpoolError>(())
-        })
+    pub fn register_idl(&self, idl: Idl, slot: Option<Slot>) {
+        self.with_svm_writer(|svm| svm.register_idl(idl, slot))
     }
 
-    pub fn get_idl(&self, address: &str) -> SurfpoolResult<Option<Idl>> {
-        self.with_svm_reader(|svm| Ok(svm.idl_registry.get(address).cloned()))
+    pub fn get_idl(&self, address: &Pubkey, slot: Option<Slot>) -> Option<Idl> {
+        self.with_svm_reader(|svm| {
```
MicaiahReid: Throughout this file, we try to keep the naming convention:
 - `self.with_svm_reader(|svm_reader|`
 - `self.with_svm_writer(|svm_writer|`
Can we do that with these methods? We want to keep track of when we're using reader/writer, and this is a way to help with that

crates/core/src/surfnet/locker.rs
```diff
@@ -1644,9 +1644,12 @@ impl SurfnetSvmLocker {
 
     pub fn get_idl(&self, address: &Pubkey, slot: Option<Slot>) -> Option<Idl> {
         self.with_svm_reader(|svm_reader| {
+            let current_slot = svm_reader.get_latest_absolute_slot();
+            let query_slot = slot.unwrap_or(current_slot);
```
```
let query_slot = slot.unwrap_or_else(|| svm_reader.get_latest_absolute_slot());
```
:)

crates/core/src/tests/integration.rs
```diff
@@ -1044,3 +1051,462 @@ async fn test_surfnet_estimate_compute_units_with_state_snapshots() {
...
+
+    let non_existent_sig_result = non_existent_sig_response.unwrap().value;
+    assert!(
+        non_existent_sig_result.is_none(),
+        "Non-existent signature should return None"
+    );
+
+    // Test 7: Verify context slot is correct
+    println!("Verifying response context");
+    let uuid_response_with_context = rpc_server
```
MicaiahReid: We're not actually verifying any slot here, is that intentional?

crates/core/src/tests/integration.rs
```diff
@@ -1044,3 +1051,462 @@ async fn test_surfnet_estimate_compute_units_with_state_snapshots() {
...
+        retrieved_idl.unwrap(),
+        idl,
+        "Retrieved IDL should match registered IDL"
+    );
+
+    println!("All IDL registration and retrieval tests passed successfully!");
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_register_and_get_same_idl_with_different_slots() {
```
MicaiahReid, line 1378: For this test, we should be registering a _different_ IDL (but with the same program ID) at each slot, so we can verify that the _correct IDL_ is retrieved.

crates/core/src/tests/integration.rs
```diff
@@ -1044,3 +1051,462 @@ async fn test_surfnet_estimate_compute_units_with_state_snapshots() {
...
+
+    assert!(
+        register_response.is_ok(),
+        "Register IDL failed at slot {}: {:?}",
+        future_slot,
+        register_response.err()
+    );
+
+    // Step 3: Test slot-based isolation
+    let test_slots = [current_slot, current_slot + 50, future_slot + 50];
```
MicaiahReid: Can we add `0` to this list of test slots?

If the user fetches the IDL _before_ one was registered, we should be returning None

crates/types/src/types.rs
```diff
@@ -643,3 +644,28 @@ pub enum DataIndexingCommand {
...
+
+impl PartialOrd for VersionedIdl {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for VersionedIdl {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.0.cmp(&other.0) // or reverse it if you want min-heap behavior
```
```suggestion
        self.0.cmp(&other.0)
```

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -378,12 +379,197 @@ pub trait SvmTricksRpc {
...
+    ///   "jsonrpc": "2.0",
+    ///   "context": {
+    ///     "slot": 355684457,
+    ///     "apiVersion": "2.2.2"
+    ///   },
+    ///   "value": { /* ...IDL object... */ },
+    ///   "id": 1
+    /// }
+    /// ```
+    #[rpc(meta, name = "surfnet_getIdl")]
```
lgalabru: `surfnet_getIdl` or `surfnet_getActiveIdl`?

crates/core/src/tests/assets/idl_v1.json
```diff
@@ -0,0 +1,394 @@
+{
```
lgalabru, line 1: these kind of files are generally referred to as `fixtures`, vs `assets`

BretasArthur1: Ok

crates/core/src/rpc/surfnet_cheatcodes.rs
BretasArthur1: It's get Idl because you can query for a specific one using the slot

BretasArthur1:
was a good one to test🫡

crates/core/src/rpc/surfnet_cheatcodes.rs
lgalabru: yeah we're not providing an Idl identifier, we're querying using a property. maybe `getIdlBySlot`? Idk, but it feels like something is missing here.

BretasArthur1: Ok, lets discuss it in tmrw's daily

PR #256, opened 2025-07-28
fix: allow multiple collections per workspace

lgalabru:


crates/cli/src/cli/mod.rs
```diff
@@ -188,10 +188,10 @@ pub struct StartSimnet {
     /// List of geyser plugins to load
     #[arg(long = "geyser-plugin-config", short = 'g')]
     pub plugin_config_path: Vec<String>,
-    /// Path to subgraph's sqlite database (default: :memory:)
-    #[arg(long = "subgraph-database-path", short = 'd')]
-    pub subgraph_database_path: Option<String>,
+    /// Subgraph database connection URL (default to sqlite ":memory:", also supports postgres: "postgres://postgres:posgres@e127.0.0.1:5432/surfpool")
     /// Disable Studio (default: false)
+    #[arg(long = "subgraph-db", short = 'd')]
+    pub subgraph_db: Option<String>,
```
MicaiahReid: I think a merge artifact, but this command got put under the docs for a different flag

MicaiahReid (changes requested):
Still want to test this one with latest txtx + this branch

crates/cli/src/cli/mod.rs
lgalabru: I had a conflict but I think this would be the right flag, as a db can be a local sqlite file, but also a remote postgres db

MicaiahReid: I think you're misunderstanding - you have:
```
/// Subgraph datab ... (comment for the `subgraph_db` opt)
/// Disable Studio (default: false) ... (comment for the `no_studio` opt
#[arg(long = "subgraph-db", short = 'd')]
pub subgraph_db: Option<String>,
#[clap(long = "no-studio")]
pub no_studio: bool,
```
you need to move the Disable studio comment

lgalabru: ah right, thanks.

PR #257, opened 2025-07-28
fix: ability to read transaction v0 from websocket

lgalabru:


crates/core/src/rpc/ws.rs
```diff
@@ -666,6 +666,12 @@ impl Rpc for SurfpoolWsRpc {
             }
         };
         let config = config.unwrap_or_default();
+        let rpc_transaction_config = RpcTransactionConfig {
+            encoding: Some(UiTransactionEncoding::Json),
```
MicaiahReid, line 670: Is there a chance we actually want `JsonParsed`? That's the one that parses according to the owner and returns a pretty json object with the data

crates/core/src/runloops/mod.rs
```diff
@@ -355,7 +355,7 @@ fn start_geyser_runloop(
                         ) {
                             Ok(tx) => tx,
                             Err(e) => {
-                                let _ = simnet_events_tx.send(SimnetEvent::error(format!("Failed to notify Geyser plugin of new transaction: failed to serialize transaction: {:?}", e)));
+                                let _ = simnet_events_tx.send(SimnetEvent::warn(format!("Failed to serialize transaction: {:?}", e)));
```
MicaiahReid: I agree with the warn, but this message seems worse to me. It's much less data, right? And if I'm not actively using a Geyser plugin I know I can ignore this warning, but if I see `Failed to serialize transaction` I'll be wondering "was my transaction processed?"

crates/core/src/rpc/ws.rs
lgalabru: After testing I can confirm that `Json` is what is needed here, both for websocket and `getTransaction`.

crates/core/src/runloops/mod.rs
lgalabru: Addressed with latest. we'll only show the warning "unable to index transaction' for setup where indexing is actually active.

PR #260, opened 2025-07-30
feat: instruction profiling

BretasArthur1:


crates/types/src/types.rs
```diff
@@ -102,34 +108,54 @@ pub struct ComputeUnitsEstimationResult {
...
+pub struct KeyedProfileResult {
+    pub slot: u64,
+    pub key: UuidOrSignature,
+    pub profile: TransactionProfileResult,
+}
+
+// Tx with ixs [A, B, C]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct TransactionProfileResult {
+    pub instruction_profiles: Option<ProfiledInstructionsMap>, // Contains { [A] : {..result}, [A+B]: {..result} }
```
lgalabru: Can we pull the `indexmap` crate and use an `IndexMap` here?
This would allow us to get ordered map, which seems important in this context.

PR #261, opened 2025-07-30
fixed logsSubscribe and added a dockerfile

spacemandev-git:
small fix to logsSubscribe to use CommitmentConfig instead of CommitmentLevel to match solana rpc spec


Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+    rm -rf /var/lib/apt/lists/*
+
+# Clone and install surfpool from source
+COPY . ./surfpool
+
+RUN cargo install --path ./surfpool/crates/cli --locked --force
+
+# Copy the surfpool manifest file
+COPY Surfpool.toml ./Surfpool.toml
+
```
lgalabru: We don't have the concept of manifests within Surfpool just yet :)
```suggestion
```

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+# Expose common Solana ports
+EXPOSE 8899 8900 8080
```
lgalabru: we're not using 8080 but we're using `8488`
```suggestion
EXPOSE 8899 8900 8488
```

Surfpool.toml
```diff
@@ -0,0 +1,13 @@
+[surfpool]
```
lgalabru: Could we get rid of this file?

spacemandev-git:
should be clean now

PR #264, opened 2025-07-31
feat: introduce `surfnet_timeTravel` cheat code

lgalabru:


crates/core/src/helpers/time_travel.rs
```diff
@@ -0,0 +1,494 @@
...
+/// A `Result` containing either a new `Clock` object representing the target time state,
+/// or a `TimeTravelError` if the operation is invalid
+pub fn calculate_absolute_timestamp_clock(
+    timestamp_target: u64,
+    current_updated_at: u64,
+    slot_time: u64,
+    epoch_info: &EpochInfo,
+) -> Result<Clock, TimeTravelError> {
+    // Ensure the timestamp is in the future
+    if timestamp_target < current_updated_at {
```
MicaiahReid, line 30: Why is there this restriction? A user may want to go back in time, right?

crates/core/src/helpers/time_travel.rs
```diff
@@ -0,0 +1,494 @@
...
+/// or a `TimeTravelError` if the operation is invalid
+pub fn calculate_absolute_slot_clock(
+    new_absolute_slot: u64,
+    current_absolute_slot: u64,
+    current_updated_at: u64,
+    slot_time: u64,
+    epoch_info: &EpochInfo,
+) -> Result<Clock, TimeTravelError> {
+    // Ensure the slot is in the future
+    if new_absolute_slot < current_absolute_slot {
```
MicaiahReid, line 88: Same question about going back in time

crates/core/src/helpers/time_travel.rs
```diff
@@ -0,0 +1,494 @@
...
+pub fn calculate_absolute_epoch_clock(
+    new_epoch: u64,
+    current_epoch: u64,
+    current_absolute_slot: u64,
+    current_updated_at: u64,
+    slot_time: u64,
+    epoch_info: &EpochInfo,
+) -> Result<Clock, TimeTravelError> {
+    // Ensure the epoch is in the future
+    if new_epoch < current_epoch {
```
MicaiahReid, line 132: Same question about going back in time

crates/types/src/types.rs
```diff
@@ -383,7 +390,7 @@ impl Default for SimnetConfig {
     fn default() -> Self {
         Self {
             remote_rpc_url: DEFAULT_RPC_URL.to_string(),
-            slot_time: 0,
+            slot_time: 400, // Default to 400ms to match CLI default
```
```suggestion
            slot_time: DEFAULT_SLOT_TIME_MS, // Default to 400ms to match CLI default
```

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -1081,4 +1195,49 @@ impl SvmTricksRpc for SurfnetCheatcodesRpc {
...
+        let clock_update: Clock =
+            calculate_time_travel_clock(&behavior, updated_at, slot_time, &epoch_info)
+                .map_err(|e| Error::invalid_params(e.to_string()))?;
+
+        epoch_info.slot_index = clock_update.slot;
+        epoch_info.epoch = clock_update.epoch;
+        epoch_info.absolute_slot =
+            clock_update.slot + clock_update.epoch * epoch_info.slots_in_epoch;
+
+        let _ = surfnet_command_tx.send(SimnetCommand::UpdateInternalClock(clock_update));
```
MicaiahReid: We've been trying to minimize logic or steps in the RPC methods, instead putting them in the `locker` file - you think that's worth moving out of here?

crates/core/src/helpers/time_travel.rs
lgalabru: Getting back in time feels like a footgun to me. 
In my mind, developers need this feature to test behavior on future state (expiration dates, vesting schedules, etc), not past state

crates/core/src/helpers/time_travel.rs
lgalabru: Same response

crates/core/src/helpers/time_travel.rs
lgalabru: Same response

crates/core/src/helpers/time_travel.rs
lgalabru: We could relax in the future if someone can make a strong argument for it.

crates/core/src/rpc/surfnet_cheatcodes.rs
lgalabru: We don't need to lock here?

crates/core/src/helpers/time_travel.rs
MicaiahReid: I believe this was a feature in the EVM simulators, but I'm cool with deferring

crates/core/src/rpc/surfnet_cheatcodes.rs
lgalabru: Addressed with https://github.com/txtx/surfpool/pull/264/commits/abb41ebebf28ce3604cc374a72d12c86e1482894

PR #265, opened 2025-08-01
feat(core): Add RPC method to retrieve local signatures

BretasArthur1:


lgalabru:
Tested locally, noticed that the encoding was off and pushed a fix.

Screenshot 2025-08-03 at 20 29 16

Screenshot 2025-08-03 at 20 29 19

Issue #238, opened 2025-07-22
Allow running surfpool offline (no remote RPC)


MicaiahReid:

lgalabru:
Running offline should be explicitly required by developers:
```console
surfpool start --offline
```

PR #267, opened 2025-08-04
feat(cli): Add offline mode support

BretasArthur1:


BretasArthur1:
close #238

crates/cli/src/cli/simnet/mod.rs
```diff
@@ -140,6 +140,7 @@ pub async fn handle_start_local_surfnet_command(
             Ok(SimnetEvent::Aborted(error)) => return Err(error),
             Ok(SimnetEvent::Shutdown) => return Ok(()),
             Ok(SimnetEvent::Connected(_)) => break,
+            Ok(SimnetEvent::Ready) => break,
```
MicaiahReid, line 143: This may have an impact on online mode - if we were previously waiting to be fully connected to proceed, this may not be safe to change.

Is it possible, in this loop, if we hit `Ready`, check to see if there is a data source url. If there is, continue and wait for connected, if there isn't, break

crates/cli/src/cli/mod.rs
```diff
@@ -254,17 +257,21 @@ impl StartSimnet {
...
-        let remote_rpc_url = match &self.network {
-            Some(NetworkType::Mainnet) => DEFAULT_RPC_URL.to_string(),
-            Some(NetworkType::Devnet) => DEVNET_RPC_URL.to_string(),
-            Some(NetworkType::Testnet) => TESTNET_RPC_URL.to_string(),
-            None => match self.rpc_url {
-                Some(ref rpc_url) => rpc_url.clone(),
-                None => match env::var("SURFPOOL_DATASOURCE_RPC_URL") {
-                    Ok(value) => value,
-                    _ => DEFAULT_RPC_URL.to_string(),
+        let remote_rpc_url = if self.offline {
```
MicaiahReid: Instead of making offline an empty string, can we make the `remote_rpc_url` field an optional?

MicaiahReid (changes requested):

crates/cli/src/cli/simnet/mod.rs
MicaiahReid: Discussed, this is safe, thanks @BretasArthur1 !

Issue #278, opened 2025-08-08
Studio: Refreshing the page at `/subgraphs` does not load subgraphs page


MicaiahReid:
Reproduction:
 1. Load studio
 2. Click `Data Indexing`
 3. Refresh page
 4. Note that you aren't on the data indexing page anymore
 5. Click `Data Indexing` again
 6. Noe you _still_ aren't on the data indexing page anymore

PR #279, opened 2025-08-11
fix(studio): appropriately route requests to correct page

MicaiahReid:
Fixes: #278


crates/studio/src/lib.rs
```diff
@@ -8,11 +8,19 @@ pub static ASSETS: Dir<'_> = include_dir!("$OUT_DIR/surfpool-studio-ui");
 #[get("/{_:.*}")]
 async fn serve_studio_static_files(req: HttpRequest) -> impl Responder {
     let path = req.path().trim_start_matches('/');
```
BretasArthur1, line 10: can we match all of these?

PR #280, opened 2025-08-13
feat(core): Record failling transactions on the svm

BretasArthur1:


crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -1134,52 +1134,54 @@ impl SurfnetCheatcodes for SurfnetCheatcodesRpc {
...
+            String,
+            Slot,
+            Option<solana_transaction_error::TransactionError>,
+            Vec<String>,
+        )> = svm_locker.with_svm_reader(|svm_reader| {
+            svm_reader
+                .transactions
+                .iter()
+                .map(|(sig, status)| {
+                    let twsm = status.expect_processed();
```
MicaiahReid: I don't love this `twsm` name - you have no idea what it is while just reading the code. A longer, descriptive name is preferred.

crates/core/src/surfnet/locker.rs
```diff
@@ -1124,6 +1127,22 @@ impl SurfnetSvmLocker {
...
+            self.with_svm_writer(|svm_writer| {
+                let twsm = TransactionWithStatusMeta::from_failure(
+                    simulated_slot,
+                    transaction,
+                    &failed_transaction_metadata,
+                    accounts_before,
+                    loaded_addresses.clone().unwrap_or_default(),
+                    0,
+                );
+                svm_writer.transactions.insert(
```
MicaiahReid: A transaction that fails during simulation is _not_ processed, and shouldn't be inserted. That's the point of the preflight check - simulate the transaction before executing so I don't make state changes/spend any funds on a transaction that will fail.

All of the changes to this `handle_simulation_failure` can be reverted

crates/core/src/surfnet/locker.rs
```diff
@@ -1164,17 +1186,36 @@ impl SurfnetSvmLocker {
         let signature = meta.signature;
 
         if do_propagate {
-            let meta = convert_transaction_metadata_from_canonical(&meta);
+            let meta_c = convert_transaction_metadata_from_canonical(&meta);
```
MicaiahReid: Let's use a more descriptive name here

crates/core/src/surfnet/locker.rs
```diff
@@ -1164,17 +1186,36 @@ impl SurfnetSvmLocker {
...
-
             let _ = status_tx.try_send(TransactionStatusEvent::ExecutionFailure((
                 err.clone(),
-                meta,
+                meta_c,
             )));
+
+            let fee = 5000 * transaction.signatures.len() as u64;
+            self.with_svm_writer(|svm_writer| {
+                let twsm = TransactionWithStatusMeta::from_failure(
```
MicaiahReid: Let's use a more descriptive name here

crates/core/src/surfnet/locker.rs
```diff
@@ -1164,17 +1186,36 @@ impl SurfnetSvmLocker {
...
                 err
             )));
-
             let _ = status_tx.try_send(TransactionStatusEvent::ExecutionFailure((
                 err.clone(),
-                meta,
+                meta_c,
             )));
+
+            let fee = 5000 * transaction.signatures.len() as u64;
```
MicaiahReid: This calculation of the fee is logic that should be in the `from_failure` function, not outside of it

crates/core/src/types.rs
```diff
@@ -296,6 +296,56 @@ impl TransactionWithStatusMeta {
...
+        failure: &litesvm::types::FailedTransactionMetadata,
+        accounts_before: &[Option<Account>],
+        loaded_addresses: LoadedAddresses,
+        fee: u64,
+    ) -> Self {
+        let pre_balances: Vec<u64> = accounts_before
+            .iter()
+            .map(|a| a.as_ref().map(|a| a.lamports).unwrap_or(0))
+            .collect();
+        let post_balances = pre_balances.clone();
```
MicaiahReid: This seems wrong - if a transaction fails during execution, some lamports will have been spent on fees still, right?

Darn, if so we'll have some more work - you can check our `handle_execution_success` function for how we're fetching the post balance.

crates/core/src/types.rs
```diff
@@ -296,6 +296,56 @@ impl TransactionWithStatusMeta {
...
+                                    instruction: ix.instruction.clone(),
+                                    stack_height: Some(ix.stack_height as u32),
+                                })
+                                .collect(),
+                        })
+                        .collect(),
+                ),
+                log_messages: Some(failure.meta.logs.clone()),
+                pre_token_balances: Some(vec![]),
+                post_token_balances: Some(vec![]),
```
MicaiahReid, line 330: Can you compare against devnet or the test validator to see how failed transactions should work here? I think no token balance changes can happen on a failed transaction - on-chain effects should be reverted.

However, maybe this is supposed to be `None`, or maybe it should just show the same pre/post balance, but actually filled with the token account data.

Please look into this and fill in accordingly!

crates/core/src/surfnet/locker.rs
MicaiahReid: I see you're doing this because we have a different fee for sim vs execution - but we don't need to be creating the meta for a sim, so we can move this logic to the function.

MicaiahReid (changes requested):

crates/core/src/types.rs
BretasArthur1:
```json
{
	"jsonrpc": "2.0",
	"result": {
		"blockTime": 1755101443,
		"meta": {
			"computeUnitsConsumed": 150,
			"err": {
				"InstructionError": [
					0,
					{
						"Custom": 1
					}
				]
			},
			"fee": 5000,
			"innerInstructions": [],
			"logMessages": [
				"Program 11111111111111111111111111111111 invoke [1]",
				"Transfer: insufficient lamports 999995000, need 1000000000000000000",
				"Program 11111111111111111111111111111111 failed: custom program error: 0x1"
			],
			"postBalances": [
				999995000,
				0,
				1
			],
			"postTokenBalances": [],   <----------- Check!
			"preBalances": [
				1000000000,
				0,
				1
			],
			"preTokenBalances": [],   <----------- Check!
			"rewards": [],
			"status": {
				"Err": {
					"InstructionError": [
						0,
						{
							"Custom": 1
						}
					]
				}
			}
		},
		"slot": 102,
		"transaction": {
			"message": {
				"accountKeys": [
					{
						"pubkey": "FcEpvbSPPkJvLemZ4gaoRK1zbUybsx3BRhQ6fo5CyWdf",
						"signer": true,
						"source": "transaction",
						"writable": true
					},
					{
						"pubkey": "FcUUrfzV9NNW8adDVN41x3UV5XQtNmAqCVsKRUf3Jp9N",
						"signer": false,
						"source": "transaction",
						"writable": true
					},
					{
						"pubkey": "11111111111111111111111111111111",
						"signer": false,
						"source": "transaction",
						"writable": false
					}
				],
				"instructions": [
					{
						"parsed": {
							"info": {
								"destination": "FcUUrfzV9NNW8adDVN41x3UV5XQtNmAqCVsKRUf3Jp9N",
								"lamports": 1000000000000000000,
								"source": "FcEpvbSPPkJvLemZ4gaoRK1zbUybsx3BRhQ6fo5CyWdf"
							},
							"type": "transfer"
						},
						"program": "system",
						"programId": "11111111111111111111111111111111",
						"stackHeight": null
					}
				],
				"recentBlockhash": "DDcqRgEcYcUj38CXyAf3UgAim3qvHPL5j6Db8SvvfamP"
			},
			"signatures": [
				"4vHv15rxdP7TWp8ksYaZo8tP1TfzmqjE3crHRjce7KLngw4aZCanmCjRFwRnMuiB3CyPxt7xhFdjLR1PYk8QukkL"
			]
		},
		"version": "legacy"
	},
	"id": 1
}
```

PR #282, opened 2025-08-14
chore(cli): don't start iac until simnet fully Ready

MicaiahReid:


BretasArthur1:
did u test this with offline flag @MicaiahReid ?

PR #284, opened 2025-08-14
Feat/fail pre/post tbalance

BretasArthur1:


crates/core/src/types.rs
```diff
@@ -324,10 +333,77 @@ impl TransactionWithStatusMeta {
...
+                                    a.amount(),
+                                    mint.decimals(),
+                                ),
+                            },
+                            owner: a.owner().to_string(),
+                            program_id: token_program.to_string(),
+                        })
+                        .collect();
+                    Some(balances)
+                },
```
MicaiahReid: Since pre/post will be the same here, can we just do this once at the top of the fn and clone?

MicaiahReid (changes requested):

PR #286, opened 2025-08-15
fix(core): exclude ALT addresses from transaction accounts when fetching pre/post tx data

MicaiahReid:


BretasArthur1:
lgtm

BretasArthur1 (approved):

crates/core/src/rpc/full.rs
```diff
@@ -1646,10 +1646,16 @@ impl Full for SurfpoolFullRpc {
...
+                .get_multiple_accounts(&remote_ctx, &transaction_pubkeys, None)
                 .await?;
 
+            let alt_updates = svm_locker
+                .get_multiple_accounts(&remote_ctx, &alt_pubkeys, None)
+                .await?
+                .inner;
+
             svm_locker.write_multiple_account_updates(&account_updates);
+            svm_locker.write_multiple_account_updates(&alt_updates);
```
BretasArthur1: Can we combine the lines into one call?

PR #296, opened 2025-08-26
Fix/ix-profiling-1-ix

MicaiahReid:


crates/core/src/surfnet/locker.rs
```diff
@@ -2101,6 +2116,17 @@ impl SurfnetSvmLocker {
         for &account_idx in &last.accounts {
             all_required_accounts_for_last_ix.insert(message_accounts[account_idx as usize]);
         }
+        {
+            if all_required_accounts_for_last_ix.len() == 0 {
+                let mut mutable_signers = mutable_signers.iter().cloned().collect::<IndexSet<_>>();
+                all_required_accounts_for_last_ix.append(&mut mutable_signers);
+            }
```
BretasArthur1, line 2108: since both use the same if statement should we just combine them?

crates/core/src/surfnet/locker.rs
```diff
@@ -2074,6 +2074,21 @@ impl SurfnetSvmLocker {
...
+                if all_required_accounts.len() == 0 {
+                    let mut mutable_signers =
+                        mutable_signers.iter().cloned().collect::<IndexSet<_>>();
+                    all_required_accounts.append(&mut mutable_signers);
+                }
+                if all_required_accounts.len() == 0 {
+                    let mut readonly_signers =
+                        readonly_signers.iter().cloned().collect::<IndexSet<_>>();
+                    all_required_accounts.append(&mut readonly_signers);
+                }
```
lgalabru, line 2090: Feels like this snippet is either buggy or something subtile is happening and should be documented?
Is the order `mutable` then `readonly` matters?

MicaiahReid: I've augmented the existing comment. This code is not fully clear to me, but it fixes the behavior for cases I was able to reproduce

lgalabru (approved):

BretasArthur1 (approved):

PR #297, opened 2025-08-27
refactor(tui): enhance airdrop messages on startup

BretasArthur1:


crates/cli/src/cli/simnet/mod.rs
```diff
@@ -144,6 +145,13 @@ pub async fn handle_start_local_surfnet_command(
         }
     }
 
+    if using_default_keypair {
+        let _ = simnet_events_tx.send(SimnetEvent::info(format!(
+            "No airdrop addresses provided; Using default one from {}",
```
MicaiahReid:
```suggestion
            "No airdrop addresses provided; Using default keypair at {}",
```

crates/cli/src/cli/mod.rs
```diff
@@ -169,7 +169,7 @@ pub struct StartSimnet {
     #[arg(long = "airdrop-amount", short = 'q', default_value = DEFAULT_AIRDROP_AMOUNT)]
     pub airdrop_token_amount: u64,
     /// List of keypair paths to airdrop
-    #[arg(long = "airdrop-keypair-path", short = 'k', default_value = DEFAULT_SOLANA_KEYPAIR_PATH.as_str())]
```
MicaiahReid, line 172: One thin this default being part of this clap struct gave us is the default being displayed in our help text:

Image

Can we update the comment to show the default now that we've removed it from here?

MicaiahReid (changes requested):

lgalabru (approved):
Looking good, thanks @BretasArthur1!

crates/cli/src/cli/mod.rs
```diff
@@ -223,22 +225,47 @@ impl StartSimnet {
...
+                Err(_) => {
+                    match using_default_keypair {
+                        true => {
+                            errors.push(
+                                format!("No keypair found at {}, if you want to airdrop to a specific keypair provide the -k flag; skipping airdrops", DEFAULT_SOLANA_KEYPAIR_PATH.to_string())
+                            );
+                        }
+                        false => {
+                            errors.push(format!(
+                                "No keypair found at provided path {}; skipping airdrops;",
```
MicaiahReid: If the user provides 3 keypairs and 2 of them are invalid, this will log:
```sh
No keypair found at provided path {}; skipping airdrops;
No keypair found at provided path {}; skipping airdrops;
Genesis airdrop for {} successful
```

It's weird we're saying "skipping" when we're not necessarily skipping

crates/cli/src/cli/mod.rs
```diff
@@ -223,22 +225,47 @@ impl StartSimnet {
...
         }
 
-        for keypair_path in self.airdrop_keypair_path.iter() {
+        let mut airdrop_keypair_path = self.airdrop_keypair_path.clone();
+        if airdrop_keypair_path.is_empty() {
+            using_default_keypair = true;
+            airdrop_keypair_path.push(DEFAULT_SOLANA_KEYPAIR_PATH.clone());
+        }
+
+        let default_resolved_path = resolve_path(DEFAULT_SOLANA_KEYPAIR_PATH.as_str());
```
MicaiahReid: Rather having multiple bool checks (`using_default_keypair` and setting the `default_resolved_path` and checking for equality later), can we just:
```rust
if airdrop_keypair_path.is_empty() {
   // if default path exists use it
   // if not just return 
}
else {
  // use provided values
}

```

crates/cli/src/cli/mod.rs
```diff
@@ -205,9 +205,11 @@ pub enum NetworkType {
 }
 
 impl StartSimnet {
-    pub fn get_airdrop_addresses(&self) -> (Vec<Pubkey>, Vec<String>) {
+    pub fn get_airdrop_addresses(&self) -> (Vec<Pubkey>, Vec<String>, bool, bool) {
```
MicaiahReid: All of these bools are getting clunky.

If instead we returned `(Vec<Pubkey>, Vec<SimnetEvent>)`, then the caller could do:
```rust
let (mut airdrop_addresses, airdrop_events) = 
        cmd.get_airdrop_addresses();

...
for event in airdrop_events {
        let _ = simnet_events_tx.send(event);
}
```

and inside the function you could push event types accordingly:
```rust
let mut events = vec![];

if none_provided_default_exists {
  events.push(SimnetEvent::info(format!(
            "No airdrop addresses provided; Using default keypair at {}",
            DEFAULT_SOLANA_KEYPAIR_PATH.as_str()
        )))
}
else if none_provided_default_does_not_exist {
  events.push(SimnetEvent::info(format!(
            "No keypair found at default location {}, if you want to airdrop to a specific keypair provide the -k flag; skipping airdrops",
            DEFAULT_SOLANA_KEYPAIR_PATH.as_str()
        )))
}
else {
  .. process provided and push warnings
}
```

MicaiahReid (changes requested):

crates/cli/src/cli/mod.rs
BretasArthur1: I was creating those bools to avoid else if blocks but ok, sg

crates/cli/src/cli/mod.rs
```diff
@@ -205,40 +205,72 @@ pub enum NetworkType {
...
+
         for address in self.airdrop_addresses.iter() {
             match Pubkey::from_str(address).map_err(|e| e.to_string()) {
-                Ok(pubkey) => {
-                    airdrop_addresses.push(pubkey);
-                }
+                Ok(pubkey) => airdrop_addresses.push(pubkey),
                 Err(e) => {
-                    errors.push(format!(
+                    events.push(SimnetEvent::error(format!(
```
MicaiahReid:
```suggestion
                    events.push(SimnetEvent::warn(format!(
```

MicaiahReid (approved):

PR #301, opened 2025-08-27
chore(cli/core/types): allow disabling ix profiling

MicaiahReid:


BretasArthur1 (approved):

crates/core/src/runloops/mod.rs
```diff
@@ -76,7 +76,11 @@ pub async fn start_local_surfnet_runloop(
     };
 
     svm_locker
-        .initialize(simnet.slot_time, &remote_rpc_client)
+        .initialize(
+            simnet.slot_time,
+            &remote_rpc_client,
+            config.simnets[0].do_profile_instructions,
```
lgalabru:
```suggestion
            simnet.do_profile_instructions,
```

crates/core/src/surfnet/svm.rs
```diff
@@ -197,6 +198,7 @@ impl SurfnetSvm {
                 write_version: 0,
                 registered_idls: HashMap::new(),
                 feature_set,
+                do_profile_instructions: true,
```
lgalabru:
```suggestion
                instruction_profiling_enabled: true,
```
?

crates/core/src/surfnet/locker.rs
```diff
@@ -731,7 +737,7 @@ impl SurfnetSvmLocker {
     }
 
     pub fn do_profile_instructions(&self) -> bool {
```
lgalabru:
```suggestion
    pub fn is_instruction_profiling_enabled(&self) -> bool {
```
could be more readable / conventional

crates/core/src/runloops/mod.rs
MicaiahReid: Done

crates/core/src/surfnet/svm.rs
MicaiahReid: Done

crates/core/src/surfnet/locker.rs
MicaiahReid: Done

PR #302, opened 2025-09-01
docs(cli): update command-line argument descriptions for StartSimnet …

BretasArthur1:
…struct to include usage examples

### DIFF
--- a/crates/cli/src/cli/mod.rs
+++ b/crates/cli/src/cli/mod.rs
@@ -125,71 +125,71 @@ enum Command {
 
 #[derive(Parser, PartialEq, Clone, Debug)]
 pub struct StartSimnet {
-    /// Path to the manifest
+    /// Path to the manifest that contains the Simnet configuration (eg. surfpool start --manifest-file-path ./Surfpool.toml)
     #[arg(
         long = "manifest-file-path",
         short = 'm',
         default_value = "./Surfpool.toml"
     )]
     pub manifest_path: String,
-    /// Set the Simnet RPC port
+    /// Set the Simnet RPC port (eg. surfpool start --port 8080)
     #[arg(long = "port", short = 'p', default_value_t = DEFAULT_RPC_PORT)]
     pub simnet_port: u16,
     /// Set the Simnet WS port
     #[arg(long = "ws-port", short = 'w', default_value_t = DEFAULT_WS_PORT)]
     pub ws_port: u16,
-    /// Set the Simnet host address
+    /// Set the Simnet host address (eg. surfpool start --host 127.0.0.1)
     #[arg(long = "host", short = 'o', default_value = DEFAULT_NETWORK_HOST)]
     pub network_host: String,
-    /// Set the slot time
+    /// Set the slot time (eg. surfpool start --slot-time 400)
     #[arg(long = "slot-time", short = 't', default_value_t = DEFAULT_SLOT_TIME_MS)]
     pub slot_time: u64,
-    /// Set a datasource RPC URL (cannot be used with --network). Can also be set via SURFPOOL_DATASOURCE_RPC_URL.
+    /// Set a datasource RPC URL (cannot be used with --network). Can also be set via SURFPOOL_DATASOURCE_RPC_URL. (eg. surfpool start --rpc-url https://api.mainnet-beta.solana.com)
     #[arg(long = "rpc-url", short = 'u', conflicts_with = "network")]
     pub rpc_url: Option<String>,
-    /// Choose a predefined network (cannot be used with --rpc-url)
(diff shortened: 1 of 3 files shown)


crates/cli/src/cli/mod.rs
```diff
@@ -125,71 +125,71 @@ enum Command {
 
 #[derive(Parser, PartialEq, Clone, Debug)]
 pub struct StartSimnet {
-    /// Path to the manifest
+    /// Path to the manifest that contains the Simnet configuration (eg. surfpool start --manifest-file-path ./Surfpool.toml)
```
MicaiahReid: This is a confusing field. We actually aren't using the file we point to for anything, just using the directory it's in. So currently, we have no concept of a "Surfpool.toml". 

I suppose we will one day. For now, maybe we could do this? Thoughts, @lgalabru 

```suggestion
    /// Path to the runbook manifest, used to locate the root of the project (eg. surfpool start --manifest-file-path ./txtx.toml)
```

crates/cli/src/cli/mod.rs
```diff
@@ -133,71 +133,71 @@ enum Command {
...
     #[arg(long = "airdrop-amount", short = 'q', default_value = DEFAULT_AIRDROP_AMOUNT)]
     pub airdrop_token_amount: u64,
-    /// List of keypair paths to airdrop (default: ~/.config/solana/id.json)
+    /// List of keypair paths to airdrop (default: ~/.config/solana/id.json) (eg. surfpool start --airdrop-keypair-path ~/.config/solana/id.json --airdrop-keypair-path ~/.config/solana/id2.json)
     #[arg(long = "airdrop-keypair-path", short = 'k')]
     pub airdrop_keypair_path: Vec<String>,
-    /// Disable explorer (default: false)
+    /// Disable explorer (default: false) (eg. surfpool start --no-explorer)
     #[clap(long = "no-explorer")]
     pub no_explorer: bool,
```
lgalabru: can we please get rid of this `no_explorer`? this is dead code!

crates/cli/src/cli/mod.rs
```diff
@@ -133,71 +133,71 @@ enum Command {
...
-    /// Set a datasource RPC URL (cannot be used with --network). Can also be set via SURFPOOL_DATASOURCE_RPC_URL.
+    /// Set a datasource RPC URL (cannot be used with --network). Can also be set via SURFPOOL_DATASOURCE_RPC_URL. (eg. surfpool start --rpc-url https://api.mainnet-beta.solana.com)
     #[arg(long = "rpc-url", short = 'u', conflicts_with = "network")]
     pub rpc_url: Option<String>,
-    /// Choose a predefined network (cannot be used with --rpc-url)
+    /// Choose a predefined network (cannot be used with --rpc-url) (eg. surfpool start --network mainnet)
     #[arg(long = "network", short = 'n', value_enum, conflicts_with = "rpc_url")]
     pub network: Option<NetworkType>,
-    /// Display streams of logs instead of terminal UI dashboard (default: false)
+    /// Display streams of logs instead of terminal UI dashboard (default: false) (eg. surfpool start --no-tui)
```
lgalabru: In the current version, if I pull the help for `surfpool start --help`, I see:
```
      --no-studio
          Disable Studio (default: true)

      --studio-port <STUDIO_PORT>
          Set the Studio port

          [default: 18488]
```

Could we get rid of the all the `(default: x)` components from the comments strings and instead do:
```rust
    /// Display streams of logs instead of terminal UI dashboard (eg. surfpool start --no-tui)
    #[clap(long = "no-tui", default_value_t = false)]
    pub no_tui: bool,
```
so we leave `clap` displaying the default values ?

crates/cli/src/cli/mod.rs
```diff
@@ -133,72 +133,79 @@ enum Command {
...
-    /// Choose a predefined network (cannot be used with --rpc-url)
-    #[arg(long = "network", short = 'n', value_enum, conflicts_with = "rpc_url")]
+    /// Choose a predefined network (cannot be used with --rpc-url) (eg. surfpool start --network mainnet)
+    #[arg(
+        long = "network",
+        short = 'n',
+        value_enum,
+        conflicts_with = "rpc_url",
+        default_value = "mainnet"
+    )]
```
lgalabru: hmm I have some doubts with this approach.
could you confirm that if I run:
```
surfpool start
```
with a env var `SURFPOOL_DATASOURCE_RPC_URL` set to, say, a helius RPC url, then it is being used instead of that `mainnet` default?

BretasArthur1: Hm ok, I will remove it then, u right

lgalabru (approved):
Awesome, thanks @BretasArthur1!

PR #303, opened 2025-09-01
feat(cli, tui): add terminal color compatibility 

BretasArthur1:
# Problem
before any terminal was using our hardcoded tailwind colors that aren't compatible with every terminal, causing a bad UX when the user it's using a non supported terminal

```rs
Self {
            background: tailwind::ZINC.c900,
            accent: color.c400,
            primary: color.c500,
            secondary: color.c300,
            white: tailwind::SLATE.c100,
            dark_gray: tailwind::ZINC.c800,
            light_gray: tailwind::ZINC.c400,
            error: tailwind::RED.c400,
            warning: tailwind::YELLOW.c500,
            info: tailwind::BLUE.c500,
            success: tailwind::GREEN.c500,
        }
```

# Solution
Added helper functions that check if the terminal is compatible with the default color schema and if isn't we use the fallback color schema provided with `ratatui` that's compatible with every terminal. Also warn the user about the compatibility.

```rs
   Self {
            background: Color::Reset,
            accent: Color::Green,
            primary: Color::Green,
            secondary: Color::White,
            white: Color::White,
            dark_gray: Color::DarkGray,
            light_gray: Color::Gray,
            error: Color::Red,
            warning: Color::Yellow,
            info: Color::Blue,
            success: Color::Green,
        }
    }
```

### DIFF
--- a/crates/cli/src/cli/simnet/mod.rs
+++ b/crates/cli/src/cli/simnet/mod.rs
@@ -104,12 +104,7 @@ pub async fn handle_start_local_surfnet_command(
         Ok((explorer_handle, _)) => Some(explorer_handle),
         Err(e) => {
             let _ = simnet_events_tx.send(SimnetEvent::warn(format!(
-                "Failed to start subgraph {}server: {}",
-                if !cmd.no_explorer {
-                    "and explorer "
-                } else {
-                    ""
-                },
+                "Failed to start subgraph and explorer server: {}",
                 e
             )));
             let _ = simnet_events_tx.send(SimnetEvent::info("Continuing with simnet startup..."));
--- a/crates/cli/src/cli/mod.rs
+++ b/crates/cli/src/cli/mod.rs
@@ -133,7 +133,7 @@ enum Command {
 
 #[derive(Parser, PartialEq, Clone, Debug)]
 pub struct StartSimnet {
-    /// Path to the manifest that contains the Simnet configuration (eg. surfpool start --manifest-file-path ./Surfpool.toml)
+    /// Path to the runbook manifest, used to locate the root of the project (eg. surfpool start --manifest-file-path ./txtx.toml)
     #[arg(
         long = "manifest-file-path",
         short = 'm',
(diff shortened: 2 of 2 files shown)


crates/cli/src/cli/simnet/mod.rs
```diff
@@ -39,6 +40,40 @@ struct CheckVersionResponse {
...
+        ));
+        info!(
+            ctx.expect_logger(),
+            "Terminal compatibility: macOS Terminal detected, using compatible color palette"
+        );
+    } else if term.starts_with("xterm") && !term.contains("256") && colorterm.is_empty() {
+        let _ = simnet_events_tx.send(SimnetEvent::info(
+            "Limited color terminal detected - using compatibility mode. If colors appear incorrect, try setting COLORTERM=truecolor".to_string()
+        ));
+        info!(
```
MicaiahReid: We've discussed in the past, but we shouldn't be using these `info!` macros.

crates/cli/src/cli/simnet/mod.rs
```diff
@@ -39,6 +40,40 @@ struct CheckVersionResponse {
...
+    // Detect macOS Terminal specifically
+    if term_program == "Apple_Terminal" {
+        let _ = simnet_events_tx.send(SimnetEvent::info(
+            "macOS Terminal detected - using compatibility color mode. For best experience, consider using iTerm2 or similar modern terminal.".to_string()
+        ));
+        info!(
+            ctx.expect_logger(),
+            "Terminal compatibility: macOS Terminal detected, using compatible color palette"
+        );
+    } else if term.starts_with("xterm") && !term.contains("256") && colorterm.is_empty() {
```
MicaiahReid: Having these strings be named constants would help. In this context, I have no idea of the relevance of "256", but a named var would fix that

MicaiahReid (changes requested):
Good find!

I'm a bit concerned of the robustness of the checks we're doing of the env vars, but I suppose we can add to these checks as cases arise in the future.

I don't love that we're having to check the env vars twice, in two different fns. If we find we have to add to this, we need to remember to add in two different places.

Can you consolidate this to one function, so there's only one set of checks on these vars?

crates/cli/src/cli/simnet/mod.rs
lgalabru: I don't think we should have this message / promoting other terms.

lgalabru: No one really care about colors fidelity

lgalabru: Like, I would not install and configure a new terminal application for better colors in surfpool.

MicaiahReid: True, agreed, can probably remove this whole fn

crates/cli/src/cli/simnet/mod.rs
```diff
@@ -39,6 +42,23 @@ struct CheckVersionResponse {
     pub deprecation_notice: Option<String>,
 }
 
+/// Check terminal compatibility and warn about potential display issues
+fn check_and_warn_terminal_compatibility(simnet_events_tx: &Sender<SimnetEvent>) {
+    let terminal_checks = TerminalChecks::new();
+
+    if terminal_checks.is_limited_terminal() {
```
MicaiahReid: Per ludo's earlier comment, I don't think we want any messages about this. Just use a different terminal if not supporting full colors

MicaiahReid (approved):
Thanks Arthur, good find!!

lgalabru (approved):
Thanks @BretasArthur1 !

PR #304, opened 2025-09-01
fix(core): deactivate enable_extend_program_checked feature

MicaiahReid:


crates/core/src/surfnet/svm.rs
```diff
@@ -140,7 +140,11 @@ impl SurfnetSvm {
         let (simnet_events_tx, simnet_events_rx) = crossbeam_channel::bounded(1024);
         let (geyser_events_tx, geyser_events_rx) = crossbeam_channel::bounded(1024);
 
-        let feature_set = FeatureSet::all_enabled();
+        let mut feature_set = FeatureSet::all_enabled();
+
+        // todo: remove once txtx deployments upgrade solana dependencies.
+        // todo: consider making this configurable via config
+        feature_set.deactivate(&enable_extend_program_checked::id());
```
BretasArthur1, line 147: this will bug that case were we have 2 different programs being deployed with different feature no?

lgalabru (approved):

BretasArthur1 (approved):

crates/core/src/surfnet/svm.rs
MicaiahReid: nope

PR #310, opened 2025-09-03
fix: stabilizing websocket

lgalabru:


crates/core/Cargo.toml
```diff
@@ -89,12 +89,18 @@ txtx-addon-network-svm-types = { workspace = true }
 uuid = "1.7.0"
 blake3 = "1.8.2"
 anyhow = "1.0.98"
-solana-geyser-plugin-manager = { workspace = true, optional = true }
+# solana-geyser-plugin-manager = { version = "2.3.7", optional = true }
```
MicaiahReid, line 92: Are there comments in this file that we can remove?

Cargo.toml
```diff
@@ -66,24 +66,24 @@ solana-commitment-config = "2.2.1"
 solana-compute-budget-interface = "2.2.1"
 solana-epoch-info = { version = "2.2.1", features = ["serde"] }
 solana-epoch-schedule = "2.2.1"
-solana-geyser-plugin-manager = "3.0.0"
+# solana-geyser-plugin-manager = "=2.3.7"
```
MicaiahReid, line 69: Can we remove this comment?

lgalabru: I'd like to follow-up with a subsequent PR for the geyser stuffs, but would love to get https://github.com/txtx/surfpool/pull/312 merged too

MicaiahReid: Sounds good, still trying to get 312 passing

PR #318, opened 2025-09-08
feat: implement a new type `FifoMap` and associated tests

BretasArthur1:
# Problem

Our currently implementation of holds profile results in a hashmap that keeps growing forever, slowing down the svm more and more.

# Solution
Implementing a new type that have a fixed size and holds the elements in the order they enter on the map. Whenever we add a element that should overflow the capacity we pop the first element and add to the last spot the new one(FIFO behavior)

### DIFF
--- a/crates/core/src/error.rs
+++ b/crates/core/src/error.rs
@@ -424,16 +424,6 @@ impl SurfpoolError {
         Self(error)
     }
 
-    pub fn transaction_not_found_in_svm<S>(signature: S) -> Self
-    where
-        S: Display,
-    {
-        let mut error = Error::internal_error();
-        error.message =
-            format!("Transaction with signature '{signature}' was not found in the SVM");
-        Self(error)
-    }
-
     pub fn tag_not_found(tag: &str) -> Self {
         let mut error = Error::internal_error();
         error.message = format!("Profile result associated with tag '{tag}' not found in the SVM");
--- a/crates/core/src/surfnet/locker.rs
+++ b/crates/core/src/surfnet/locker.rs
@@ -2331,22 +2331,10 @@ impl SurfnetSvmLocker {
         config: &RpcProfileResultConfig,
     ) -> SurfpoolResult<Option<UiKeyedProfileResult>> {
         let result = match &signature_or_uuid {
-            UuidOrSignature::Signature(signature) => {
-                let profile = self.with_svm_reader(|svm| {
-                    svm.executed_transaction_profiles.get(signature).cloned()
-                });
-                let transaction_exists =
-                    self.with_svm_reader(|svm| svm.transactions.contains_key(signature));
-                if profile.is_none() && transaction_exists {
-                    return Err(SurfpoolError::transaction_not_found_in_svm(signature));
-                } else {
-                    profile
-                }
-            }
+            UuidOrSignature::Signature(signature) => self
+                .with_svm_reader(|svm| svm.executed_transaction_profiles.get(signature).cloned()),
             UuidOrSignature::Uuid(uuid) => {
-                let profile = self
-                    .with_svm_reader(|svm| svm.simulated_transaction_profiles.get(uuid).cloned());
-                profile
+                self.with_svm_reader(|svm| svm.simulated_transaction_profiles.get(uuid).cloned())
             }
         };
         Ok(result.map(|profile| self.encode_ui_keyed_profile_result(profile, config)))


crates/types/src/types.rs
```diff
@@ -829,6 +829,66 @@ impl Ord for VersionedIdl {
...
+    /// Returns the old value if this was an update.
+    pub fn insert(&mut self, key: K, value: V) -> Option<V> {
+        if self.map.contains_key(&key) {
+            // Update doesn't change insertion order in IndexMap
+            return self.map.insert(key, value);
+        }
+        if self.map.len() == CAPACITY {
+            // Evict oldest (index 0). O(n) due shifting the rest of the map
+            //We could use a hashmap + vecdeque to get O(1) here, but then we'd have to handle removing from both maps, storing the index, and managing the eviction.
+            //This is a good compromise between performance and simplicity. And thinking about memory usage, this is probably the best way to go.
```
```suggestion
            // We could use a hashmap + vecdeque to get O(1) here, but then we'd have to handle removing from both maps, storing the index, and managing the eviction.
            // This is a good compromise between performance and simplicity. And thinking about memory usage, this is probably the best way to go.
```

crates/types/src/types.rs
```diff
@@ -829,6 +829,66 @@ impl Ord for VersionedIdl {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct ProfilingMap<K, V, const CAPACITY: usize> {
```
MicaiahReid: You've made this map generic so that it can accept any type as the key + value (which is cool!!), but you've given it a specific name - `ProfilingMap` limits this to just store profiles, though the implementation can store anything.

We should either:
 1. Remove generics and keep the name OR
 2. Keep generics and make a more generic name like `FifoMap` (preferred IMO)

MicaiahReid: Additionally, this `CAPACITY` being passed in as a generic will make it difficult to pass in the capacity at runtime. I think the `new` function should receive the capacity instead.

crates/cli/src/cli/mod.rs
```diff
@@ -197,15 +197,23 @@ pub struct StartSimnet {
...
     #[arg(long = "log-path", default_value = DEFAULT_LOG_DIR.as_str())]
     pub log_dir: String,
+    /// Set the maximum capacity of the map that holds the instruction profiling results
+    /// Changing this will affect the memory usage of surfpool. (eg. surfpool start --instruction-profiling-map-capacity 2000)
+    #[arg(
+        long = "instruction-profiling-map-capacity",
+        short = 'c',
+        default_value = "200"
+    )]
+    pub instruction_profiling_map_capacity: usize,
```
MicaiahReid: Sorry, in your DM I didn't know you were talking about CLI help text. I thought you meant just a comment.

What do you think about:
```rust

    /// The maximum number of transaction profiles to hold in memory.
    /// Changing this will affect the memory usage of surfpool. (eg. surfpool start --max-profiles 2000)
    #[arg(
        long = "max-profiles",
        short = 'c',
        default_value = "200"
    )]
    pub max_profiles: usize,
```

Let's discuss, maybe what I haven't isn't right either. But my thoughts are:
 1. Using the word "map" feels like internal kitchen stuff - the user doesn't need to know we're storing in a map
 2. These aren't necessarily only ix profiles. If ix profiling is completely disabled, we still store tx profiles here, so this is just general profile storage.

crates/core/src/surfnet/svm.rs
```diff
@@ -180,8 +185,8 @@ impl SurfnetSvm {
                 account_subscriptions: HashMap::new(),
                 slot_subscriptions: Vec::new(),
                 profile_tag_map: HashMap::new(),
-                simulated_transaction_profiles: HashMap::new(),
-                executed_transaction_profiles: HashMap::new(),
+                simulated_transaction_profiles: FifoMap::default(), // With DEFAULT_PROFILING_MAP_CAPACITY capacity
```
MicaiahReid: Would it make sense for us to be passing in the CLI param to this `new` function instead of having a default, then changing it later in `initialize`?

crates/core/src/surfnet/svm.rs
```diff
@@ -180,8 +185,8 @@ impl SurfnetSvm {
                 account_subscriptions: HashMap::new(),
                 slot_subscriptions: Vec::new(),
                 profile_tag_map: HashMap::new(),
-                simulated_transaction_profiles: HashMap::new(),
```
MicaiahReid: I think we'd discussed leaving simulated profiles as boundless - if a user is using `surfnet_profileTransaction`, it's quite intentional and I doubt they want some deleted. Thoughts?

crates/cli/src/cli/mod.rs
BretasArthur1: hm, ok sounds good I will make a patch

Issue #128, opened 2025-05-27
Introduce `surfnet_resetAccount`


lgalabru:

sonicfromnewyoke:
this is should be essential part of the further functionality for passing whole block of some subsequence of txs (testing reactive apps, backtesting of strategies, etc) to be able to rerun the same sequence without stopping/resetting all the local node

MicaiahReid:
@sonicfromnewyoke, are you interested in implementing this cheatcode? 👀

sonicfromnewyoke:
> [@sonicfromnewyoke](https://github.com/sonicfromnewyoke), are you interested in implementing this cheatcode? 👀

yeah, will take it just after finishing geyser part.

to be able to grow into more complex scenarios like [accounts-db-plugin](https://github.com/solana-labs/solana-accountsdb-plugin-postgres/blob/master/tests/test_postgres_plugin.rs#L61) has - we have to pay extra attention to the internal design of this cheatcode

MicaiahReid:
Cool, let me know if you want to hop on a call to discuss some time!

frenzox:
Looking forward to this one as it's going to help us using `surfpool` to estimate gas costs of transaction batches before broadcasting them. As discussed on Discord.

> Hi there 👋 ,
> Thank you guys for this project, it's awesome!
> I'm wondering if there's a way to remove accounts from surfpool and, in case they're used again, re-fetch them from the remote RPC.
> Some context: I have a long running surfpool instance where I send a batch of transactions in order to estimate the costs of the whole batch. It's working fine, but I'm concerned memory usage will get out of hand over time, thus I'd like to clean up after the estimation (that would also help in case an account is reused later and we need it to be synced). I've tried setting the accounts lamports to 0, but when the accounts are used later, they don't seem to get synced again and the transactions just fail with insufficient funds errors.

lgalabru:
Related:
We need to introduce `surfnet_resetNetwork`

frenzox:
@lgalabru is my understanding correct that this has been introduced in https://github.com/txtx/surfpool/pull/328?

lgalabru:
@frenzox correct, it's available in 0.10.9!
Will close this issue.

BretasArthur1:
We still need to refactor this patch

https://github.com/txtx/surfpool/issues/332

PR #328, opened 2025-09-12
feat(core): add reset_account cheatcode RPC method and implement acco…

BretasArthur1:
…unt reset functionality

### DIFF
--- a/crates/core/src/rpc/surfnet_cheatcodes.rs
+++ b/crates/core/src/rpc/surfnet_cheatcodes.rs
@@ -729,7 +729,7 @@ pub trait SurfnetCheatcodes {
     ///   "jsonrpc": "2.0",
     ///   "id": 1,
     ///   "method": "surfnet_resetAccount",
-    ///   "params": [ "4EXSeLGxVBpAZwq7vm6evLdewpcvE2H56fpqL2pPiLFa", { "cascadeToOwned": true } ]
+    ///   "params": [ "4EXSeLGxVBpAZwq7vm6evLdewpcvE2H56fpqL2pPiLFa", { "recursive": true } ]
     /// }
     /// ```
     ///


BretasArthur1:
closes #128

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -1223,6 +1226,16 @@ impl SurfnetCheatcodes for SurfnetCheatcodesRpc {
...
+
+    fn reset_account(&self, meta: Self::Metadata, pubkey: String) -> Result<RpcResponse<()>> {
+        let svm_locker = meta.get_svm_locker()?;
+        let pubkey = verify_pubkey(&pubkey)?;
+        svm_locker.reset_account(pubkey)?;
+        Ok(RpcResponse {
+            context: RpcResponseContext::new(svm_locker.get_latest_absolute_slot()),
+            value: (),
+        })
+    }
```
MicaiahReid: CLEAN ✨

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -711,6 +711,9 @@ pub trait SurfnetCheatcodes {
     /// ```
     #[rpc(meta, name = "surfnet_resumeClock")]
     fn resume_clock(&self, meta: Self::Metadata) -> Result<EpochInfo>;
+
+    #[rpc(meta, name = "surfnet_resetAccount")]
+    fn reset_account(&self, meta: Self::Metadata, pubkey_str: String) -> Result<RpcResponse<()>>;
```
MicaiahReid: Currently we aren't allowing the user to pass any config, can we add that?


```
    fn reset_account(&self, meta: Self::Metadata, pubkey_str: String, config: ResetAccountConfig) -> Result<RpcResponse<()>>;
```

and in our types something like
```rust
pub struct ResetAccountConfig {
    pub cascade_to_owned: Option<bool>
}
```

MicaiahReid (changes requested):
I think you've nailed all of the core logic.

My request is a reorg for clarity and consistency with our other functions. I think we should have:
 1. One function on the `SurfnetSvm` (svm.rs file) called `reset_account` that does the actual removing from indexes and setting the inner svm state. This would look something like the current `reset_regular_account` fn
 2. One function on the locker with sig `pub fn reset_account(&self, pubkey: Pubkey, options: ResetAccountOptions) -> SurfpoolResult<()>` that:
    1. Resets the program data account if a program account is passed (by calling the new `SurfnetSvm` file's `reset_account` fn
    2. Cascades to owned if provided, each account calling the one fn on svm.rs
    3. Resets the actual account last

This gives us one function (on the svm) that actually writes to the internals, and one function that coordinates how that internal fn is called. It also prevents us from needing to make `remove_from_indexes` public

crates/core/src/rpc/surfnet_cheatcodes.rs
BretasArthur1: 🚀

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -711,6 +712,14 @@ pub trait SurfnetCheatcodes {
     /// ```
     #[rpc(meta, name = "surfnet_resumeClock")]
     fn resume_clock(&self, meta: Self::Metadata) -> Result<EpochInfo>;
+
+    #[rpc(meta, name = "surfnet_resetAccount")]
+    fn reset_account(
+        &self,
+        meta: Self::Metadata,
+        pubkey_str: String,
+        config: ResetAccountConfig,
```
MicaiahReid: Ahh, to match other RPCs, I think the config will need to be optional `config: Option<ResetAccountConfig>` with defaults set so you can do `let config = config.unwrap_or_default();`

crates/core/src/surfnet/locker.rs
```diff
@@ -1605,6 +1620,42 @@ impl SurfnetSvmLocker {
...
         });
     }
+
+    /// Resets an account in the SVM state
+    ///
+    /// This function coordinates the reset of accounts by calling the SVM's reset_account method.
+    /// It handles program accounts (including their program data accounts) and can optionally
+    /// cascade the reset to all accounts owned by a program.
+    pub fn reset_account(&self, pubkey: Pubkey, config: ResetAccountConfig) -> SurfpoolResult<()> {
+        let cascade_to_owned = config.cascade_to_owned.unwrap_or_else(|| false);
```
MicaiahReid: The default value for a bool is false, so let's do `unwrap_or_default()` here instead. I know it amounts to the same thing, but it implies "the struct handles the default case"

crates/core/src/surfnet/locker.rs
```diff
@@ -1605,6 +1620,42 @@ impl SurfnetSvmLocker {
...
+                        svm_writer.reset_account(&program_data_address)?;
+                    }
+
+                    if cascade_to_owned {
+                        let owned_accounts = svm_writer.get_account_owned_by(pubkey);
+                        for (owned_pubkey, _) in owned_accounts {
+                            // Avoid infinite recursion by not cascading further
+                            svm_writer.reset_account(&owned_pubkey)?;
+                        }
+                    }
```
MicaiahReid: This block is currently inside `account.executable`, which means we'll only be cascading to owned accounts if this is a program account. Is that right? I don't think so, I think we need to move this outside the if statement.

crates/core/src/surfnet/locker.rs
```diff
@@ -1605,6 +1620,42 @@ impl SurfnetSvmLocker {
...
+                }
+                // Reset the account itself (this handles both regular and program accounts)
```
```suggestion
                // Reset the account itself
```
It's not doing anything special for program accounts, so I think that's not relevant

crates/core/src/surfnet/svm.rs
```diff
@@ -641,6 +641,30 @@ impl SurfnetSvm {
...
+            data: vec![],
+            owner: solana_sdk_ids::system_program::id(),
+            executable: false,
+            rent_epoch: 0,
+        };
+
+        // Set the empty account
+        self.inner
+            .set_account(*pubkey, empty_account)
+            .map_err(|e| SurfpoolError::set_account(*pubkey, e))?;
```
MicaiahReid: Rather than creating an empty account, you should be able to just do `Account::default()`:
```rust
 self.inner
            .set_account(*pubkey, Account::default())
            .map_err(|e| SurfpoolError::set_account(*pubkey, e))?;
```

MicaiahReid (changes requested):

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -711,6 +712,14 @@ pub trait SurfnetCheatcodes {
     /// ```
     #[rpc(meta, name = "surfnet_resumeClock")]
     fn resume_clock(&self, meta: Self::Metadata) -> Result<EpochInfo>;
+
```
lgalabru, line 715: Could you please add some docs + update the json for our doc website?

crates/types/src/types.rs
```diff
@@ -903,6 +903,19 @@ impl<K: std::hash::Hash + Eq, V> FifoMap<K, V> {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct ResetAccountConfig {
+    pub cascade_to_owned: Option<bool>,
```
lgalabru: I like `cascade` but `cascade_to_owned` sounds funny. 
Could `recursive` work?

crates/core/src/surfnet/locker.rs
```diff
@@ -1605,6 +1620,40 @@ impl SurfnetSvmLocker {
...
+
+                        // Reset the program data account first
+                        svm_writer.reset_account(&program_data_address)?;
+                    }
+                }
+                if cascade_to_owned {
+                    let owned_accounts = svm_writer.get_account_owned_by(pubkey);
+                    for (owned_pubkey, _) in owned_accounts {
+                        // Avoid infinite recursion by not cascading further
+                        svm_writer.reset_account(&owned_pubkey)?;
```
lgalabru, line 1648: Are we missing some recursion here?

crates/core/src/surfnet/locker.rs
```diff
@@ -1605,6 +1620,40 @@ impl SurfnetSvmLocker {
...
+                    if account.owner == solana_sdk_ids::bpf_loader_upgradeable::id() {
+                        let program_data_address =
+                            solana_sdk::bpf_loader_upgradeable::get_program_data_address(&pubkey);
+
+                        // Reset the program data account first
+                        svm_writer.reset_account(&program_data_address)?;
+                    }
+                }
+                if cascade_to_owned {
+                    let owned_accounts = svm_writer.get_account_owned_by(pubkey);
```
lgalabru, line 1645: Should we rename this method to `get_accounts_owned_by`?

BretasArthur1: probably yes

crates/core/src/surfnet/locker.rs
BretasArthur1: wdym?

crates/core/src/surfnet/locker.rs
lgalabru: when a owned account is deleted, should we be looking for the "sub-owned" account and reset? (and recursive)

MicaiahReid: It kinda seems like overkill to me, but I could be wrong. We could ship as is and see if the feature is requested?

lgalabru: Sure

lgalabru: It will be a bug report, not a feature request 😅

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -711,6 +712,48 @@ pub trait SurfnetCheatcodes {
...
+    /// ## Returns
+    /// An `RpcResponse<()>` indicating whether the account reset was successful.
+    ///
+    /// ## Example Request
+    /// ```json
+    /// {
+    ///   "jsonrpc": "2.0",
+    ///   "id": 1,
+    ///   "method": "surfnet_resetAccount",
+    ///   "params": [ "4EXSeLGxVBpAZwq7vm6evLdewpcvE2H56fpqL2pPiLFa", { "cascadeToOwned": true } ]
```
lgalabru: If we're changing the setting name to `recursive`, it should be reflected here!

lgalabru: (otherwise, revert the name)

BretasArthur1: ops, I forgot about it

crates/core/src/surfnet/locker.rs
BretasArthur1: agree with micaiah

lgalabru (approved):
Looking great, thank you @BretasArthur1!

PR #340, opened 2025-09-24
feat(cli): add `--yes` option to auto-accept all prompts when generating runbooks

MicaiahReid:


lgalabru (approved):
Looking good thanks @MicaiahReid !

PR #341, opened 2025-09-24
feat(core): add `surfnet_getSurfnetInfo` RPC endpoint

MicaiahReid:


crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -754,6 +754,9 @@ pub trait SurfnetCheatcodes {
         pubkey_str: String,
         config: Option<ResetAccountConfig>,
     ) -> Result<RpcResponse<()>>;
+
+    #[rpc(meta, name = "surfnet_readyCheck")]
```
lgalabru: I could see this endpoint exposing more data. how about 
```
surfnet_getSurfnetInfo
```

exposing a structure that would look like:
```rust
struct GetSurfnetInfoResponse {
    runbook_executions: Vec<RunbookExecutionReport>
}

struct RunbookExecutionReport {
    started_at: u32,
    completed_at: Option<u32>,
    runbook_id: String,
    status: String
}
```

lgalabru: Once settled, could we add some documentation + update `docs.surfpool.run`?
I believe we are a few cheat codes behind - cc @BretasArthur1

BretasArthur1: yes, we're

MicaiahReid: I like it! Idk about the status field though. What kind of statuses would we want to store?

My guess is running/complete/failed. Which could be inferred from a combo of `completed_at: Option<u32>` and an `error: Option<u32>`

MicaiahReid: Implemented!

PR #343, opened 2025-09-25
refactor: use more granular solana crates

sonicfromnewyoke:
### Problem

`surfpool` uses bloated `solana-sdk` and `solana-program` crates

### Solution

replace them with more granular crates

cc @MicaiahReid


lgalabru (approved):
Looking sharp, thanks @sonicfromnewyoke !

sonicfromnewyoke:
> This is awesome, thanks @sonicfromnewyoke!! Excited to see the difference in bin size once this is built

to actually win in compile time and binary size we have to apply the next 2 improvens:
- use workspace dependencies in the members
- disable default feature by default and include only needed, e.g. `solana-pubkey = { version = "2.3.0", default-features = false  }` in the root and `solana-pubkey = { workspace = true, features = ["serde"] }` in the crate that needs `serde` feature of pubkey
-some minor tweaks

all those things i'll implement shortly (EOW)

PR #349, opened 2025-10-01
fix(core): fetch mint for token accounts on `get_multiple_accounts` endpoint

MicaiahReid:


crates/core/src/surfnet/remote.rs
```diff
@@ -157,14 +157,27 @@ impl SurfnetRemoteClient {
...
-                    accounts_result.push(GetAccountResult::FoundAccount(
-                        *pubkey,
-                        remote_account,
-                        // Mark this account as needing to be updated in the SVM, since we fetched it
-                        true,
-                    ));
-                } else {
+                if is_supported_token_program(&remote_account.owner) {
+                    if let Some(token_account) = TokenAccount::unpack(&remote_account.data).ok() {
+                        let mint = self
```
sonicfromnewyoke, line 163:
```suggestion
                       // TODO: move the query out of the loop to prevent rate-limiting by `api.mainnet-beta.solana.com`
                        let mint = self
```

PR #351, opened 2025-10-03
feat(cli/core): add log bytes limit cli argument

valentinmadrid:
adds log byte limit with 10_000 bytes default limit


lgalabru (approved):
Thanks @valentinmadrid !

PR #354, opened 2025-10-05
fix(core): use config in `getLatestBlockhash` rpc method

MicaiahReid:


lgalabru:
It looks like we have a regression?

Screenshot 2025-10-07 at 10 34 57

PR #355, opened 2025-10-07
add setMultipleAccounts cheatcode

mubarizkyc:


sonicfromnewyoke (commented):
i'm not sure about the api

```rust
 fn set_multiple_accounts(
        &self,
        meta: Self::Metadata,
        pubkeys_str: Vec<String>,
        updates: Vec<AccountUpdate>,
    ) -> BoxFuture<Result<RpcResponse<()>>> 
  ```
  
 this means the order of keys and account updates should be manually mapped and ensured before the call, otherwise it leads to bugs.
 
 i'd prefer something like `HashMap<String, AccountUpdate>` that minimize cognetive load and surface for bugs.

lgalabru:
Thank you for this contribution @mubarizkyc 🙏
Could you update the description of this PR, unpacking how you'd like to use this cheat code? 
Do you think you could add some test coverage?

PR #363, opened 2025-10-11
Added missing solana clock dependency

AllenGeorge08:
There was a missing dependency in the `Cargo.toml` for the Solana clock in the `hello_geyser` example and a error was being thrown while running `cargo build --release`. I added the dependency to the `Cargo.toml` and the build is now working perfectly fine.

Thank you.


lgalabru:
Good catch, thanks @AllenGeorge08!

PR #364, opened 2025-10-11
feat: introduce the `surfnet_exportSnapshot` cheatcode

Shradhesh71:
## Flat Key/Value Dump with Slot Tracking

- Added `account_update_slots: HashMap<Pubkey, Slot>` field to **SurfnetSvm** to track when each account was last modified.  
- Updated `set_account()` method to record the modification slot for each account update.  
- Implemented `export_accounts_as_fixtures()` method that exports all accounts as a flat map with slot information.  
- Added `AccountFixture` struct with the following fields:  
  - **pubkey**: Account public key  
  - **lamports**: Account balance  
  - **owner**: Program owner  
  - **executable**: Executable flag  
  - **rent_epoch**: Rent epoch  
  - **data**: Account data (encoded based on specified encoding)  
  - **parsed_data**: Optional parsed data (when using `JsonParsed` encoding)  
  - **slot**: Absolute slot index when the account was last updated 

## RPC Method

- Added `surfnet_exportSnapshot` RPC method to **SurfnetCheatcodes** trait.  
- Accepts optional `encoding` parameter (defaults to `Base64`).  
- Returns `HashMap<String, AccountFixture>` with all accounts.  

---

## Test Coverage

-  `test_export_accounts_as_fixtures` — Basic export with Base64 encoding  
-  `test_export_accounts_as_fixtures_with_json_parsed` — Export with JsonParsed encoding for token accounts  
-  `test_export_accounts_at_different_slots` — Verify slot tracking across multiple slot updates

### DIFF
--- a/crates/core/src/rpc/surfnet_cheatcodes.rs
+++ b/crates/core/src/rpc/surfnet_cheatcodes.rs
@@ -4,7 +4,6 @@ use base64::{Engine as _, engine::general_purpose::STANDARD};
 use jsonrpc_core::{BoxFuture, Error, Result, futures::future};
 use jsonrpc_derive::rpc;
 use solana_account::Account;
-use solana_account_decoder::UiAccountEncoding;
 use solana_client::rpc_response::{RpcLogsResponse, RpcResponseContext};
 use solana_clock::Slot;
 use solana_commitment_config::CommitmentConfig;
@@ -15,8 +14,9 @@ use solana_system_interface::program as system_program;
 use solana_transaction::versioned::VersionedTransaction;
 use spl_associated_token_account_interface::address::get_associated_token_address_with_program_id;
 use surfpool_types::{
-    ClockCommand, GetSurfnetInfoResponse, Idl, ResetAccountConfig, RpcProfileResultConfig,
-    SimnetCommand, SimnetEvent, StreamAccountConfig, UiKeyedProfileResult,
+    AccountSnapshot, ClockCommand, ExportSnapshotConfig, GetSurfnetInfoResponse, Idl,
+    ResetAccountConfig, RpcProfileResultConfig, SimnetCommand, SimnetEvent, StreamAccountConfig,
+    UiKeyedProfileResult,
     types::{AccountUpdate, SetSomeAccount, SupplyUpdate, TokenAccountUpdate, UuidOrSignature},
 };
 
@@ -27,7 +27,7 @@ use crate::{
         State,
         utils::{verify_pubkey, verify_pubkeys},
     },
-    surfnet::{GetAccountResult, locker::SvmAccessContext, svm::AccountFixture},
+    surfnet::{GetAccountResult, locker::SvmAccessContext},
     types::{TimeTravelConfig, TokenAccount},
 };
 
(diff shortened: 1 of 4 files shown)


crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -756,6 +758,54 @@ pub trait SurfnetCheatcodes {
         config: Option<ResetAccountConfig>,
     ) -> Result<RpcResponse<()>>;
 
+    /// A cheat code to export all accounts as fixtures for testing.
+    ///
+    /// ## Parameters
+    /// - `meta`: Metadata passed with the request, such as the client's request context.
```
lgalabru: Apologies, this param `meta` being here is a copy-pasta in the first place, could you please get rid of it?
This docstring is meant to be documenting the RPC endpoint, not the rust method.

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -756,6 +758,54 @@ pub trait SurfnetCheatcodes {
...
+    /// - `encoding` (optional): The encoding to use for account data. Defaults to Base64.
+    ///   - `Base64`: Returns raw account data as base64 encoded strings
+    ///   - `JsonParsed`: Attempts to parse known account types (tokens, programs with IDLs, etc.)
```
lgalabru: Is the case correct? I see that we're using `base64` in the example below.

crates/core/src/surfnet/svm.rs
```diff
@@ -1656,6 +1673,55 @@ impl SurfnetSvm {
...
+    /// # Arguments
+    /// * `encoding` - The encoding to use for account data (Base64, JsonParsed, etc.)
+    /// 
+    /// # Returns
+    /// A HashMap of pubkey -> AccountFixture that can be serialized to JSON
+    pub fn export_accounts_as_fixtures(
+        &self,
+        encoding: UiAccountEncoding
+    ) -> HashMap<String, AccountFixture> {
+        let mut fixtures = HashMap::new();
```
lgalabru: Can we use a `BTreeMap` instead?
Having the accounts in the same order, from one fixture export to another could be useful to track diffs.

Shradhesh71: Yes, we can. `BTreeMap` keeps the accounts in a consistent order by pubkey, which helps track diffs easily, slight performance hit `(O(log n) inserts vs O(1) in HashMap`) doesn’t matter here since fixture exports aren’t performance critical

Shradhesh71:
@lgalabru Pushed updates addressing all feedback

crates/core/src/surfnet/svm.rs
lgalabru: Thanks @Shradhesh71 !

lgalabru:
Thanks @Shradhesh71 ! Could you run the command
```
cargo +nightly fmt --all
```
on your machine? that should fix the CI!

I just added this button in the studio, looking forward to get this new cheat code in the next release :)

Screenshot 2025-10-15 at 06 39 35

lgalabru (approved):
Thanks so much for your contribution @Shradhesh71, the code looks good to me!

lgalabru:
hey @Shradhesh71! I talked with @MicaiahReid and it looks like your approach could be missing something. If a transaction is updating the state of an account, it looks like the current approach would not capture that.
Do you think you could look into this issue?

Shradhesh71:
> hey @Shradhesh71! I talked with @MicaiahReid and it looks like your approach could be missing something. If a transaction is updating the state of an account, it looks like the current approach would not capture that. Do you think you could look into this issue?

yep, I’ll look into why transactions that update account state aren’t being captured, reproduce the case, and push a follow up here

lgalabru:
@Shradhesh71 we'd love to get this feature wrapped-up for our incoming release, would you have some bandwidth to take it to the finish line or would you like us to takeover?
Thanks again!!

Shradhesh71:
yeah, everything looks perfect now and is ready for incoming release

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -746,6 +749,66 @@ pub trait SurfnetCheatcodes {
...
+    /// A cheat code to export a snapshot of all accounts in the Surfnet SVM.
+    ///
+    /// This method retrieves the current state of all accounts stored in the Surfnet Virtual Machine (SVM)
+    /// and returns them as a mapping of account public keys to their respective account snapshots.
+    ///
+    /// ## Parameters
+    /// - `config`: An optional `ExportSnapshotConfig` to customize the export behavior. The config fields are:
+    ///     - `includeParsedAccounts`: If true, includes parsed account data in the snapshot.
+    ///     - `filter`: An optional filter config to limit which accounts are included in the snapshot. Fields include:
+    ///         - `includeProgramAccounts`: A list of program IDs to include accounts for.
```
lgalabru, line 761: Copy pasta, this is a `bool`?
We know that at some point we'll differentiate local vs fetched, should we already introduce this distinction?  It could be an `enum` instead - `all` (supported with this version), (`remoteOnly` to be supported)

crates/types/src/types.rs
```diff
@@ -955,6 +955,54 @@ impl<K: std::hash::Hash + Eq, V> FifoMap<K, V> {
...
+#[serde(rename_all = "camelCase")]
+pub struct ExportSnapshotConfig {
+    pub include_parsed_accounts: Option<bool>,
+    pub filter: Option<ExportSnapshotFilter>,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ExportSnapshotFilter {
+    pub include_program_accounts: Option<bool>,
```
lgalabru, line 1001: `include_program_accounts` or `include_executable_accounts` ?

MicaiahReid: `include_executable_accounts` doesn't cover it - a buffer account or a programdata account (which stores the bulk of program data) isn't executable, but it's a program account (owned by one of the bpf loader programs)

PR #366, opened 2025-10-11
fix: preserve account order in getMultipleAccounts RPC method

serejke:
The `getMultipleAccounts` RPC endpoint was not preserving the request order when returning results. The RPC handler was iterating through account_updates in whatever order they were returned, rather than matching the original pubkeys array order.

Changes:
- Build a `HashMap` from pubkey to `GetAccountResult` to enable order-preserving lookup
- Add test to verify order preservation with remote client

This ensures the response array matches the request array order exactly, as required by the Solana RPC specification.

### DIFF
--- a/crates/core/src/rpc/accounts_data.rs
+++ b/crates/core/src/rpc/accounts_data.rs
@@ -1,9 +1,16 @@
+use super::{RunloopContext, SurfnetRpcContext};
+use crate::{
+    error::{SurfpoolError, SurfpoolResult},
+    rpc::{utils::verify_pubkey, State},
+    surfnet::locker::{is_supported_token_program, SvmAccessContext},
+    types::{MintAccount, TokenAccount},
+};
 use jsonrpc_core::{BoxFuture, Result};
 use jsonrpc_derive::rpc;
 use solana_account_decoder::{
-    UiAccount,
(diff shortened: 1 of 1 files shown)


crates/core/src/rpc/accounts_data.rs
```diff
@@ -437,25 +437,41 @@ impl AccountsData for SurfpoolAccountsDataRpc {
 
             svm_locker.write_multiple_account_updates(&account_updates);
```
lgalabru, line 438: Hmm interesting. It feels like the underlying method `get_multiple_accounts` should be returning the data correctly ordered by default. Should we be fixing that method instead of patching the downstream code?
cc @MicaiahReid

serejke: That's better for sure. I considered it but decided to preserve the semantics of the name `get_multiple_accounts_local_then_remote`. After checking the usages, the function must be renamed  preserve the order.

I opened an alternative PR #367 — feel free to choose which one to merge.

lgalabru: Awesome thanks @serejke, we'll look into this!

MicaiahReid:
Let's close this guy in favor of #367

PR #367, opened 2025-10-13
fix: preserve account order in `get_multiple_accounts_with_remote_fallback`

serejke:
The `getMultipleAccounts` RPC endpoint was returning accounts in incorrect order when mixing local and remote results. The bug was in the underlying `get_multiple_accounts_local_then_remote` method, which was appending all found accounts first, then all remote accounts, breaking the original request order.

Root cause:
- `get_multiple_accounts_local_then_remote` split results into found/missing arrays
- Combined results as `[all_found, all_remote]` instead of preserving original order
- Example: requesting [pk1, pk2, pk3] with pk2 missing would return [pk1, pk3, pk2]

Changes:
- Renamed `get_multiple_accounts_local_then_remote → get_multiple_accounts_with_remote_fallback`
- Fixed order preservation by explicitly iterating through original pubkeys array
- Optimized to avoid cloning accounts: only clone remote results, move found accounts
- Simplified RPC handler by removing HashMap workaround (no longer needed)
- Updated documentation to clarify order preservation guarantee

Implementation:
- Build remote_map only for missing accounts
- Zip through (pubkeys, local_results) to merge in correct order
- Early return path when all found locally now has zero clones

This fixes the issue at the source, benefiting all callers of the method, and ensures the response array matches the request array order exactly as required by the Solana RPC specification.

### DIFF
--- a/crates/core/src/rpc/accounts_data.rs
+++ b/crates/core/src/rpc/accounts_data.rs
@@ -1,16 +1,9 @@
-use super::{RunloopContext, SurfnetRpcContext};
-use crate::{
-    error::{SurfpoolError, SurfpoolResult},
-    rpc::{utils::verify_pubkey, State},
-    surfnet::locker::{is_supported_token_program, SvmAccessContext},
-    types::{MintAccount, TokenAccount},
-};
 use jsonrpc_core::{BoxFuture, Result};
 use jsonrpc_derive::rpc;
 use solana_account_decoder::{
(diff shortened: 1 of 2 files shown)


serejke:
Alternative to #366

MicaiahReid (approved):
Dude @serejke, this is awesome! Fixed a bug while improving performance AND included tests. 🏆 

Thanks!

PR #368, opened 2025-10-15
feat(core): revert get_account_local and get_multiple_accounts_local

### LINKED ISSUE #332: Revert `get_accounts_local` and `get_multiple_accounts`
wait for https://github.com/LiteSVM/litesvm/pull/218 to get merged so I can revert this patch: https://github.com/txtx/surfpool/pull/328/commits/813d255d267d37626dcf62e19dc37ce4cd898aca

vict0rcarvalh0:
LiteSVM got merged to update set_account to remove 0 lamports accounts. So get_account_local and get_multiple_accounts_local were reverted to the previous implementation. Re-ran surfpool-core integration tests that validate reset behavior:
- tests::integration::test_reset_account
- tests::integration::test_reset_account_cascade

Both passed successfully.

 LiteSVM reference: https://github.com/LiteSVM/litesvm/pull/218

Fixes #332

### DIFF
--- a/crates/core/src/surfnet/locker.rs
+++ b/crates/core/src/surfnet/locker.rs
@@ -230,28 +230,17 @@ impl SurfnetSvmLocker {
     pub fn get_account_local(&self, pubkey: &Pubkey) -> SvmAccessContext<GetAccountResult> {
         self.with_contextualized_svm_reader(|svm_reader| {
             match svm_reader.inner.get_account(pubkey) {
-                Some(account) => {
-                    // Treat zero-lamports or default accounts as deleted
-                    if account.eq(&Account::default()) || account.lamports == 0 {
-                        return GetAccountResult::None(*pubkey);
-                    }
-                    GetAccountResult::FoundAccount(
+                Some(account) => GetAccountResult::FoundAccount(
+                    *pubkey, account,
+                    // mark as not an account that should be updated in the SVM, since this is a local read and it already exists
+                    false,
+                ),
+                None => match svm_reader.get_account_from_feature_set(pubkey) {
+                    Some(account) => GetAccountResult::FoundAccount(
                         *pubkey, account,
                         // mark as not an account that should be updated in the SVM, since this is a local read and it already exists
                         false,
-                    )
-                }
-                None => match svm_reader.get_account_from_feature_set(pubkey) {
-                    Some(account) => {
-                        if account.eq(&Account::default()) || account.lamports == 0 {
-                            return GetAccountResult::None(*pubkey);
-                        }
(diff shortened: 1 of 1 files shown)


BretasArthur1:
Nice start @vict0rcarvalh0! You did a great job finding this patch...

Here some considerations

crates/core/src/surfnet/locker.rs
```diff
@@ -243,11 +242,16 @@ impl SurfnetSvmLocker {
                     )
                 }
                 None => match svm_reader.get_account_from_feature_set(pubkey) {
-                    Some(account) => GetAccountResult::FoundAccount(
-                        *pubkey, account,
-                        // mark as not an account that should be updated in the SVM, since this is a local read and it already exists
-                        false,
-                    ),
+                    Some(account) => {
+                        if account.eq(&Account::default()) || account.lamports == 0 {
```
BretasArthur1: A default instance would look like:

lamports: 0

data: Vec::new()

owner: Pubkey::default() (all zeros)

executable: false

rent_epoch: 0


So checking it's redundant looking for both since looking for a Account::default(), already haves account.lamports == 0, and in the case where a account have lamports == 0 it means that the account does not exist. And after the fix of LiteSVM where accounts with 0 lamports are excluded from the accounts_db

You should be able to just 

```rust
 Some(account) => GetAccountResult::FoundAccount(
                    *pubkey, account,
                    
                    false,
                ),
```

Because LiteSVM now handles the deleting of accounts with 0 lamports we should just try to find it and it will returns the account or false since it does not exists.

crates/core/src/surfnet/locker.rs
```diff
@@ -317,11 +319,17 @@ impl SurfnetSvmLocker {
                         }
                     }
                     None => match svm_reader.get_account_from_feature_set(pubkey) {
-                        Some(account) => GetAccountResult::FoundAccount(
-                            *pubkey, account,
-                            // mark as not an account that should be updated in the SVM, since this is a local read and it already exists
-                            false,
-                        ),
+                        Some(account) => {
+                            if account.eq(&Account::default()) || account.lamports == 0 {
```
BretasArthur1: Same as above here :)

BretasArthur1 (changes requested):

BretasArthur1:
What do you think @MicaiahReid ?

crates/core/src/surfnet/locker.rs
vict0rcarvalh0: Makes sense Arthur! Just made a new commit refactoring that

BretasArthur1 (approved):

BretasArthur1:
LGTM

Issue #330, opened 2025-09-15
Implement Missing RPC Admin Methods


lalitcap23:
Overview

Several administrative RPC endpoints in admin.rs currently return not_implemented_err instead of actual functionality. These endpoints are critical for managing Surfpool effectively, especially for debugging, monitoring, and plugin lifecycle management.

Current State

The following methods are placeholders and return errors instead of working logic:

- [x] reload_plugin (line 777)
- [x] unload_plugin (line 781)
- [x] list_plugins (line 805)
- [x] rpc_addr (line 809)
- [x] set_log_filter (line 813)
- [x] start_time (line 817)

MicaiahReid:
Ahh, good call @lalitcap23! Any interest in taking a stab at any of these?

lalitcap23:
@MicaiahReid  ,  would like pick this issue.

0xsouravm:
Is this issue still available? I would like to take on this.

lalitcap23:
@0xsouravm ,go ahead sir and tell u r working on which

MicaiahReid:
Yep @0xsouravm, it's available!

0xsouravm:
> [@0xsouravm](https://github.com/0xsouravm) ,go ahead sir and tell u r working on which

Picked the first one [`reload_plugin`]. Still new to Surfpool so digging around the repo a bit! @lalitcap23

0xsouravm:
I finished `reload_plugin` and `unload_plugin`. But I am unable to test it. Need an actual plugin config file. Is it available?

Will try to: `load` -> `reload` -> `unload`

@MicaiahReid @lalitcap23 ?

MicaiahReid:
@0xsouravm, we have an example geyser plugin, does that help? https://github.com/txtx/surfpool/tree/main/examples/hello-geyser

0xsouravm:
I think the admin RPCs like `load`, `unload` and `reload` are for the subgraph plugin only? I would need a valid subgraph plugin config to test those. How was the `load_plugin` method tested? @MicaiahReid

MicaiahReid:
I'm not sure if we have tests for `load_plugin` currently. If you open up a PR with what you have so far, I think I'll have an easier time helping you brainstorm how to test. Thanks @0xsouravm!

0xsouravm:
Gotcha! Bringing up the PR right away.

0xsouravm:
Please refer to PR #369 @MicaiahReid !

[...]

PR #369, opened 2025-10-15
feat: add reload_plugin and unload_plugin admin RPCs

0xsouravm:
Changes:
  - Added `unload_plugin(meta, name)` and `reload_plugin(meta, name, config_file)` RPC methods
  - Extended `PluginManagerCommand` enum with `UnloadPlugin` and `ReloadPlugin` variants
  - Created helper functions `load_subgraph_plugin()` and `unload_plugin_by_uuid()` to eliminate code duplication
  - Refactored the handler `LoadConfig` to use helpers and added `UnloadPlugin` and `ReloadPlugin` handler
  - Added `plugin_uuid_map: HashMap<Uuid, usize>` to track plugin indices for efficient unload/reload
  - Added `DestoryCollection` command to `SubgraphCommand` to destroy associated collection after unloading a plugin
  - Added `unregister_collection` and `remove_collection` methods to use in `DestroyCollection` command
  
  Reference: #330

### DIFF
--- a/crates/cli/src/http/mod.rs
+++ b/crates/cli/src/http/mod.rs
@@ -317,8 +317,8 @@ fn start_subgraph_runloop(
                                 lookup.add_collection(metadata);
                                 gql_schema.replace(new_dynamic_schema(lookup.clone()));
 
-                                let endpoint_url = format!("http://localhost:8899/subgraph/{}", uuid);
-                                let _ = sender.send(endpoint_url);
+                                let console_url = format!("{}/accounts", config.studio_url.clone());
+                                let _ = sender.send(console_url);
                             }
                             SubgraphCommand::ObserveCollection(subgraph_observer_rx) => {
                                 observers.push(subgraph_observer_rx);


lgalabru:
That's awesome, thanks @0xsouravm! 
Have you been able to test this additions with Yellowstone or any other Geyser plugin?

0xsouravm:
No worries @lgalabru !!
I am actually struggling with the config file. I am trying to `curl` but I cannot get the config JSON right. Hence the serde parsing fails during `load_plugin`

lgalabru:
@0xsouravm what's the issue you're seeing? if you're using Yellowstone, I think there's an issue with the proposed config file from the example in the repo - something with the plugin name that should not be there or something.

0xsouravm:
@lgalabru I was trying with the example config as @MicaiahReid suggested but the config JSON doesn't sit correctly. Can you tell me how `load_config` was previously tested? I could get the others working if I get the `load_config` working.

MicaiahReid:
Apologies on the latency for this PR, @0xsouravm 🙏 

I spent some time testing this today. Here are the steps to test:
 1. Build an anchor project with subgraphs set up. This example project would work: https://github.com/txtx/surfpool-examples/tree/main/anchor/simple-pda
 2. Start surfpool in the root of that project with subgraphs
 3. You'll see the subgraphs deployed and a UUID for the subgraph
 4. Send the uuid in an `unloadPlugin` RPC method

I found that this does not work out of the box, but actually crashes surfpool.

crates/core/src/rpc/admin.rs
```diff
@@ -798,7 +854,9 @@ impl AdminRpc for SurfpoolAdminRpc {
         let Ok(endpoint_url) = rx.recv_timeout(Duration::from_secs(10)) else {
             return Box::pin(async move { Err(jsonrpc_core::Error::internal_error()) });
         };
-        Box::pin(async move { Ok(endpoint_url) })
+        // Return a JSON string containing both UUID and endpoint URL
+        let response = format!(r#"{{"uuid": "{}", "endpoint": "{}"}}"#, uuid, endpoint_url);
```
MicaiahReid: Previously, the return result would yield the TUI logging:
image

Now, the TUI logs this, which doesn't really make much sense:
image

If we're going to log the UUID, it should be via `simnet_events_tx.try_send(SimnetEvent::info(...))`, and return the endpoint_url for this fn

0xsouravm:
Hey @MicaiahReid ! No worries and thank you for guiding me how to test it. I ran into the same issues as you and now I fixed those and `unload_plugin` works as intended! 

The issue - We weren't destroying the associated collection after unloading the plugin
The fix - Added a new subgraph command that handles destroying collections upon unload.

Screenshot 2025-11-13 at 9 20 22 AM

Please check out the above screenshot for the logs!

crates/core/src/rpc/admin.rs
0xsouravm: I added logging through `simnet_events_tx` as you mentioned!

crates/core/src/runloops/mod.rs
```diff
@@ -471,6 +475,93 @@ fn start_geyser_runloop(
...
+        let load_subgraph_plugin = |uuid: uuid::Uuid,
+                                      config: txtx_addon_network_svm_types::subgraph::PluginConfig,
+                                      surfpool_plugin_manager: &mut Vec<Box<dyn GeyserPlugin>>,
+                                      plugin_uuid_map: &mut HashMap<uuid::Uuid, usize>,
+                                      indexing_enabled: &mut bool|
+         -> Result<String, String> {
+            let _ = subgraph_commands_tx.send(SubgraphCommand::CreateCollection(
+                uuid,
+                config.data.clone(),
+                crossbeam_channel::bounded(0).0,  // Temporary sender, will be replaced
```
MicaiahReid: This will need to be fixed before merge, you should be using the `notifier` passed into `PluginManagerCommand::LoadConfig`

Issue #100, opened 2025-05-10
Introduce `surfnet_streamAccount`


lgalabru:
With this cheat code, developers would start a routine updating a given account by pulling from the remote, either on a regular basis (we update every second) or lazily (the account is being updated every time it's being read accessed).

MicaiahReid:
I have some concerns about this that I'd love to get community feedback on.

Whenever an account is pulled from the remote and inserted in the local state, in many cases this implies that the pulled account will be mutated in some way.

If we are regularly pulling an account from the remote and using the remote data to update our local state, there is a high risk that we will be overwriting state changes that the user is applying to the SVM via their own transactions.

For example:
 1. A user has a program deployed to mainnet
 2. Every time the program is invoked, the program's state is updated (say the program tracks calls to the protocol, or more likely, the program tracks the protocol's NAV)
 3. The user configures their surfnet to `surfnet_streamAccount(<program>)`, to pull updates for the program from mainnet
 4. The user sends a transaction to their surfnet which calls the program
    a. Surfpool sees a transaction referencing the program, which it is always supposed to update from mainnet
    b. Surfpool pulls the program from mainnet, getting the current state
    c. The user's transaction alters the state of the program
 5. The user sends another transaction to the program, expecting to have the state changes that were made in the last transaction be reflected in the program state
    a. Surfpool sees a transaction referencing the program, which it is always supposed to update from mainnet
    b. Surfpool pulls the program from mainnet, getting the current state
    c. The program state no longer reflects the changes made by the user's transaction
    d. The user's transaction alters the state of the program again
 6. The user does not find expected changes to program state

This would be a difficult bug to track in the user's expected behavior, so a major footgun.

---

Rather, I think the approach we will have to take is to stream all _transactions_ that _invoke_ an account, and replay those transactions on a surfnet.

To play out the same example:
 1. A user has a program deployed to mainnet
 2. Every time the program is invoked, the program's state is updated (say the program tracks calls to the protocol, or more likely, the program tracks the protocol's NAV)
 3. The user sends a transaction to their surfnet which calls the program
    a. Surfpool pulls the program from mainnet, getting the current state
    b. The user's transaction alters the state of the program
 4. Every time the program is invoked on mainnet, the mainnet transactions are forwarded to the surfnet to be played out
    a. Each of these transactions also alters the program's state, according the the program's logic
 5. The user sends another transaction to the program, expecting to have the state changes that were made in the last transaction be reflected in the program state
[...]

PR #371, opened 2025-10-16
feat: introduce cheat-code `surfnet_streamAccount`

lgalabru:
Address #100


crates/core/src/surfnet/svm.rs
```diff
@@ -626,12 +628,44 @@ impl SurfnetSvm {
...
+        self.purge_account_from_cache(&account, pubkey)?;
+        Ok(())
+    }
+
+    fn purge_account_from_cache(
+        &mut self,
+        account: &Account,
+        pubkey: &Pubkey,
+    ) -> SurfpoolResult<()> {
+        // Get the existing account if it exists
```
```suggestion
```

PR #379, opened 2025-10-19
feat(cli): solve the no studio flag

zsh28:
## Problem:
The --no-studio flag only affected which URL was displayed in the TUI, but the HTTP server always served studio static files regardless of the flag.

## Solution:

• Modified start_subgraph_and_explorer_server() to accept an enable_studio parameter
• Conditionally add the studio static files service only when !cmd.no_studio
• The HTTP server still runs (for GraphQL endpoints), but studio UI returns 404 when --no-studio is used

### DIFF
--- a/crates/cli/src/http/mod.rs
+++ b/crates/cli/src/http/mod.rs
@@ -96,11 +96,11 @@ pub async fn start_subgraph_and_explorer_server(
                     .route("/v1/graphql", web::post().to(post_graphql))
                     .route("/v1/subscriptions", web::get().to(subscriptions)),
             );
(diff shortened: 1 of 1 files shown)


lgalabru:
@zsh28 could you run
```
cargo +nightly fmt --all
```
and push your changes? thanks!

zsh28:
> @zsh28 could you run
> 
> ```
> cargo +nightly fmt --all
> ```
> 
> and push your changes? thanks!

on it!

zsh28:
@lgalabru  done!

PR #382, opened 2025-10-21
feat: introduce cheatcode resetNetwork

lgalabru:


crates/core/src/surfnet/svm.rs
```diff
@@ -651,6 +651,51 @@ impl SurfnetSvm {
...
+            .unwrap();
+        let parsed_mint_account = MintAccount::unpack(&native_mint_account.data).unwrap();
+
+        // Load native mint into owned account and token mint indexes
+        let accounts_by_owner = HashMap::from([(
+            native_mint_account.owner,
+            vec![spl_token_interface::native_mint::ID],
+        )]);
+        let token_mints =
+            HashMap::from([(spl_token_interface::native_mint::ID, parsed_mint_account)]);
```
MicaiahReid, line 700: Think it's worth putting this in a function that's used by both `new`/`reset_network` to DRY this up?

crates/core/src/surfnet/locker.rs
```diff
@@ -1675,6 +1675,15 @@ impl SurfnetSvmLocker {
         })
     }
 
+    /// Resets SVM state
+    ///
+    /// This function coordinates the reset of accounts by calling the SVM's reset_account method.
+    pub fn reset_network(&self) -> SurfpoolResult<()> {
+        let simnet_events_tx = self.simnet_events_tx();
+        let _ = simnet_events_tx.send(SimnetEvent::info("Network will be reset"));
```
```suggestion
        let _ = simnet_events_tx.send(SimnetEvent::info("Resetting network..."));
```
Not a strong opinion, this just felt sterile

crates/core/src/surfnet/svm.rs
lgalabru: I thought about it, we already have a `new` and `initialize` functions, let's keep it in our visor and pull the trigger if it gets more complex?

PR #387, opened 2025-10-23
fix: use block height instead of slot for `lastValidBlockHeight` in `getLatestBlockhash`

serejke:
The `getLatestBlockhash` RPC endpoint was incorrectly calculating `lastValidBlockHeight` using the slot number instead of the actual block height. This caused the returned value to be orders of magnitude larger than expected.

For example, with slot 375,249,725 and block height 561,160, it was returning `lastValidBlockHeight` as 375,249,994 instead of the correct ~561,460.

Changes:
- Updated `get_latest_blockhash` to use `svm_locker.get_epoch_info().block_height` instead of `committed_latest_slot` when calculating `lastValidBlockHeight`
- Updated all related test cases to use the correct block height calculation
- Removed unused committed_slot variables from tests

### DIFF
--- a/crates/core/src/rpc/full.rs
+++ b/crates/core/src/rpc/full.rs
@@ -2152,8 +2152,8 @@ impl Full for SurfpoolFullRpc {
             .get_latest_blockhash(&commitment)
             .unwrap_or_else(|| svm_locker.latest_absolute_blockhash());
 
-        let last_valid_block_height =
-            committed_latest_slot + MAX_RECENT_BLOCKHASHES_STANDARD as u64;
+        let current_block_height = svm_locker.get_epoch_info().block_height;
+        let last_valid_block_height = current_block_height + MAX_RECENT_BLOCKHASHES_STANDARD as u64;
         Ok(RpcResponse {
             context: RpcResponseContext::new(svm_locker.get_latest_absolute_slot()),
             value: RpcBlockhash {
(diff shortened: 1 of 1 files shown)


serejke:
The actual expiration policy is time based (30 seconds)
https://github.com/txtx/surfpool/blob/0d2a3b69294c18d6e449011f24617642326e6f9d/crates/core/src/surfnet/svm.rs#L90

https://github.com/txtx/surfpool/blob/0d2a3b69294c18d6e449011f24617642326e6f9d/crates/core/src/surfnet/svm.rs#L505

but the `MAX_RECENT_BLOCKHASHES_EXTERNAL` is equal to 500 blocks (~3.33 minutes). 

Looks like an overkill: the clients will have to wait for 3.33 minutes for true expiration, however it is enough to wait for ~75 slots.

@MicaiahReid

MicaiahReid:
> The actual expiration policy is time based (30 seconds)
> 
> https://github.com/txtx/surfpool/blob/0d2a3b69294c18d6e449011f24617642326e6f9d/crates/core/src/surfnet/svm.rs#L90
> 
> https://github.com/txtx/surfpool/blob/0d2a3b69294c18d6e449011f24617642326e6f9d/crates/core/src/surfnet/svm.rs#L505
> 
> but the `MAX_RECENT_BLOCKHASHES_EXTERNAL` is equal to 500 blocks (~3.33 minutes).
> 
> Looks like an overkill: the clients will have to wait for 3.33 minutes for true expiration, however it is enough to wait for ~75 slots.
> 
> @MicaiahReid


Yeah, unfortunately we have to keep this. For some reason, the `@solana/web.js` npm package expects blockhashes returned within 30 seconds to be valid, rather than polling for current block height compared to the last valid block height.

Since Surfpool supports "transaction block production mode" - where a new block is produced every time a tx is sent rather than every 400ms - how many blockhashes that are stored at once is a moving target. How many will fit in a 30s period depends on how quickly you're sending transactions.

I can investigate some other approaches to this.

PR #389, opened 2025-10-24
attempt AWS upload

MicaiahReid:


lgalabru:
Reminder for this PR, we want to either hard code the region or declare it as an ENV var instead of a secret.

PR #392, opened 2025-10-28
feat: scenarios

lgalabru:


crates/cli/src/http/mod.rs
```diff
@@ -263,10 +313,14 @@ fn start_subgraph_runloop(
...
-                                gql_schema.replace(new_dynamic_schema(collections_metadata_lookup.clone()));
 
-                                let console_url = format!("{}/subgraphs", config.studio_url.clone());
+                                let mut lookup = collections_metadata_lookup.write().map_err(|_| {
+                                    format!("{err_ctx}: Failed to acquire write lock on collections metadata lookup")
+                                })?;
+                                lookup.add_collection(metadata);
+                                gql_schema.replace(new_dynamic_schema(lookup.clone()));
+
+                                let console_url = format!("{}/accounts", config.studio_url.clone());
```
MicaiahReid, line 320: Hmm, don't know if I love `/accounts`, cause we could still be indexing events

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -883,6 +883,45 @@ pub trait SurfnetCheatcodes {
...
+    /// ## Example Response
+    /// ```json
+    /// {
+    ///   "jsonrpc": "2.0",
+    ///   "result": {
+    ///     "context": {
+    ///       "slot": 123456789,
+    ///       "apiVersion": "2.3.8"
+    ///     },
+    ///     "value": null
```
MicaiahReid: I don't know how helpful this example response is if the actual value is just null 😆

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -883,6 +883,45 @@ pub trait SurfnetCheatcodes {
...
+    ///
+    /// ## Example Request
+    /// ```json
+    /// {
+    ///   "jsonrpc": "2.0",
+    ///   "id": 1,
+    ///   "method": "surfnet_streamAccount",
+    ///   "params": [ "4EXSeLGxVBpAZwq7vm6evLdewpcvE2H56fpqL2pPiLFa", { "includeOwnedAccounts": true } ]
+    /// }
+    /// ```
```
MicaiahReid, line 903: Wrong method here!

crates/core/src/scenarios/registry.rs
```diff
@@ -0,0 +1,85 @@
...
+        };
+
+        // Convert all templates in the collection
+        let templates = collection.to_override_templates(idl);
+
+        // Register each template
+        for template in templates {
+            let template_id = template.id.clone();
+            self.templates.insert(template_id.clone(), template);
+            debug!("  Registered template: {}", template_id);
```
MicaiahReid: Don't mind this being here, but the space at the beginning makes me think you meant to delete this

crates/types/src/scenarios.rs
```diff
@@ -0,0 +1,370 @@
...
+    /// Human-readable name
+    pub name: String,
+    /// Description of what this template does
+    pub description: String,
+    /// Protocol this template is for (e.g., "Pyth", "Switchboard")
+    pub protocol: String,
+    /// IDL for the account structure - defines all available fields and types
+    pub idl: Idl,
+    /// How to determine the account address
+    pub address: AccountAddress,
```
MicaiahReid, line 53: This `AccountAddress` type seems unnecessary to me - what am I missing? I'd think it could just be address.

The protocol with this IDL will be deployed to an address, and the accounts it generates will have their PDA derivations defined in the IDL.

I'm not sure of protocols themselves being deployed to PDAs.

crates/core/src/surfnet/locker.rs
```diff
@@ -351,6 +391,10 @@ impl SurfnetSvmLocker {
                 local_results,
             ));
         }
+        debug!(
+            "Missing accounts will be fetched: {}",
+            missing_accounts.iter().join(", ")
+        );
```
MicaiahReid, line 364: I'm good with us adding better debug/trace logging, but this doesn't give very good context imo, did you mean to delete?

crates/core/src/surfnet/locker.rs
```diff
@@ -377,7 +421,10 @@ impl SurfnetSvmLocker {
                             .cloned()
                             .unwrap_or(GetAccountResult::None(*pubkey))
                     }
-                    found => found, // Keep found accounts (no clone, just move)
+                    found => {
+                        debug!("Keeping local account: {}", pubkey);
```
MicaiahReid, line 392: Intentional log?

crates/core/src/surfnet/locker.rs
```diff
@@ -862,6 +909,7 @@ impl SurfnetSvmLocker {
     ) -> SurfpoolResult<KeyedProfileResult> {
         let signature = transaction.signatures[0];
 
+        // Can we avoid this write?
```
```suggestion
```
How? Notify subscribers sooner?

crates/core/src/surfnet/locker.rs
```diff
@@ -890,12 +938,18 @@ impl SurfnetSvmLocker {
...
+            transaction.get_signature(),
+            transaction_accounts.iter().join(", ")
+        );
 
         let account_updates = self
             .get_multiple_accounts(remote_ctx, &transaction_accounts, None)
             .await?
             .inner;
 
+        // I don't think this code is required
```
```suggestion
```
I don't think YOU'RE required!

crates/core/src/surfnet/locker.rs
```diff
@@ -890,12 +938,18 @@ impl SurfnetSvmLocker {
                     .map(|l| l.all_loaded_addresses()),
             )
             .clone();
+        debug!(
+            "Transaction {} accounts inputs: {}",
+            transaction.get_signature(),
+            transaction_accounts.iter().join(", ")
+        );
```
MicaiahReid, line 911: Intentional log?

crates/core/src/surfnet/locker.rs
```diff
@@ -1048,137 +1102,129 @@ impl SurfnetSvmLocker {
...
+                    .collect::<Vec<_>>()
+            );
+            eprintln!(
+                "DEBUG: After retain - post keys = {:?}",
+                profile_result
+                    .post_execution_capture
+                    .keys()
+                    .map(|p| p.to_string())
+                    .collect::<Vec<_>>()
+            );
```
MicaiahReid: logs

crates/core/src/surfnet/locker.rs
lgalabru: Yeah, I like this log a lot

crates/core/src/surfnet/locker.rs
lgalabru: Yeah, I think we should start adding debug logs that can help us diagnose tricky edge cases faced by developers.
Account reconciliation is a category of leads.

Issue #391, opened 2025-10-25
Feature request: stealing core NFTs


IVSOP:
I would love to see a feature that allows changing ownership of core collections and NFTs. I have implemented this through the get/set account endpoints, but a native feature would be useful to more people: https://github.com/IVSOP/nft_robber

This could be used to test programs that interact with NFTs with an exact copy of a mainnet NFT, minimizing issues due to small nuances when minting.

To do this, I am deserializing the header using borsh, changing the owner/authority fields, then serializing it again while not touching the bytes after the header. This is trivial even without borsh, as the owner and authority fields in the headers are always at the same offset.

Relevant structs for the headers:
https://docs.rs/mpl-core/latest/mpl_core/accounts/struct.BaseAssetV1.html
https://docs.rs/mpl-core/latest/mpl_core/accounts/struct.BaseCollectionV1.html

MicaiahReid:
Thanks for the issue! If I'm understanding this correctly, it's kind of like the existing [`surfnet_setTokenAccount` cheatcode](https://docs.surfpool.run/rpc/cheatcodes#surfnet-settokenaccount), but for updating mint accounts - is that right? `surfnet_setMintAccount`?

IVSOP:
Not exactly. Currently surfpool has cheatcodes for tokens, and I would like a similar feature but for [core NFTs](https://developers.metaplex.com/core) and collections, so not really a mint like USDC etc. [Here](https://core.metaplex.com/explorer/2mNhQ5GHcn2y4CtR32LSKvhNzRumeVH1PMkFdACwqtNz) is an example of a random core NFT. With the proposed feature, you would be able to take ownership of this NFT.

MicaiahReid:
I see, got ya. Are you interested in opening a PR for this?

IVSOP:
Sure! Could you just quickly explain what your overall workflow of adding a new endpoint is? Are the website docs automatically generated? What about adding documentation [here](https://github.com/txtx/surfpool/blob/main/crates/types/src/rpc_endpoints.json)?
Also, running tests for the core crate shows that `test_simulate_add_alt_entries_fetching` and `test_add_alt_entries_fetching` failed. Is there some specific setup needed before running the tests?

PR #394, opened 2025-10-29
Feat: steal core NFTs and collections

IVSOP:
Adds feature described in #391, allowing changing ownership of a core NFT and authority of a core collection, through the new `surfnet_stealCore` endpoint.

Example request, stealing [this](https://core.metaplex.com/explorer/2mNhQ5GHcn2y4CtR32LSKvhNzRumeVH1PMkFdACwqtNz) random core NFT
```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "surfnet_stealCore",
  "params": [
    "2mNhQ5GHcn2y4CtR32LSKvhNzRumeVH1PMkFdACwqtNz",
    "<key of the new owner>"
  ]
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5198,6 +5198,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "mpl-core"
+version = "0.11.0"
+source = "git+https://github.com/ivsop/mpl-core?rev=27bbab1997fdd39fd010d78406cc7ee2e7e2601d#27bbab1997fdd39fd010d78406cc7ee2e7e2601d"
+dependencies = [
+ "base64 0.22.1",
+ "borsh 0.10.4",
+ "kaigan",
+ "modular-bitfield",
+ "num-derive 0.3.3",
+ "num-traits",
+ "rmp-serde",
+ "serde",
+ "serde_json",
+ "serde_with",
+ "solana-program",
+ "thiserror 1.0.69",
+]
+
+[[package]]
 name = "multimap"
 version = "0.8.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


MicaiahReid:
Thanks for the PR, @IVSOP! This helps me understand the feature better.

Thoughts @lgalabru and @IVSOP on this PR as-is vs two cheatcodes:
 - `surfnet_setMplCollection` - sets data according to this struct https://docs.rs/mpl-core/latest/mpl_core/accounts/struct.BaseCollectionV1.html
 - `surfnet_setMplAsset` - sets data according to this struct https://docs.rs/mpl-core/latest/mpl_core/accounts/struct.BaseAssetV1.html

This would make each endpoint more "full-featured", like `surfnet_setAccount`/`surfnet_setTokenAccount`

IVSOP:
I didn't split the endpoint because the logic is the exact same (just write a key into `data[1..33]`). If you want to set the full data, then I think having two separate cheatcodes would be cleaner. as the arguments and logic would be slightly different.
In the future, if you want to also change the plugin information, then it would make even more sense to split into two cheatcodes, so this is future proofing as well

lgalabru:
What would be the next steps on this PR @IVSOP @MicaiahReid ?

IVSOP:
I've changed the endpoint from `stealCore` to `updateCoreAsset` and allowed changing every part of the header, and optionally deleting all the plugins.
I had to drag in the mpl-core crate because it was too cumbersome to replicate the functionalities and types without relying on it. However, I had many challenges with dependency conflicts and with the fact that its autogenerated client interface doesn't expose all of the functionality I need / exposes it in a way that is very hard to work with.
I ended up forking the crate and adding very minimal functionality so I could avoid the dependency conflicts. The code for this endpoint ended up being very large and might be hard to follow, due to the unergonomic data the crate returns, and having to manually implement serialization and deserialization of the plugins.

In short, the plugins use a byte offset system. An offset of `N` means some plugin is located at `bytes[N..M]`. If we change the owner in the header, its length does not change as pubkeys have a static size. However, if any of the strings change, we have to re-serialize the plugins so that they all have the correct offsets.

I'd like to hear your thoughts before I implement a similar endpoint for core collections and not just assets.
New example:
```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "surfnet_updateCoreAsset",
    "params": [
      "2mNhQ5GHcn2y4CtR32LSKvhNzRumeVH1PMkFdACwqtNz",
      {
        "owner": "<<<some new owner pubkey>>>",
        "name": "some new name",
      }
    ]
  }
  ```
  
  You can test this for yourself using the NFT viewer I provided in the top message and using the localhost mode.

MicaiahReid:
Maybe we'll revisit latter, but closing for now

PR #408, opened 2025-11-15
feat: add list_plugins admin RPC and remove rpc_addr admin RPC method from trait

0xsouravm:
Changes:

- Added `list_plugins(meta)` RPC method definition
- Extended PluginManagerCommand enum with `ListPlugins` variant
- Updated `plugin_uuid_map: HashMap<Uuid, usize>` into `plugin_map: HashMap<Uuid, (usize, String)>to store plugin names along with UUID for efficient listing
- Removed `rpc_addr` method from the trait

Reference: #330


crates/core/src/runloops/mod.rs
```diff
@@ -599,18 +599,32 @@ fn start_geyser_runloop(
...
-                                    if let Err(e) = load_subgraph_plugin(uuid, config, notifier, &mut surfpool_plugin_manager, &mut plugin_uuid_map, &mut indexing_enabled) {
+                                    if let Err(e) = load_subgraph_plugin(uuid, config, notifier, &mut surfpool_plugin_manager, &mut plugin_map, &mut indexing_enabled) {
                                         let _ = simnet_events_tx.try_send(SimnetEvent::error(format!("Failed to reload plugin: {}", e)));
                                     }
                                 }
+                                #[cfg(not(feature = "subgraph"))]
+                                PluginManagerCommand::ListPlugins(_) => {
+                                    continue;
+                                }
+                                #[cfg(feature = "subgraph")]
```
MicaiahReid:
```suggestion
```
I don't think we need to feature gate this - we can just return an empty list if not enabled

crates/core/src/lib.rs
```diff
@@ -52,11 +52,18 @@ pub async fn start_local_surfnet(
     .await
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct PluginInfo {
```
MicaiahReid, line 57:
```suggestion
#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(rename_all = "camelCase")]
pub struct PluginInfo {
```
for most user-interfacing structs, we'll have camel case

MicaiahReid (changes requested):
As I was reviewing this PR, I realized the `rpc_addr` method is quite a silly one 😂 

You need to know the network's rpc address in order to fetch it's rpc address via rpc... so why do we need the method? hahaha

Would you be able to remove the changes from that method, and that method from the `AdminRpc` trait completely?

0xsouravm:
> As I was reviewing this PR, I realized the `rpc_addr` method is quite a silly one 😂
> 
> You need to know the network's rpc address in order to fetch it's rpc address via rpc... so why do we need the method? hahaha
> 
> Would you be able to remove the changes from that method, and that method from the `AdminRpc` trait completely?

When put that way, it does sound funny. 😹 . Sure I will remove all instances of the method and the structure field I added to tests and elsewhere!

0xsouravm:
@MicaiahReid I have added the changes you requested. Please take a look!

MicaiahReid (commented):
LGTM, thanks @0xsouravm!!

PR #410, opened 2025-11-15
feat: set_log_filter rpc and start_time rpc

JkrishnaD:
### Changes

- Added `set_log_filter` admin RPC for updating the system log level.
- Added `start_time` admin RPC to return the runloop start timestamp.
- Added corresponding `PluginManagerCommand` variants.
- Implemented handlers inside the geyser runloop.
- Updated admin RPC trait and implementations.

Issue: #330

### DIFF
--- a/crates/core/src/rpc/admin.rs
+++ b/crates/core/src/rpc/admin.rs
@@ -1,8 +1,4 @@
-use std::{
-    collections::HashMap,
-    net::SocketAddr,
-    time::{Duration, SystemTime},
-};
+use std::{collections::HashMap, net::SocketAddr, time::Duration};
 
 use jsonrpc_core::{BoxFuture, Result};
 use jsonrpc_derive::rpc;
@@ -219,7 +215,7 @@ pub trait AdminRpc {
     /// ```
     ///
     /// # Notes
-    /// - The result is a `SystemTime` in UTC, reflecting the moment the system was initialized.
+    /// - The result is a `String` in UTC, reflecting the moment the system was initialized.
     /// - This method is useful for monitoring system uptime and verifying system health.
     #[rpc(meta, name = "startTime")]
     fn start_time(&self, meta: Self::Metadata) -> Result<String>;


jkrishnad:
> Hey @JkrishnaD, thanks for the PR!
> 
> I've got a few notes:
> 
> #### Remove `set_log_filter` method
> Sadly, I think we should remove the `set_log_filter` admin method. I don't see it as particularly valuable to the user, and our log management is a bit more complicated than you have here. We manage logs for the CLI and a written log, so we have a designated configuration on startup of the simnet. I think what we have at the CLI level is sufficient, so we can just remove this admin RPC method completely. (both from the implementation and the `AdminRpc` trait.
> 
> #### Fixes for `start_time`
> The current implementation of the `start_time` is gated behind the `subgraph` feature, which is unnecessary. Actually, it doesn't need to be using the `PluginManagerCommand` at all. The use of an `rx/tx` pair to send a command to get a time back is necessary. I think a better approach is:
> 
> 1. Add a `start_time` field to the `SurfnetSvm` struct in the `svm.rs` file
> 2. Add a function to the `SurfnetSvmLocker` structure (in `locker.rs`) to get this value
> 3. Have something like the following in the `start_time` admin RPC function to just directly fetch the time:
> 
> ```rust
> let svm_locker = meta.get_svm_locker()?;
> let start_time = svm_locker.get_start_time();
> Ok(start_time)
> ```

@MicaiahReid I’ve applied all the requested changes:

- removed the `set_log_filter` RPC (trait + implementation + command)
- implemented `start_time` using the direct getter on `SurfnetSvmLocker`
- added `start_time` to the `SurfnetSvm` struct as suggested

Please take another look when you get a chance. Thanks for the clear guidance!

crates/core/src/runloops/mod.rs
```diff
@@ -562,6 +564,8 @@ fn start_geyser_runloop(
 
             Ok(())
         };
+        #[cfg(feature = "subgraph")]
+        let system_start_time = SystemTime::now();
```
MicaiahReid:
```suggestion
```

crates/core/src/runloops/mod.rs
```diff
@@ -1,5 +1,7 @@
 #![allow(unused_imports, dead_code, unused_mut, unused_variables)]
 
+#[cfg(feature = "subgraph")]
+use std::time::SystemTime;
```
MicaiahReid:
```suggestion
```

crates/core/src/rpc/admin.rs
```diff
@@ -902,12 +863,10 @@ impl AdminRpc for SurfpoolAdminRpc {
...
-    fn set_log_filter(&self, _filter: String) -> Result<()> {
-        not_implemented_err("set_log_filter")
-    }
-
-    fn start_time(&self, _meta: Self::Metadata) -> Result<SystemTime> {
-        not_implemented_err("start_time")
+    fn start_time(&self, meta: Self::Metadata) -> Result<SystemTime> {
+        let svm_locker = meta.get_svm_locker()?;
+        let start_time = svm_locker.get_start_time();
+        Ok(start_time)
```
MicaiahReid: Ahh I just tested and realized this output is wrong. When you fetch, it's returning:
```
{
        "secs_since_epoch": 1763659511,
        "nanos_since_epoch": 735414000
    }
```

This isn't the most ergonomic return value, can we return something that's more like the example return value in the trait? Something like `2025-04-24T12:34:56Z`

jkrishnad: For this i changed the method return type from `SystemTime` to `String` so that we can return the time in string format

crates/core/src/runloops/mod.rs
jkrishnad: sry, by mistake these were added there

crates/core/src/runloops/mod.rs
jkrishnad: I forgot to remove these while fixing

jkrishnad:
@MicaiahReid ser check the implementation now

PR #417, opened 2025-11-22
feat: add runtime feature toggling via CLI and RPC

### LINKED ISSUE #308: Allow [de]activating feature sets via config/cli command
Would be cool if users could experiment with different feature sets being activated + deactivated.

AvhiMaz:
Implement comprehensive feature toggling system allowing users to experiment with different feature sets at runtime without restarting.

 ### Changes:
  - Add Admin RPC methods: setInstructionProfiling, setMaxProfiles,
    setLogBytesLimit, getFeatureStates
  - Add CLI commands: `surfpool feature` and `surfpool plugin` for
    runtime management
  - Add runtime setters in SurfnetSvm and SurfnetSvmLocker for
    thread-safe state updates
  - Add FeatureStates struct to query current configuration
  - Support multiple input formats for boolean values (true/false,
    yes/no, on/off, 1/0)
  - Track log_bytes_limit in SurfnetSvm to make it queryable

  Closes #308

### DIFF
--- a/crates/cli/src/cli/mod.rs
+++ b/crates/cli/src/cli/mod.rs
@@ -491,9 +491,9 @@ pub struct FeatureGetCommand {
 
 #[derive(Parser, PartialEq, Clone, Debug)]
 pub struct FeatureProfilingCommand {
-    /// Enable or disable profiling (true/false)
+    /// Enable or disable profiling
     #[arg(value_name = "ENABLED")]
-    pub enabled: String,
+    pub enabled: bool,
     /// RPC URL of the running surfpool instance
     #[arg(long = "rpc-url", default_value = "http://localhost:8899")]
     pub rpc_url: String,
@@ -754,19 +754,9 @@ async fn handle_plugin_commands(cmd: PluginCommand) -> Result<(), String> {
     match cmd {
         PluginCommand::List(cmd) => {
             let result = rpc_call(&cmd.rpc_url, "listPlugins", serde_json::json!([]))?;
-            match result.get("result") {
-                Some(plugins) => {
-                    println!("{}", serde_json::to_string_pretty(plugins).unwrap());
-                    Ok(())
-                }
-                None => {
-                    if let Some(error) = result.get("error") {
-                        Err(format!("RPC error: {}", error))
-                    } else {
-                        Err("Unexpected RPC response".to_string())
-                    }
-                }
-            }
+            let plugins = handle_rpc_response(result)?;
+            println!("{}", serde_json::to_string_pretty(&plugins).unwrap());
+            Ok(())
         }
         PluginCommand::Load(cmd) => {
             let config_content = std::fs::read_to_string(&cmd.config_file)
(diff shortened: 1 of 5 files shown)


AvhiMaz:
hi @MicaiahReid let me know what you think about this implementation.

crates/cli/src/cli/mod.rs
```diff
@@ -633,6 +750,232 @@ async fn handle_cloud_commands(cmd: CloudCommand) -> Result<(), String> {
...
+    match cmd {
+        FeatureCommand::Get(cmd) => {
+            let result = rpc_call(&cmd.rpc_url, "getFeatureStates", serde_json::json!([]))?;
+            match result.get("result") {
+                Some(states) => {
+                    println!("{}", serde_json::to_string_pretty(states).unwrap());
+                    Ok(())
+                }
+                None => {
+                    if let Some(error) = result.get("error") {
```
tanmay4l: can we  push common RPC response handling into one helper ?

crates/core/src/rpc/admin.rs
```diff
@@ -192,6 +192,90 @@ pub trait AdminRpc {
...
+    /// ## Example Request (JSON-RPC)
+    /// ```json
+    /// {
+    ///   "jsonrpc": "2.0",
+    ///   "id": 107,
+    ///   "method": "setInstructionProfiling",
+    ///   "params": [true]
+    /// }
+    /// ```
+    #[rpc(meta, name = "setInstructionProfiling")]
```
tanmay4l, line 217: toggle is good can we get error codes/messages when meta is None too ?

crates/core/src/lib.rs
```diff
@@ -59,6 +59,15 @@ pub struct PluginInfo {
...
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct FeatureStates {
+    pub instruction_profiling_enabled: bool,
+    pub max_profiles: usize,
+    pub log_bytes_limit: Option<usize>,
+    pub loaded_plugins: Vec<PluginInfo>,
+}
```
tanmay4l: looks good , maybe we can make it Default

crates/cli/src/cli/mod.rs
```diff
@@ -633,6 +750,232 @@ async fn handle_cloud_commands(cmd: CloudCommand) -> Result<(), String> {
...
+                    } else {
+                        Err("Unexpected RPC response".to_string())
+                    }
+                }
+            }
+        }
+    }
+}
+
+fn rpc_call(
```
tanmay4l, line 846: not sure but maybe Don’t spawn curl Use a Rust HTTP client

crates/cli/src/cli/mod.rs
```diff
@@ -404,6 +410,115 @@ pub enum CloudCommand {
...
+
+#[derive(Parser, PartialEq, Clone, Debug)]
+pub struct FeatureProfilingCommand {
+    /// Enable or disable profiling (true/false)
+    #[arg(value_name = "ENABLED")]
+    pub enabled: String,
+    /// RPC URL of the running surfpool instance
+    #[arg(long = "rpc-url", default_value = "http://localhost:8899")]
+    pub rpc_url: String,
+}
```
tanmay4l, line 500: can we enabled a bool parsed by clap.Free form strings are error prone ?

lgalabru:
Thanks for the energy you've put into this @AvhiMaz ! 
I'm realizing the spec was quite light to get the ball rolling on this feature request.
This is what I initially had in mind: https://github.com/txtx/surfpool/pull/423

I don't have a strong opinion wether or not we should have a cheatcode to achieve this at runtime, I don't believe it's been requested at this point, but if you have a use case for this, could we be building off #423 instead?

Thanks, happy to breakdown specs further in the future if it's helpful 🙏

AvhiMaz:
> Thanks for the energy you've put into this @TanmayDhobale! I'm realizing the spec was quite light to get the ball rolling on this feature request. This is what I initially had in mind: #423
> 
> I don't have a strong opinion wether or not we should have a cheatcode to achieve this at runtime, I don't believe it's been requested at this point, but if you have a use case for this, could we be building off #423 instead?
> 
> Thanks, happy to breakdown specs further in the future if it's helpful 🙏

appreciate the clarity! i notice #423 is already merged - does that fully address what you had in mind, or are there gaps this issue should still cover?

if there's additional work needed, i'd love a bit more detailed spec, would love to contribute.

lgalabru:
We should be good for now, let's reconsider this feature if we see some demand.
Thanks again @AvhiMaz 🙏

Issue #416, opened 2025-11-21
Add Scenario Support for Switchboard


lgalabru:
**Switchboard**:

https://docs.rs/switchboard-on-demand/0.11.1/src/switchboard_on_demand/on_demand/oracle_quote/quote_account.rs.html#119-134

lgalabru:
Additional note: https://gist.github.com/mgild/532cd5aa207ea4809069e91ed901a181

tanmay4l:
@lgalabru I’ll take it up if no one’s working on it rn

lgalabru:
Addressed with #424, thanks @TanmayDhobale 🙏

PR #424, opened 2025-11-25
feat: Add Switchboard protocol support with `SwitchboardQuote` override template.

tanmay4l:
#416  cc @lgalabru

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12268,6 +12268,7 @@ dependencies = [
  "test-case",
  "txtx-addon-kit",
  "txtx-addon-network-svm-types",
+ "url 1.7.2",
  "uuid",
 ]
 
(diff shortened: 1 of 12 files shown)


lgalabru:
Oh no, I introduced some conflicts by merging another PR first :(
You mind updating your branch @TanmayDhobale? 

As is this PR won't work because Switchboard is using 2 bytes discriminators so I need to tweak something in the code, but we'll merge your PR and I'll follow-up with a patch right after.

tanmay4l:
> Oh no, I introduced some conflicts by merging another PR first :( You mind updating your branch @TanmayDhobale?
> 
> As is this PR won't work because Switchboard is using 2 bytes discriminators so I need to tweak something in the code, but we'll merge your PR and I'll follow-up with a patch right after.

yea btw currently have an 8-byte discriminator

PR #426, opened 2025-11-26
fix: make SURFPOOL_DATASOURCE_RPC_URL env var work and hide API keys in UI

serejke:
The env var was never being read because the --rpc-url CLI arg had a default_value, making the Option always Some. Changed to use clap's built-in `env` attribute which properly handles the env var fallback.

Also sanitized the displayed datasource URL to show only scheme and host (e.g., "https://example.com") to prevent leaking API keys in the UI.


lgalabru:
Can you please run `cargo +nightly fmt --all` ?

PR #427, opened 2025-11-26
feat(scenario): add drift on-chain perp dex V2 scenario

JkrishnaD:
- Added Drift v2 protocol
- Added cleaned + converted idl.json for Drift v2
- Added `overrides.yaml` with 4 override templates (PerpMarket, SpotMarket, User, Margin/Liquidation)
- Registered Drift protocol and templates in the scenario registry

This enables creating Drift perp, spot, and user-state scenarios directly in Surfpools.


jkrishnad:
@lgalabru check this implementation for the Drift protocol

PR #429, opened 2025-11-26
 Fix --watch flag not working on Linux.

### LINKED ISSUE #362: Ensure `--watch` flag is working on linux
See [this comment](https://github.com/txtx/surfpool/issues/295#issuecomment-3242673159) and the remaining thread - would be good to get some validation on if this feature is working on linux

daedlock:
### Two issues:

  1. Event filtering too strict for Linux - inotify reports `Data(Any)` instead of `Data(Content)`, and
   atomic file replacements (how cargo build works) emit `Name(To)/Name(Both)` events. Added these
  patterns to the match.
  2. Watch thread dying silently - Runbook execution was panicking, killing the watch thread.
  Wrapped in catch_unwind to keep the thread alive.

  Fixes #362

### DIFF
--- a/crates/cli/src/cli/simnet/mod.rs
+++ b/crates/cli/src/cli/simnet/mod.rs
@@ -552,6 +552,18 @@ async fn write_and_execute_iac(
                                 kind: EventKind::Create(CreateKind::File),
                                 paths,
                                 attrs: _,
+                            })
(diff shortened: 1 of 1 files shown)


MicaiahReid (approved):
Awesome, thanks @daedlock!!

PR #437, opened 2025-12-01
fix: surfnet_setAccount dropping zero-lamport accounts

rajgoesout:
Fixes https://github.com/txtx/surfpool/issues/433


MicaiahReid:
I'm not sure if the issue is actually relevant, and I'd like to avoid the unsafe rust

Issue #438, opened 2025-12-01
Add better error messaging when starting a surfnet on the same port as another network


MicaiahReid:
> Just hit this today, it happen when you have more than one surfpool running. So workaround is kill the old process. 

 _Originally posted by @katopz in [#331](https://github.com/txtx/surfpool/issues/331#issuecomment-3592622541)_

GauravBurande:
https://github.com/txtx/surfpool/blob/92e9df6a3fdf8545e18c53f4cfcf0f406d68be0b/crates/core/src/rpc/full.rs#L1615 is this where this error is being handled?

MicaiahReid:
No, this error should be returned on startup, not when a transaction fails to be processed.

If you try to start a surfnet and the port is already being used, we should return an error

dotsvm:
gm @GauravBurande are you working on this issue?

GauravBurande:
yeah, but u can go ahead, I'm not able to find source.

dotsvm:
> yeah, but u can go ahead, I'm not able to find source.

ok

PR #439, opened 2025-12-01
fix(simnet): improve error handling and port availability checks

Pawank06:
This PR introduces safer and more graceful error handling in the Simnet threading flow:

- Replace process exit with event-based error propagation in simnet thread
- Add port availability validation before starting RPC and WebSocket servers
- Extract port checking logic into reusable check_port_availability function
- Print error messages to stderr for better visibility when ports are in use
- Remove unnecessary sleep and direct process exit calls
- Provide clearer error messages when address binding fails
- Allow graceful error handling through the event channel instead of abrupt termination

Before:
Screenshot 2025-12-02 at 12 15 43 AM

After: 
Screenshot 2025-12-02 at 12 16 24 AM

fixes: #438

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4530,6 +4530,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "kaigan"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


dotsvm:
@MicaiahReid ser can you please check this PR? thank you!

PR #445, opened 2025-12-08
hotfix for date time travel 

Peacanduck:
added check to convert timestamp to ms and remved the default + 1hour

### DIFF
--- a/crates/core/src/rpc/surfnet_cheatcodes.rs
+++ b/crates/core/src/rpc/surfnet_cheatcodes.rs
@@ -1647,7 +1647,14 @@ impl SurfnetCheatcodes for SurfnetCheatcodesRpc {
         config: Option<TimeTravelConfig>,
     ) -> Result<EpochInfo> {
         let key = meta.as_ref().map(|ctx| ctx.id.clone()).unwrap_or_default();
-        let time_travel_config = config.unwrap_or_default();
+        let mut time_travel_config = config.unwrap_or_default();
+        // FIX: Convert timestamp from seconds to milliseconds if needed
+        if let TimeTravelConfig::AbsoluteTimestamp(ref mut ts) = time_travel_config {
+         // comment If timestamp is in seconds (< 10 billion), convert to milliseconds
+        if *ts < 10_000_000_000 {
+            *ts *= 1000;
+        }
+         }
         let simnet_command_tx = meta.get_surfnet_command_tx()?;
         let svm_locker = meta.get_svm_locker()?;
 
--- a/crates/core/src/types.rs
+++ b/crates/core/src/types.rs
@@ -789,8 +789,8 @@ pub enum TimeTravelConfig {
 
 impl Default for TimeTravelConfig {
     fn default() -> Self {
-        // chrono timestamp in ms, 1 hour from now
-        Self::AbsoluteTimestamp(Utc::now().timestamp_millis() as u64 + 3600000)
+        // chrono timestamp in ms, 1 hour from now + 3600000 removed
+        Self::AbsoluteTimestamp(Utc::now().timestamp_millis() as u64)
     }
 }
 


crates/core/src/types.rs
```diff
@@ -789,8 +789,8 @@ pub enum TimeTravelConfig {
 
 impl Default for TimeTravelConfig {
     fn default() -> Self {
-        // chrono timestamp in ms, 1 hour from now
-        Self::AbsoluteTimestamp(Utc::now().timestamp_millis() as u64 + 3600000)
+        // chrono timestamp in ms, 1 hour from now + 3600000 removed
+        Self::AbsoluteTimestamp(Utc::now().timestamp_millis() as u64)
```
MicaiahReid, line 793: Hmm, it doesn't really make sense to me for the default time to be `now` - why would the user time travel to now?

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -1647,7 +1647,14 @@ impl SurfnetCheatcodes for SurfnetCheatcodesRpc {
         config: Option<TimeTravelConfig>,
     ) -> Result<EpochInfo> {
         let key = meta.as_ref().map(|ctx| ctx.id.clone()).unwrap_or_default();
-        let time_travel_config = config.unwrap_or_default();
+        let mut time_travel_config = config.unwrap_or_default();
+        // FIX: Convert timestamp from seconds to milliseconds if needed
+        if let TimeTravelConfig::AbsoluteTimestamp(ref mut ts) = time_travel_config {
+         // comment If timestamp is in seconds (< 10 billion), convert to milliseconds
+        if *ts < 10_000_000_000 {
```
MicaiahReid, line 1654: I think I'd rather be clear in the docs + erroring that the timestamp should be in ms, rather than play this guessing game of should we/should we not convert. Thoughts @lgalabru?

crates/core/src/types.rs
Peacanduck: > Hmm, it doesn't really make sense to me for the default time to be `now` - why would the user time travel to now?

when time traveling it to say x time it ends up (x time + 1hour) so seems like default value is still being applied when entering your own value

crates/core/src/rpc/surfnet_cheatcodes.rs
Peacanduck: > I think I'd rather be clear in the docs + erroring that the timestamp should be in ms, rather than play this guessing game of should we/should we not convert. Thoughts @lgalabru?

Ideally fixing the source of the bug would be best I'm just not sure where to find the timeTravel ui i think that may be where it is

MicaiahReid:
I pushed a minor fix to the studio UI, but I wasn't able to reproduce the time travel bugs on the surfpool side. Closing this PR, feel free to reopen if you think this is a mistake.

Peacanduck:
checked using Surfpool version 1.1.1 looks like it's time traveling but reverting so functionally isn't working 

I did 30 days 3 times then select 2 months and I got this in the terminal note the 3 30 day time travels end up at the same day just out by a couple minutes also there was no change in the studio ui for the slot and epoch 

 18:35:34.905 Time travel to 2026-06-03 16:35:34 successful (epoch 980 / slot 423576591)                                           
⏐ 18:36:14.683 Time travel to 2026-06-03 16:36:14 successful (epoch 980 / slot 423576691)                                           
⏐ 18:36:38.850 Time travel to 2026-06-03 16:36:37 successful (epoch 980 / slot 423576749)                                           
⏐ 18:36:49.411 Time travel to 2026-07-02 16:36:48 successful (epoch 994 / slot 429624002)

only thing in chrome console was this warnings no errors like last time looks like time travel could be succeeding but immediately reverting or being overwritten back to the old state 

⏰ slot subscription timeout - no response received
⏰ transaction subscription timeout - no response received

PR #447, opened 2025-12-13
feat(core): add scenario scope to `surfnet_exportSnapshot` RPC

MicaiahReid:
- change registerScenario to proactively evaluate account overrides for scenario
 - add new scope to exportSnapshot rpc to allow exporting a scenario

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12399,6 +12399,7 @@ name = "surfpool-types"
 version = "1.0.0-rc1"
 dependencies = [
  "anchor-lang-idl",
+ "base64 0.22.1",
  "blake3",
  "chrono",
  "crossbeam-channel",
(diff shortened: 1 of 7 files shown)


lgalabru:
I'm confused, do we still need this PR @MicaiahReid ?

MicaiahReid:
> I'm confused, do we still need this PR @MicaiahReid ?

Hmm, good question.

This PR does a few potentially helpful things, if scenarios are to be more developed:
 - proactively compiles all account changes for a scenario rather than lazily
   - so, when you register the scenario, it fetches all accounts, and generates the new account states and stores that
   - when you step through slots it's just applying the already computed account state, rather than fetching accounts and computing on the fly
 - allows exporting those scenario account states as a snapshot, to be imported for tests

If we see scenarios being fleshed out further as a feature, this PR should probably be taken over the finish line. If not, we can probably close.

PR #453, opened 2025-12-19
fix(core): validate durable nonce blockhashes

MicaiahReid:


lgalabru (approved):
Really cool, thanks @MicaiahReid!

Issue #455, opened 2025-12-24
Feature Request: allow sigverify: false for sending a TX


Taylor123:
I have a testing environment where I'm simulating multiple transactions in a sequence. I'm unable to access the permissioned keypair that would be the sender on mainnet, so I'd like to be able to send the transactions from that address, but cannot do so without signature verification.

This does not seem possible based on the current documentation.

Simulate method is unsuitable due to the need for persistence between transactions.

Proposed solutions that fit the case would be to either:
- add option to send single tx with sig verification turned off
- have the entire environment run without sig verification

lgalabru:
thanks @Taylor123! looks like it's being addressed in #456 (thanks @tanmay4l!)
we're just missing some unit tests, you could already give it a spin if this a blocker on your end.

lgalabru:
@Taylor123 you should be all set https://github.com/txtx/surfpool/issues/455 got merged - thanks again @tanmay4l 🙏

PR #456, opened 2025-12-24
feat: Add `skip_signature_verification` option to `SimnetConfig` 

tanmay4l:
#455 

cc @lgalabru

### DIFF
--- a/crates/core/src/rpc/full.rs
+++ b/crates/core/src/rpc/full.rs
@@ -4460,4 +4460,176 @@ mod tests {
             })
         )
     }
+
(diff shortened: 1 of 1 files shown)


lgalabru (commented):
Looking sharp, thanks @tanmay4l!
The code looks correct to me, do you think you could add some unit tests?

tanmay4l:
> Looking sharp, thanks @tanmay4l! The code looks correct to me, do you think you could add some unit tests?

oh ive wrriten test alrdy but didnt push bcs thought not needed but pushing now

PR #460, opened 2026-01-04
feat: add accounts preloading with its for geyser plugins

### LINKED ISSUE #459: Accounts preloading feature
Hey! I'm using awesome geyser feature for plugin development with surfpool, and noticed that there's no way to preload accounts, `is_startup=false` in `update_account`: https://github.com/anza-xyz/agave/blob/v3.1/geyser-plugin-interface/src/geyser_plugin_interface.rs#L424

It would be nice, to leverage runbooks for example to specify accounts to have on startup.

Possible use cases, aside from pretty niche geyser example:
1) Have a consistent set of accounts at load time, i.e. several dexes with tick arrays and related pdas (single getMultipleAccounts call)
2) cli flag `--accounts <file>` for preloading prepared account dump for env bootstrapping (states, modified mints/owners whatever), similar to payer accounts initialization functionality, just more flexible
3) Probably it can evolve to snapshot functionality, i.e.:
- mirror mainnet
- modify state, persist
- next session start from snapshot

The core issue, at least in my case, that runbooks executed after geyser plugin initialization: github.com/txtx/surfpool/blob/main/crates/core/src/runloops/mod.rs#L124-L143
And i suppose, after whole runtime initialization.

Anantdadhich:
Closes #459  

This PR adds the Accounts Preloading feature requested in #459. It introduces a new CLI flag, --preload-accounts, which lets users load accounts from a JSON file instantly when Surfpool starts, before the Geyser plugins are initialized.

Right now, Geyser plugins in Surfpool can’t receive account updates with is_startup=true because:
Runbooks run after plugins are initialized
RPC calls only happen once the server has started

With his change fills that gap and makes it possible to:
Test Geyser plugin behavior that depends on startup state
Set up development environments with a predefined set of accounts, such as pre-minted tokens or program state

Just want clear reviews and correction to confirm whether this implementation is correct or not .

### DIFF
--- a/crates/cli/src/cli/mod.rs
+++ b/crates/cli/src/cli/mod.rs
@@ -245,9 +245,8 @@ pub struct StartSimnet {
     /// A set of inputs to use for the runbook (eg. surfpool start --runbook-input myInputs.json)
     #[arg(long = "runbook-input", short = 'i')]
     pub runbook_input: Vec<String>,
-   
-    #[arg(long = "preload-accounts")]
-    pub preload_accounts: Option<String>,
+    #[arg(long = "snapshot")]
+    pub snapshot: Vec<String>,
 }
 
 fn parse_svm_feature(s: &str) -> Result<SvmFeature, String> {
@@ -376,7 +375,9 @@ impl StartSimnet {
     }
 
     pub fn simnet_config(&self, airdrop_addresses: Vec<Pubkey>) -> SimnetConfig {
-        use surfpool_types::PreloadConfig;
+        use std::collections::HashMap;
+        use std::str::FromStr;
+        use surfpool_types::AccountSnapshot;
 
         let remote_rpc_url = if !self.offline {
             Some(self.datasource_rpc_url())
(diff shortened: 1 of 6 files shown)


lgalabru:
@ivs-ug would you have some bandwidth to test this PR?

lgalabru:
@Anantdadhich I added to the spec here https://github.com/txtx/surfpool/issues/459, wdyt?

Anantdadhich:
Hey thanks for the detailed spec and review my PR ! This makes a lot of sense. Reusing the existing snapshot format keeps things consistent between export and import, and the null approach for accounts fetched over the network feels clean and straightforward.

I’ll handle this by adding a --snapshot flag , keeping the JSON aligned with the current snapshot export format, and wiring it up to SimnetConfig.snapshots: Vec<IndexMap<Pubkey, Option<AccountSnapshot>>>.

When a value is null, I’ll just fetch the account from the remote RPC.

ivs-ug:
> @ivs-ug would you have some bandwidth to test this PR?

Sure, ready to test! Could you guys give brief usage guide? I briefly checked pr, it seems snapshot json should be `<HashMap<String, Option<AccountSnapshot>>>` where `AccountSnapshot` is your internal `Account` extension, it's due to you already have some snapshotting functionality?

And Option<> means if no snapshot, given pubkey will be automatically fetched?

My concern is, it make it pretty limited in usage, typically, accounts serialized to `Vec<(Pubkey, Account)>` since HashMap can't have Pubkey key for serde serialization. So what I actually thought of, either list of keys(cli flag with just list or json with Vec<Pubkey>) and surfpool fetches them on load automatically, same as typical account fetching during normal operation.
Or just `Vec<(Pubkey, Account)>` json dump, also recent solana crates has start migrating to `UiAccount` but it doesn't look usable actually(get_multiple_accounts marked deprecated in favor of Ui* version).

In such case it's pretty universal interface, dump from anywhere - load in surfpool with no friction. Wdyt?

P.S.: Guys, your feature request resolution performance is just unbelievable! Thx for putting an attention for my issue!

lgalabru:
@Anantdadhich we did realized a few hours ago that this import snapshot feature was on our internal critical path for a larger effort, so @MicaiahReid ended up expediting a solution, and we ended up folding #459 in it.
Thanks again for your help 🙏

lgalabru:
Closing in favor of https://github.com/txtx/surfpool/pull/464

Anantdadhich:
Thanks for the clarification and for the kind words 🙏

Totally understand the need to fold this into a larger internal effort. I’m glad the discussion around #459 was useful and helped shape the final solution.

Happy to contribute again in the future!

PR #464, opened 2026-01-06
feat(core): add support for loading account snapshots at startup

MicaiahReid:


crates/cli/src/cli/simnet/mod.rs
```diff
@@ -82,8 +82,40 @@ pub async fn handle_start_local_surfnet_command(
         Some(keypair)
     };
 
+    // Parse and merge snapshot files (multiple files supported, later files override earlier ones)
+    // The actual loading happens in the runloop after the locker is created
+    let snapshot = {
+        let mut merged_snapshot: std::collections::BTreeMap<
```
lgalabru, line 88: could we use an index map instead?

PR #466, opened 2026-01-08
feat: snapshot events through ws

lgalabru:


crates/core/src/rpc/ws.rs
```diff
@@ -1397,4 +1523,147 @@ impl Rpc for SurfpoolWsRpc {
...
+            } else {
+                log::error!("Failed to acquire write lock on snapshot_subscription_map");
+                return;
+            }
+
+            // Generate a unique snapshot ID for this import operation
+            let snapshot_id = format!("snapshot_{}", chrono::Utc::now().timestamp_nanos_opt().unwrap_or(0));
+
+            // Send initial notification that snapshot import has started
+            let start_notification = crate::surfnet::SnapshotImportNotification {
```
MicaiahReid: Hmm, it seems fishy that we're sending this hear as well as in the actual `subscribe_for_snapshot_import_updates` function - it should just be in the one in the svm file, right?

crates/core/src/surfnet/svm.rs
```diff
@@ -2288,6 +2290,86 @@ impl SurfnetSvm {
...
+                snapshot_id: snapshot_id.clone(),
+                status: super::SnapshotImportStatus::InProgress,
+                accounts_loaded: 0,
+                total_accounts,
+                error: None,
+            });
+
+            // TODO: The actual loading should be done via the load_snapshot method
+            // For now, we simulate the process and send completion
+            tokio::time::sleep(tokio::time::Duration::from_millis(1000)).await;
```
MicaiahReid: I think the `load_snapshot` fn should now exist! And I believe it's synchronous so that's nice.

crates/core/src/surfnet/svm.rs
```diff
@@ -2288,6 +2290,86 @@ impl SurfnetSvm {
...
+            });
+        });
+        
+        rx
+    }
+
+    async fn fetch_snapshot_from_url(
+        snapshot_url: &str,
+    ) -> Result<std::collections::BTreeMap<String, Option<surfpool_types::AccountSnapshot>>, Box<dyn std::error::Error + Send + Sync>> {
+        // For now, we'll use reqwest to fetch the snapshot data
```
MicaiahReid: In the future what will this be?

crates/core/src/rpc/ws.rs
```diff
@@ -659,6 +659,130 @@ pub trait Rpc {
...
+    /// - Each subscription runs in its own async task for optimal performance.
+    ///
+    /// ## See Also
+    /// - `snapshotUnsubscribe`: Remove an active snapshot subscription
+    #[pubsub(
+        subscription = "snapshotNotification",
+        subscribe,
+        name = "snapshotSubscribe"
+    )]
+    fn snapshot_subscribe(
```
MicaiahReid, line 729: This seems like a weird name, cause it's not just subscribing to notifications about a snapshot being imported, it's actually initiating the import.

Also, why are we having surfpool fetch snapshots instead of just receiving them via some RPC method?

MicaiahReid (commented):
I think a little more context would help me understand some of the whys! Plus one or two hygiene comments

PR #468, opened 2026-01-09
fix: execute runbooks even if project framework isn't detected

MicaiahReid:


crates/cli/src/cli/simnet/mod.rs
```diff
@@ -480,76 +441,82 @@ async fn write_and_execute_iac(
...
+
+    let base_location =
+        FileLocation::from_path_string(&cmd.manifest_path)?.get_parent_location()?;
+    let mut txtx_manifest_location = base_location.clone();
+    txtx_manifest_location.append_path("txtx.yml")?;
+    let txtx_manifest_exists = txtx_manifest_location.exists();
+
+    let mut on_disk_runbook_data = None;
+    let mut in_memory_runbook_data = None;
+    let runbook_input = cmd.runbook_input.clone();
```
BretasArthur1, line 454: I see that you're cloning this inputs sometimes along this file, shouldn't be better to change the `assemble_runbook_execution_futures` signature to hold a `&Arc<Vec<String>>` instead and avoid some of this deep clones with cheap clones or no clones at all?

I think this could save some memory and since the inputs aren't changed on the process I don't see a case where the shared reference can be end up breaking things.

Also following the data, I found that the clone in the line 538 looks redundant - it's already moved into the closure.

crates/cli/src/cli/simnet/mod.rs
BretasArthur1: here

crates/cli/src/cli/simnet/mod.rs
MicaiahReid: In most cases, `runbook_input` is a vec of very small strings. It's rarely, if ever large data being shared. For example, if you run `surfpool run <runbook> --runbook-input network_id=localnet`, the `network_id=localnet` is the runbook input. Using Arc adds some overhead, without saving much in terms of memory, so I don't think it's worth it. 

> Also following the data, I found that the clone in the line 538 looks redundant - it's already moved into the closure.

Great find! Will remove!

BretasArthur1: gotcha, thx!

PR #476, opened 2026-01-14
fix(core): rpc simulateTransaction loadedAccountsDataSize is None (#475)

pymongo:
#475

### DIFF
--- a/crates/core/src/tests/integration.rs
+++ b/crates/core/src/tests/integration.rs
@@ -609,6 +609,15 @@ async fn test_simulate_add_alt_entries_fetching(test_type: TestType) {
         simulation_res.value.err, None,
         "Unexpected simulation error"
     );
+    assert!(
+        simulation_res.value.loaded_accounts_data_size.is_some(),
+        "Expected loaded_accounts_data_size to be present"
+    );
+    assert_eq!(
+        simulation_res.value.loaded_accounts_data_size.unwrap(),
+        140134,
+        "Incorrect loaded_accounts_data_size value"
+    );
     let simulation_res2 = full_client
         .simulate_transaction(
             data,
@@ -627,6 +636,14 @@ async fn test_simulate_add_alt_entries_fetching(test_type: TestType) {
     assert_eq!(
         simulation_res2.value.err, None,
         "Unexpected simulation error"
+    );
+    assert!(
+        simulation_res2.value.loaded_accounts_data_size.is_some(),
+        "Expected loaded_accounts_data_size to be present"
+    );
+    assert!(
+        simulation_res2.value.loaded_accounts_data_size.unwrap() > 0,
+        "Expected loaded_accounts_data_size to be greater than 0"
     );
 }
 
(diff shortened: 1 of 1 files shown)


crates/core/src/rpc/full.rs
```diff
@@ -1687,6 +1688,46 @@ impl Full for SurfpoolFullRpc {
...
+                        }
+                        if let Some(td) = td_account {
+                            if seen_accounts.insert(*td_pubkey) {
+                                loaded_accounts_data_size += td.data.len() as u64;
+                            }
+                        }
+                    }
+                    GetAccountResult::None(_) => {}
+                }
+            }
```
MicaiahReid: Since this logic is used twice (once for regular accounts and once for ALT accounts), can we pull this logic into a reusable closure?

crates/core/src/rpc/full.rs
```diff
@@ -1687,6 +1688,46 @@ impl Full for SurfpoolFullRpc {
...
+                    ) => {
+                        if seen_accounts.insert(*pubkey) {
+                            loaded_accounts_data_size += account.data.len() as u64;
+                        }
+                        if let Some(td) = td_account {
+                            if seen_accounts.insert(*td_pubkey) {
+                                loaded_accounts_data_size += td.data.len() as u64;
+                            }
+                        }
+                    }
```
MicaiahReid, line 1730: This bit is a bit interesting. Our `GetAccountResult` is what is returned when a transaction is fetched from the remote. If the queried account is a program account we also fetch the program data account, to make sure they are loaded together. Likewise, if the queried account is a token account, we also fetch the token mint alongside it.

That being said, if some program account `P` is included in a transaction, and thus it is fetched, its program data account, `PD`, is not necessarily one of the transaction's accounts, and shouldn't be included in the data size. Similarly, if a token account is included in the transaction, the mint isn't necessarily there as well.

So for each of these accounts, you'll need some mechanism to check if the account is listed in the actual transaction or not.

MicaiahReid (changes requested):
Thanks for the PR, @pymongo! I have a few changes needed to get this over the finish line.

MicaiahReid:
@pymongo do you think you'll be able to get these fixes in?

pymongo:
I am researching, fogo svm don't count program size as data size, but solana count program size as data size, how surfpool to compatiable both fogo and solana

MicaiahReid:
> I am researching, fogo svm don't count program size as data size, but solana count program size as data size, how surfpool to compatiable both fogo and solana

Nice good find. According to [this SIMD](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0186-loaded-transaction-data-size-specification.md), the program data account is included. However, the mint account is not. Are you able to get that change in, or do you need me to take over? We'd like to get this into our upcoming 1.0 release!

pymongo:
Sorry but I am busy rencently, no time to make the additional changes

You can resign/assigned to this issue/problem, thanks

MicaiahReid (approved):

PR #479, opened 2026-01-14
chore: update litesvm

lgalabru:


Aursen:
Is this PR affected by https://github.com/LiteSVM/litesvm/issues/279?

lgalabru:
@Aursen I don't think so - this PR is building just fine, but we have an issue with `solana-geyser-plugin-manager` (disabled in this PR)

PR #480, opened 2026-01-15
feat(core): add Jito sendBundle RPC support

serejke:
## Summary
- Add non-standard `sendBundle` RPC endpoint for Jito bundles imitation
- Allows clients to submit multiple transactions that are executed sequentially
- Useful for testing applications that use Jito bundles

## Implementation details
- Transactions are processed sequentially in the order provided
- Forces `skip_preflight: true` to match Jito Block Engine behavior
- Returns bundle ID as SHA-256 hash of comma-separated signatures (Jito-compatible)
- If any transaction fails, the entire bundle is rejected with error indicating which transaction failed

## Notes
This is not a 100% bundles implementation as they are not atomic all-or-nothing like real Jito bundles. But it's useful for testing apps that want to verify their Jito integration works.

## Test plan
- [x] Added unit tests for empty bundle rejection
- [x] Added unit tests for missing context error
- [x] Added async tests for single transaction bundle
- [x] Added async tests for multiple transaction bundle with bundle ID verification
- [x] All tests passing: `cargo test -p surfpool-core jito`

### DIFF
--- a/crates/core/src/rpc/jito.rs
+++ b/crates/core/src/rpc/jito.rs
@@ -2,6 +2,7 @@ use std::str::FromStr;
 
 use jsonrpc_core::{Error, Result};
 use jsonrpc_derive::rpc;
+use sha2::{Digest, Sha256};
 use solana_client::{rpc_config::RpcSendTransactionConfig, rpc_custom_error::RpcCustomError};
 use solana_signature::Signature;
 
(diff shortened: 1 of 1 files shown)


lgalabru:
@serejke I'm doing a bit of triage and noticing that we had a similar PR in the past - https://github.com/txtx/surfpool/pull/480. 

I'd be supportive of getting of these 2 merged, would love to get your 2 cents!

bonlime:
@lgalabru  you probably wanted to tag this PR #412 
imo this feature seems use-full for testing things that are stateful and can't fit into a single transaction

PR #484, opened 2026-01-16
fix(db): only store blocks with txs; reconstruct sysvars from db on restart

MicaiahReid:


crates/core/src/rpc/full.rs
```diff
@@ -1833,12 +1830,19 @@ impl Full for SurfpoolFullRpc {
...
-                    svm_reader
-                        .blocks
-                        .get(&slot)?
-                        .map(|block| (block.block_time / 1_000) as UnixTimestamp),
-                )
+                Ok::<_, jsonrpc_core::Error>(match svm_reader.blocks.get(&slot)? {
+                    Some(block) => Some(block.block_time),
+                    None => {
+                        // With sparse block storage, calculate time for missing blocks
+                        let latest_slot = svm_reader.get_latest_absolute_slot();
```
lgalabru: Should we move that piece of logic in a code site where it can easily be unit tested?

crates/core/src/surfnet/surfnet_lite_svm.rs
```diff
@@ -91,18 +95,33 @@ impl SurfnetLiteSvm {
     /// Perform garbage collection by resetting the SVM state while retaining the database.
     /// This is useful for cleaning up unused accounts and reducing memory usage.
     /// If no database is configured, this function is a no-op.
+    #[allow(deprecated)]
```
lgalabru, line 98: `deprecated` in favor of what?

crates/core/src/surfnet/svm.rs
```diff
@@ -892,6 +959,61 @@ impl SurfnetSvm {
         })
     }
 
+    /// Reconstructs RecentBlockhashes, SlotHashes, and Clock sysvars deterministically
+    /// from the current slot. Called on startup and after garbage collection to ensure
+    /// consistent sysvar state without requiring database persistence.
+    ///
+    /// Note: SyntheticBlockhash uses chain_tip.index (relative index), while SlotHashes
+    /// and Clock use absolute slots (chain_tip.index + genesis_slot).
+    #[allow(deprecated)]
```
lgalabru, line 1022: `deprecated` in favor of what?

lgalabru: Could we get some test coverage for this method?

crates/core/src/surfnet/svm.rs
```diff
@@ -1778,17 +1900,32 @@ impl SurfnetSvm {
...
+
+        // Checkpoint the latest slot periodically (~every 1 minute)
+        // This allows recovery after restart without storing every empty block
+        const CHECKPOINT_INTERVAL_MS: u64 = 60_000; // 1 minute
+        let current_time_ms = Utc::now().timestamp_millis() as u64;
+        if current_time_ms.saturating_sub(self.last_slot_checkpoint_ms) >= CHECKPOINT_INTERVAL_MS {
+            self.slot_checkpoint
+                .store("latest_slot".to_string(), slot)?;
+            self.last_slot_checkpoint_ms = current_time_ms;
+        }
```
lgalabru, line 1979: Should we go with a slot based approach here too?

crates/core/src/surfnet/svm.rs
```diff
@@ -107,6 +107,10 @@ use crate::{
     },
 };
 
+/// Interval (in slots) at which to perform garbage collection on the lite SVM cache
+/// About 1 hour at standard 400ms slot time
+const GARBAGE_COLLECTION_INTERVAL_SLOTS: u64 = 9_000;
```
lgalabru: Could we get a lazy static reading from `env.SURFPOOL_GARBAGE_...`  and default to 9000?
I could see us wanting to tweak that variable on the go, based on cluster's load, etc.

crates/core/src/surfnet/surfnet_lite_svm.rs
MicaiahReid: I don't know :/

The imports from `solana_sysvar` don't give an alternative, they just say: `Please do not use, will no longer be available in the future`. But currently, this is what is used by the VM for getting RecentBlockhashes

crates/core/src/rpc/full.rs
MicaiahReid: Addressed 0e4c8e4d

crates/core/src/surfnet/svm.rs
MicaiahReid: Fixed: fc089dbe

crates/core/src/surfnet/svm.rs
MicaiahReid: Fixed: 3ad23511

Issue #462, opened 2026-01-06
Update litesvm


lgalabru:

subhdotsol:
can you properly have the description for the issue ? @lgalabru

lgalabru:
hey @subhdotsol! LiteSVM 0.9 was released 2 days ago, the goal here is to pull that latest version.
Beside the version bump, there could be some new feature gates that needs to be listed here:

https://github.com/txtx/surfpool/blob/main/crates/types/src/features.rs

Some features should now be enabled by defaults, some should be added, etc.

subhdotsol:
can i work on this issue ? @lgalabru

lgalabru:
of course! will assign you to avoid collisions!

subhdotsol:
@lgalabru you can close the issue now i raised the PR for this and it got merged

MicaiahReid:
@subhdotsol your PR (https://github.com/txtx/surfpool/pull/471) does not update litesvm, it only does some prep work for it. Do you still want to do the update?

serejke:
I tried to upgrade the litesvm to 0.9.0 in the surfpool repo, it contains my merged PR on the priority fees https://github.com/LiteSVM/litesvm/pull/264

But Solana crates are currently published incorrectly
https://github.com/anza-xyz/agave/issues/9927#issuecomment-3737816633

Since the surfpool depends on the `solana-geyser-plugin-manager` it doesn't seem to be possible to resolve these conflicts until Agave publishes 3.1.7

Issue #186, opened 2025-06-24
Remove homebrew in favor of surfpool install script


lgalabru:

MicaiahReid:
Alternative: remove homebrew from our install process and instead use our installer (`curl -sL https://run.surfpool.run/ | bash`)

PR #487, opened 2026-01-16
fix: correct airdrop tx pre/post balances using LiteSVM data

### LINKED ISSUE #229: Update airdrop transaction meta to include correct pre/post balances
The airdrop transaction is created by litesvm, and references a keypair that we don't have access to. If this changes in the future, we should update our transaction meta to query the airdrop account's before/after balance

ANISH-SR:
### Summary of Changes
This change fixes incorrect airdrop transaction metadata by leveraging LiteSVM’s newly exposed airdrop account pubkey.
Previously, Surfpool synthesized a hardcoded airdrop keypair &  balance deltas when constructing airdrop transaction meta, because the real airdrop account used by LiteSVM was not accessible. This resulted in inaccurate pre + post Balances.

With LiteSVM now exposing the airdrop pubkey, Surfpool:
- Exposes that pubkey through SurfnetLiteSvm.
- Captures real account balances before and after the airdrop for:
    - the LiteSVM airdrop account
    - the recipient account
    - the system program account.
- Uses these real balances to populate TransactionStatusMeta.pre_balances and post_balances.

tl:dr,
Use svm.airdrop_pubkey() to fetch the airdrop account balance before and after the transaction, and use those values to populate preBalances and postBalances in the airdrop transaction metadata.

Resolves #229


sranish:
Hi @MicaiahReid , can you please review this and lmk if this approach is correct..
Also, I am encountering build failures due to the workspace was still using LiteSVM 0.8.1, which does not expose the required function i.e airdrop_pubkey(). 

This PR depends on the LiteSVM change from [#186](https://github.com/LiteSVM/litesvm/issues/186) / [#238](https://github.com/LiteSVM/litesvm/pull/238) that added this function so that the method was unresolved until LiteSVM was bumped to a compatible version.

Can you please guide me here on how should I proceed with the change. thanks : )

MicaiahReid:
> Hi @MicaiahReid , can you please review this and lmk if this approach is correct..
> 
> Also, I am encountering build failures due to the workspace was still using LiteSVM 0.8.1, which does not expose the required function i.e airdrop_pubkey(). 
> 
> 
> 
> This PR depends on the LiteSVM change from [#186](https://github.com/LiteSVM/litesvm/issues/186) / [#238](https://github.com/LiteSVM/litesvm/pull/238) that added this function so that the method was unresolved until LiteSVM was bumped to a compatible version.
> 
> 
> 
> Can you please guide me here on how should I proceed with the change. thanks : )

There's a PR open(#479) to update litesvm. Maybe building this PR off that branch would do the trick?

sranish:
I’ve rebased this PR on top of #479 so it picks up the LiteSVM upgrade there, and I’ve kept this PR scoped to just the airdrop transaction metadata fix.

The branch now builds cleanly on top of that change. Would appreciate a review when you have a chance @MicaiahReid 
: )

MicaiahReid:
@ANISH-SR, I've pushed a fix to this - you were constructing the tx with an invalid signer (based on the pubkeys in the tx), so the airdrop failed.

But since we have the signature and we're just reconstructing the tx, there's no need to have it signed in the first place.

We'll merge this after #462 

---

As a side note, it's a great practice to build and run your branches as a smoke test whenever you open a PR! This bug was easy to catch - it crashed surfpool right on startup!

Thank you for the PR :)

PR #488, opened 2026-01-16
feat: polish tui design

lgalabru:


MicaiahReid:
Love the separate Logs/Transactions tab, but idk if I'm sold on the long links for the tx list. The link is very long (a full tx hash), so it's easy for it to be cut off in the TUI. Then, when you click the link, it's not to the full tx sig, so it's not taking you anywhere:
image


Also, as of now the tx links don't work in the studio - should that already be working in the studio?

MicaiahReid:
Also, since we have a separate txs list, should we not log this in logs? To keep the logs cleaner?
image

crates/cli/src/tui/simnet.rs
```diff
@@ -234,8 +231,9 @@ impl ColorTheme {
     fn new_safe_colors(_color: &tailwind::Palette) -> Self {
         Self {
             background: Color::Reset,
-            accent: Color::Green,
-            primary: Color::Green,
+            content_background: Color::Black,
+            accent: Color::Cyan,
+            primary: Color::Cyan,
             secondary: Color::White,
```
MicaiahReid: Currently primary/secondary are never used, introducing new warnings.

PR #495, opened 2026-01-21
fix(core): surfnet encode_ui_account to handle ReadableAccount inputs and enable agave unstable api

BretasArthur1:
# Problem
- `get_program_accounts_with_config` was deprecated in favor of `get_program_ui_accounts_with_config` in newer `solana_rpc_client` release
- `solana_runtime::commitment::BlockCommitmentArray` has been marked for formal inclusion in the agave unstable api (causing warns about it)

# Solution

- Make `encode_ui_account` generic over `ReadableAccount` with a trait-based access to raw bytes for IDL parsing, require `Sync` for locker usage, and pass through remote UiAccounts without re-encoding. 
- Enable agave unstable api feature set in the `solana-runtime` crate

### DIFF
--- a/crates/core/src/storage/sqlite.rs
+++ b/crates/core/src/storage/sqlite.rs
@@ -1,6 +1,9 @@
 use std::{
-    collections::HashSet,
-    sync::{Mutex, OnceLock},
+    collections::{HashMap, HashSet},
+    sync::{
+        Mutex, OnceLock,
+        atomic::{AtomicU64, Ordering},
+    },
 };
 
 use log::debug;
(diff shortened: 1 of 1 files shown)


MicaiahReid:
Thanks @BretasArthur1! Sorry for the delayed review. I'm getting the update to `get_program_ui_accounts_with_config` and the addition of the unstable agave API feature.

However, I don't understand the change to the `encode_ui_account` in the locker file or svm file. If I revert the changes in the locker, I see no compile errors or downside. And I want to understand if the changes in the SVM file are actually fixing anything or just a refactor or what?

crates/core/src/surfnet/locker.rs
```diff
@@ -2928,7 +2925,7 @@ impl SurfnetSvmLocker {
             .iter()
             .map(|(pubkey, account)| RpcKeyedAccount {
                 pubkey: pubkey.to_string(),
-                account: self.encode_ui_account(pubkey, account, encoding, None, data_slice),
+                account: account.clone(),
```
MicaiahReid: Very minor but easy optimization: `into_iter` to take ownership or remote accounts, then remove `account.clone()` (just use `account`)

BretasArthur1: nice catch!

BretasArthur1:
Wassup brother! np, I saw you guys rocking other stuff so I presumed u guys were busy :)

> However, I don't understand the change to the `encode_ui_account` in the locker file or svm file. If I revert the changes in the locker, I see no compile errors or downside. And I want to understand if the changes in the SVM file are actually fixing anything or just a refactor or what?

About this, so when I  changed `get_program_accounts_with_config` -> `get_program_ui_accounts_with_config` I had to change the return parameter of the `get_program_accounts` from `Account` to `UiAccount` this cascade until I had to change the parameter from `encode_ui_account` to `UiAccount` and this cause and error on the svm `encode_ui_account` where: 

```
the trait bound `solana_account_decoder::UiAccount: ReadableAccount` is not satisfied
the following other types implement trait `ReadableAccount`:
  AccountSharedData
  TransactionAccountView<'_>
  TransactionAccountViewMut<'_>
  solana_account::Account
  solana_accounts_db::account_storage::stored_account_info::StoredAccountInfo<'_>
  solana_accounts_db::accounts_db::LoadedAccount<'_>
  solana_accounts_db::accounts_update_notifier_interface::AccountForGeyser<'_>
  solana_accounts_db::append_vec::meta::StoredAccountMeta<'append_vec>
and 4 others
```

that's coming from the generic bound from the svm `encode_ui_account` so then I applied the fix on the locker to use a generic as well...

But if you found something different let me know, I can change

PR #506, opened 2026-01-26
feature: Replay onchain transactions

ExoMonk:
Issue: [Replay On-Chain transactions](https://github.com/txtx/surfpool/issues/505)

## Summary
Add the ability to replay On-Chain transactions locally for debugging and analysis. This feature enables developers to:
- Reproduce and debug production transactions in an isolated environment
- Analyze transaction execution with full profiling data
- Test protocol behavior against real historical transactions

## Changes

| File | Description |
|------|-------------|
| `crates/types/src/types.rs` | Add `ReplayConfig` and `ReplayResult` types |
| `crates/core/src/error.rs` | Add replay-specific error variants |
| `crates/core/src/surfnet/locker.rs` | Add `replay_transaction` method with blockhash replacement |
| `crates/core/src/rpc/surfnet_cheatcodes.rs` | Add `surfnet_replayTransaction` RPC method |
| `crates/cli/src/cli/mod.rs` | Add `ReplayCommand` struct and CLI wiring |
| `crates/cli/src/cli/replay/mod.rs` | New standalone CLI handler |
| `crates/cli/Cargo.toml` | Add `solana-signature` dependency |

## Usage

**CLI (standalone):**
```bash
surfpool replay <TX_SIGNATURE> -n mainnet --profile
surfpool replay --from-file sigs.txt -n mainnet -o results.json
TX_SIGNATURES=sig1,sig2 surfpool replay -n mainnet
```

**RPC method (with running surfpool):**
```bash
curl -X POST http://127.0.0.1:8899 -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","id":1,"method":"surfnet_replayTransaction","params":["<SIG>",{"profile":true}]}'
```

Test plan
[x] Replay single mainnet transaction via CLI
[x] Replay transaction via RPC method with surfpool start -n mainnet
[x] Verify profiling data includes account state diffs
[x] Verify compute units and logs match expected values

### DIFF
--- a/crates/cli/src/cli/mod.rs
+++ b/crates/cli/src/cli/mod.rs
@@ -25,6 +25,7 @@ use txtx_gql::kit::{helpers::fs::FileLocation, types::frontend::LogLevel};
 
 use crate::{cloud::CloudStartCommand, runbook::handle_execute_runbook_command};
 
+mod replay;
 mod simnet;
 
 #[derive(Clone)]
@@ -127,6 +128,9 @@ enum Command {
     /// Start MCP server
     #[clap(name = "mcp", bin_name = "mcp")]
     Mcp,
+    /// Replay mainnet transactions locally
+    #[clap(name = "replay", bin_name = "replay")]
+    Replay(ReplayCommand),
 }
 
 #[derive(Parser, PartialEq, Clone, Debug)]
@@ -271,6 +275,58 @@ pub enum NetworkType {
     Devnet,
     /// Solana Testnet (https://api.testnet.solana.com)
     Testnet,
+}
+
+/// Replay mainnet transactions locally for debugging and analysis.
+#[derive(Parser, PartialEq, Clone, Debug)]
+pub struct ReplayCommand {
+    /// Transaction signature(s) to replay
+    #[arg()]
+    pub signatures: Vec<String>,
+
+    /// Read transaction signatures from a file (one per line)
+    #[arg(long = "from-file", short = 'f')]
+    pub from_file: Option<String>,
+
+    /// RPC URL for fetching transaction data (eg. --rpc-url https://api.mainnet-beta.solana.com)
+    #[arg(
+        long = "rpc-url",
+        short = 'u',
+        conflicts_with = "network",
+        env = "SURFPOOL_DATASOURCE_RPC_URL"
+    )]
+    pub rpc_url: Option<String>,
+
+    /// Predefined network (mainnet, devnet, testnet)
+    #[arg(long = "network", short = 'n', value_enum, conflicts_with = "rpc_url")]
+    pub network: Option<NetworkType>,
+
+    /// Enable transaction profiling for detailed execution metrics
+    #[clap(long = "profile", action = ArgAction::SetTrue, default_value = "false")]
+    pub profile: bool,
+
+    /// Output results to a JSON file
+    #[arg(long = "output", short = 'o')]
+    pub output: Option<String>,
+
+    /// Skip time travel (use current slot instead of transaction's original slot)
+    #[clap(long = "skip-time-travel", action = ArgAction::SetTrue, default_value = "false")]
+    pub skip_time_travel: bool,
+}
+
+impl ReplayCommand {
+    /// Returns the RPC URL to use for fetching transaction data.
+    pub fn datasource_rpc_url(&self) -> String {
+        match self.network {
+            Some(NetworkType::Mainnet) => DEFAULT_MAINNET_RPC_URL.to_string(),
+            Some(NetworkType::Devnet) => DEFAULT_DEVNET_RPC_URL.to_string(),
+            Some(NetworkType::Testnet) => DEFAULT_TESTNET_RPC_URL.to_string(),
+            None => self
+                .rpc_url
+                .clone()
+                .unwrap_or_else(|| DEFAULT_MAINNET_RPC_URL.to_string()),
+        }
+    }
 }
 
 impl StartSimnet {
--- a/crates/cli/src/cli/mod.rs
+++ b/crates/cli/src/cli/mod.rs
@@ -702,6 +702,9 @@ fn handle_command(opts: Opts, ctx: &Context) -> Result<(), String> {
         Command::List(cmd) => hiro_system_kit::nestable_block_on(handle_list_command(cmd, ctx)),
         Command::Cloud(cmd) => hiro_system_kit::nestable_block_on(handle_cloud_commands(cmd)),
         Command::Mcp => hiro_system_kit::nestable_block_on(handle_mcp_command(ctx)),
+        Command::Replay(cmd) => {
+            hiro_system_kit::nestable_block_on(replay::handle_replay_command(cmd, ctx))
+        }
     }
 }
 
--- a/crates/cli/src/cli/replay/mod.rs
+++ b/crates/cli/src/cli/replay/mod.rs
@@ -1,4 +1,4 @@
-use std::{fs, str::FromStr};
+use std::{env, fs, str::FromStr};
 
 use log::info;
 use solana_signature::Signature;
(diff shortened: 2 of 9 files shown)


procdump:
This feature with the help of (https://github.com/txtx/surfpool/pull/496, https://github.com/anza-xyz/agave/pull/10155) can provide even more insight of the dynamics in place.

crates/cli/src/cli/replay/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+use super::{Context, ReplayCommand};
+
+/// Handles the replay command by fetching and re-executing transactions from mainnet.
+///
+/// This is a lightweight standalone handler that initializes an ephemeral SVM
+/// without a full RPC server. For interactive use with full surfpool features,
+/// use `surfpool start` with the `surfnet_replayTransaction` RPC method instead.
+pub async fn handle_replay_command(cmd: ReplayCommand, _ctx: &Context) -> Result<(), String> {
+    // Step 1: Parse signatures from args or file
+    let signatures = if let Some(file_path) = &cmd.from_file {
```
procdump, line 20: I'd like to propose here also the ability to encode the TX signatures in an environment variable as an option to get them from - something like:

`TX_SIGNATURES=signature1;signature2;etc surfpool replay -n mainnet --profile`

The motivation is that with such an unified approach they can reach all the way down the stack directly to `LiteSVM`'s before callback (https://github.com/LiteSVM/litesvm/blob/a9ab814e83583ac36a040782bfbf92e1dc4cbcfa/crates/litesvm/src/lib.rs#L1083) where it would be possible to actually invoke the debugger only for these specific signatures. This can be achieved easily with (https://github.com/anza-xyz/agave/pull/10155) provided it's approved.

ExoMonk: Great adding, that makes totally sense and I agree a lot of people are using that LiteSVM
I'm adding this right away !

ExoMonk: Added !

crates/core/src/surfnet/locker.rs
```diff
@@ -1084,6 +1084,202 @@ impl SurfnetSvmLocker {
...
+                    .map(|l| l.all_loaded_addresses()),
+            )
+            .clone();
+
+        // Step 6: Fetch account states (current state - historical not available)
+        let state_warning = Some(
+            "Account states are fetched at current slot, not at original transaction slot. \
+             Results may differ from original execution due to state changes."
+                .to_string(),
+        );
```
failfmi, line 1171: hey, nice feature 🚀 

Should this be called `replay`, when you do not actually execute it against the correct state, e.g. it might result in a different outcome?

ExoMonk: Thanks !
Valid concern indeed ! True historical replay would require an archive node or historical state service (which Solana mainnet RPC doesn't provide)

The current implementation fetches current account states and warns here so the value is in re-executing the exact transaction structure with profiling - useful for debugging/understanding transactions even if balances may have changed

Open to renaming if you have a preference (reexecute, reexec)?
I thought that `replay` was pretty cool naming based of the surpool context

ExoMonk: What replay actually do provide:

- Time-travel to the original slot (for Clock sysvar, epoch info)
- The exact same transaction instructions and accounts
- Profiling/debugging of the transaction structure
- A clear state_warning in the response explaining this limitation

crates/cli/src/cli/replay/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+    // Step 1: Parse signatures from args, file, or environment variable
+    let signatures = if let Some(file_path) = &cmd.from_file {
+        parse_signatures_from_file(file_path)?
+    } else if !cmd.signatures.is_empty() {
+        cmd.signatures
+            .iter()
+            .map(|s| Signature::from_str(s))
+            .collect::<Result<Vec<_>, _>>()
+            .map_err(|e| format!("Invalid signature: {}", e))?
+    } else if let Ok(env_sigs) = env::var(TX_SIGNATURES_ENV) {
```
MicaiahReid, line 28: As far as I can tell, providing sigs via env var isn't documented in the CLI instructions - can you add those docs?

crates/cli/src/cli/replay/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+        return Err("No transaction signatures provided".to_string());
+    }
+
+    // Step 2: Determine RPC URL
+    let rpc_url = cmd.datasource_rpc_url();
+    println!("Fetching transactions from: {}", rpc_url);
+
+    // Step 3: Initialize ephemeral SVM (no persistence, no RPC server overhead)
+    let (surfnet_svm, _simnet_events_rx, _geyser_events_rx) =
+        SurfnetSvm::new_with_db(None, "replay")
```
MicaiahReid, line 47: `SurfnetSvm::default()` is a cleaner interface for this!

crates/cli/src/cli/replay/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+        .collect()
+}
+
+/// Prints a human-readable summary of a replay result.
+fn print_replay_result(result: &ReplayResult) {
+    println!("\n--- Transaction {} ---", result.signature);
+    println!(
+        "Status: {}",
+        if result.success { "SUCCESS" } else { "FAILED" }
+    );
```
MicaiahReid, line 156: Could we get some green/red color on success/fail to make it easy for the user to see success status?

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -268,6 +268,35 @@ pub trait SurfnetCheatcodes {
...
+    ///   "method": "surfnet_replayTransaction",
+    ///   "params": ["5N7Lw...", {"profile": true, "timeTravel": true}]
+    /// }
+    /// ```
+    #[rpc(meta, name = "surfnet_replayTransaction")]
+    fn replay_transaction(
+        &self,
+        meta: Self::Metadata,
+        signature: String,
+        config: Option<surfpool_types::ReplayConfig>,
```
lgalabru, line 297: The command 
```console
surfpool replay <tx-sig>
```
is terse and feels really cool.

Regarding the underlying cheatcode, I think being able to execute any transaction (including some that have not been broadcasted to mainnet) could be more useful (the CLI can still accept a `tx-sig`, fetch the transaction first and pass the transaction bytes).

I'm not incredibly excited by the idea of introducing a 4th entrypoint executing transaction, what's our take on augmenting `surfnet_profileTransaction` with a `time_travel_config` attribute?
Devs will most likely be playing / replaying to get all the data you can get out of the execution, which is what `surfnet_profileTransaction` gives you.

crates/core/src/surfnet/locker.rs
```diff
@@ -1084,6 +1084,202 @@ impl SurfnetSvmLocker {
...
+        let local_blockhash = self.with_svm_reader(|svm| svm.latest_blockhash());
+        let versioned_tx = replace_transaction_blockhash(versioned_tx, local_blockhash);
+
+        // Step 9: Time travel to transaction's slot (if enabled)
+        let replay_slot = if config.should_time_travel() {
+            let time_travel_config = TimeTravelConfig::AbsoluteSlot(original_slot);
+            match self.time_travel(None, simnet_command_tx.clone(), time_travel_config) {
+                Ok(epoch_info) => epoch_info.absolute_slot,
+                Err(_) => self.get_latest_absolute_slot(),
+            }
```
MicaiahReid, line 1213: So currently time travel to a past slot isn't actually supported - I think this is erroring every time, and the error is being swallowed here.

We're discussing internally how to handle, will get back to you!

crates/cli/src/cli/mod.rs
```diff
@@ -127,6 +128,9 @@ enum Command {
     /// Start MCP server
     #[clap(name = "mcp", bin_name = "mcp")]
     Mcp,
+    /// Replay mainnet transactions locally
+    #[clap(name = "replay", bin_name = "replay")]
```
MicaiahReid, line 132:
```suggestion
    #[clap(
        name = "replay",
        bin_name = "replay",
        about = "Re-execute transactions from the remote against a local surfnet",
        long_about = "Re-executes transactions from the remote against a local surfnet. \nNote: Different account states at execution time can lead to completely different transaction results. While replay can be helpful in transaction debugging and introspection, it is not meant to exactly replicate the remote transaction."
    )]
```

MicaiahReid (changes requested):

MicaiahReid:
@lgalabru and I discussed internally, and I think we'd feel more comfortable with a few high-level changes. We specifically worry about the `replay` wording being confusing, about the cheatcode muddying an already-running network's state, and internal state issues if we have to allow "backward" time travel. I think these changes would avoid that:
 1. Remove the cheatcode, only allowing this feature through a direct CLI call
   a. With this, in the future if we provide archive node access features, we can bake that right into the CLI call without mixing live/archive state in the cheatcode implementation
 2. Rename the command from `surfpool replay` to `surfpool profile` - it's essentially profiling a transaction. For now the input is a txid, in the future it could also be a base58 transaction or something
 3. Currently the `initialize` function is never called on the svm, so the initial epoch is never set. Then, `time_travel` is called to update the clock. Instead, the `initialize` function on the svm should be called, and the initial `EpochInfo` should be the epoch info of the transaction being profiled

Do those changes sound reasonable to you, @ExoMonk? Do you think you'll have the bandwidth to get this over the finish line?

marijamijailovic:
Hi @MicaiahReid, @lgalabru, I'm interested in picking this up and working on the transaction replay/profiling features. I've been going through the branch locally and would be happy to take it forward. Is it okay if I continue work on this PR?

marijamijailovic:
Hi @MicaiahReid, @lgalabru just a quick ping on this. Is there any update on the preferred direction for this PR, and is it okay if I address the comments above and work on getting it ready to merge?

MicaiahReid:
> Hi @MicaiahReid, @lgalabru just a quick ping on this. Is there any update on the preferred direction for this PR, and is it okay if I address the comments above and work on getting it ready to merge?

Hey @marijamijailovic! Sure, we'd love to see you take this on. This branch and my comments are probably a good starting place. However, solana foundation repositories require verified commit signatures, and the previous commits from @ExoMonk are not verified. So if you build off this branch, we won't be able to merge unless he [enables commit signature verification](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification). You can either ask him to enable and re-push, or just start your own branch from scratch and we can close this PR.

MicaiahReid:
@marijamijailovic this branch is pretty old, and a rebase might be brutal. Could be easier to start fresh.

marijamijailovic:
@MicaiahReid thanks for your replay! I will start working on the fresh branch from scratch!

PR #509, opened 2026-01-28
fix: graceful shutdown and PTY disconnect handling

ioxde:
Fixes https://github.com/txtx/surfpool/issues/507

Putting this PR in with a caveat/warning that I used [Claude Code to very much help with these changes](https://github.com/ghostty-org/ghostty/blob/main/AI_POLICY.md). [atty](https://github.com/softprops/atty?tab=readme-ov-file#atty) was removed as part of this PR because it has not been maintained in a long time and [standard IsTerminal](https://doc.rust-lang.org/std/io/trait.IsTerminal.html) is now the recommended replacement.

Additionally the tests that were added add ~5-6 seconds to the test suite. Ideally they would only be ran during CI / integration testing but there wasn't the foundation for that setup yet.

### DIFF
--- a/crates/cli/tests/shutdown.rs
+++ b/crates/cli/tests/shutdown.rs
@@ -201,6 +201,57 @@ mod shutdown_tests {
         assert_signal_exits_log_mode(libc::SIGHUP, "SIGHUP");
     }
 
+    #[test]
+    fn sigterm_with_db_causes_clean_exit() {
+        let temp_dir = std::env::temp_dir();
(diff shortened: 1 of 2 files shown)


MicaiahReid:
Hey @ioxde, thanks for the PR! Tbh, this is a lot of changes and I'm wary of merging. Have you personally recreated/tested the new codepaths you're adding?

ioxde:
I completely agree- going to close this. I'll open a new PR with a more TDD approach if needed / can find the time.

PR #513, opened 2026-01-29
fix: reset_network to correctly reset time info

### LINKED ISSUE #511: `surfnet_resetNetwork` doesn't reset time
Hi, it would be expected that `reset_network` also reset time/clock, but currently it doesn't happen. Bellow is small reproduction snippet (vibe coded, so a little verbose, but gets the point). You can call the script with local surfpool instance running. On current main the second time it prints this:

```
{
    "jsonrpc": "2.0",
    "error": {
        "code": -32603,
        "message": "Internal error",
        "data": "Cannot travel to past timestamp: target=3250368000000, current=3250368000400"
    },
    "id": 1
}
```
which means that the reset indeed didn't happen

<details>
  <summary>Snippet code</summary>
  
  
  ```bash
#!/bin/bash

# Reproduction: time travel, resetNetwork, then time travel again should work but fails

set -e

SURFNET_URL="${SURFNET_URL:-http://127.0.0.1:8899}"
TIMESTAMP=3250368000000

echo "================================================"
echo "Reproducing resetNetwork bug with time travel"
echo "================================================"
echo ""
echo "Target timestamp: $TIMESTAMP"
echo ""

echo "Step 1: First time travel to timestamp $TIMESTAMP"
echo "---"
RESPONSE1=$(curl -s -X POST "$SURFNET_URL" \
  -H "Content-Type: application/json" \
  -d '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "surfnet_timeTravel",
    "params": [{"absoluteTimestamp": '"$TIMESTAMP"'}]
  }')

echo "$RESPONSE1" | python3 -m json.tool

if echo "$RESPONSE1" | grep -q '"error"'; then
  echo ""
  echo "❌ First time travel failed"
  exit 1
else
  echo ""
  echo "✓ First time travel succeeded"
fi

echo ""
echo "Step 2: Reset network (should reset updated_at)"
echo "---"
RESET_RESPONSE=$(curl -s -X POST "$SURFNET_URL" \
  -H "Content-Type: application/json" \
  -d '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "surfnet_resetNetwork",
    "params": []
  }')

echo "$RESET_RESPONSE" | python3 -m json.tool

if echo "$RESET_RESPONSE" | grep -q '"error"'; then
  echo ""
  echo "❌ Reset network failed"
  exit 1
else
  echo ""
  echo "✓ Reset network succeeded"
fi

echo ""
echo "Step 3: Second time travel to same timestamp $TIMESTAMP (after reset)"
echo "---"
RESPONSE2=$(curl -s -X POST "$SURFNET_URL" \
  -H "Content-Type: application/json" \
  -d '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "surfnet_timeTravel",
    "params": [{"absoluteTimestamp": '"$TIMESTAMP"'}]
  }')

echo "$RESPONSE2" | python3 -m json.tool

if echo "$RESPONSE2" | grep -q '"error"'; then
  echo ""
  echo "❌ BUG CONFIRMED!"
  echo "Second time travel to same timestamp failed after resetNetwork"
  echo "Expected: should succeed because resetNetwork should have reset updated_at"
  exit 1
else
  echo ""
  echo "✓ Second time travel succeeded"
fi

echo ""
echo "================================================"
  ```
 
</details>

bonlime:
Fixes #511 

Tests were _vibed_ so if you think it's a little verbose, can shorten them

### DIFF
--- a/crates/core/src/rpc/surfnet_cheatcodes.rs
+++ b/crates/core/src/rpc/surfnet_cheatcodes.rs
@@ -778,7 +778,7 @@ pub trait SurfnetCheatcodes {
     ///
 
     #[rpc(meta, name = "surfnet_resetNetwork")]
-    fn reset_network(&self, meta: Self::Metadata) -> Result<RpcResponse<()>>;
+    fn reset_network(&self, meta: Self::Metadata) -> BoxFuture<Result<RpcResponse<()>>>;
 
     /// A cheat code to export a snapshot of all accounts in the Surfnet SVM.
     ///
@@ -1687,12 +1687,24 @@ impl SurfnetCheatcodes for SurfnetCheatcodesRpc {
         })
     }
 
-    fn reset_network(&self, meta: Self::Metadata) -> Result<RpcResponse<()>> {
-        let svm_locker = meta.get_svm_locker()?;
-        svm_locker.reset_network()?;
-        Ok(RpcResponse {
-            context: RpcResponseContext::new(svm_locker.get_latest_absolute_slot()),
-            value: (),
+    fn reset_network(&self, meta: Self::Metadata) -> BoxFuture<Result<RpcResponse<()>>> {
+        let SurfnetRpcContext {
+            svm_locker,
+            remote_ctx,
+        } = match meta.get_rpc_context(CommitmentConfig::confirmed()) {
+            Ok(res) => res,
+            Err(e) => return e.into(),
+        };
+
+        // Extract just the remote client from the tuple (ignore commitment config)
+        let remote_client = remote_ctx.as_ref().map(|(client, _)| client.clone());
+
+        Box::pin(async move {
+            svm_locker.reset_network(&remote_client).await?;
+            Ok(RpcResponse {
+                context: RpcResponseContext::new(svm_locker.get_latest_absolute_slot()),
+                value: (),
+            })
         })
     }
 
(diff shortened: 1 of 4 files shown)


MicaiahReid:
Hmmm, I'm not sure about this.

When you first start a surfnet, the epoch info will come from the mainnet datasource. Now, we're clearing it out to zero.

I'm not sure what the expected behavior should be, one of these:
 1. Clear it out to zero like you have it
 2. When we first set the Epoch info after fetching from mainnet, cache "genesis info", and reset to that
 3. On reset, refetch from mainnet and have that new value be our epoch info

I'm of the opinion that we should go with option 2. Thoughts, @bonlime, @lgalabru?

bonlime:
that's a good point, I personally would go with option 3, since it would be closer to "full reset", I'll see if it's easy enough to do

lgalabru:
I'd go with option 3 when the surfnet has a datasource, and keep option 1 when the surfnet is `offline`

MicaiahReid:
Okay, sounds good. Let's go with option 3, @bonlime. Thanks!

MicaiahReid:
Hey, @bonlime, just checking in - do you think you'll have time to get this over the finish line?

bonlime:
Hey, i've tried to implement this, but struggled with rust async, since i'm not very proficient with it... if you could implement it, it probably would be faster

lgalabru:
hey @bonlime, thanks for your contribution! 
As long as your additions are tested, it's fine by us if you get a bit of help from Claude!

bonlime:
@lgalabru I've tried. The problem is that it requires changing previously sync functions to async to allow for rpc requests, in multiple attempts I did it resulted in pretty significant functions signatures changed, so I threw it away. it definitely needs some guided input from a person much better in both rust and this code base

PR #516, opened 2026-02-02
fix: share SQLite connection pools across storage instances

ioxde:
Using macOS `15.7.3` I was hitting "too many open files" running `surfpool start --no-tui --db ./surfnet.sqlite`

Same fix as postgres - share pools across storage instances. Pragmas now go through r2d2's CustomizeConnection.

Tests in 8b61fc8 show the issue and verify the fix.


ioxde:
No problem! I just pushed a fixed for the failing test the PR caused.

PR #519, opened 2026-02-03
fix: offline slots_in_epoch

Arrowana:
Fix for `surfpool start --offline` immediately panicking due to `slots_in_epoch` propagated being 0
thread 'main' panicked at crates/cli/src/tui/simnet.rs:346:33: attempt to calculate the remainder with a divisor of zero

### DIFF
--- a/crates/core/src/surfnet/locker.rs
+++ b/crates/core/src/surfnet/locker.rs
@@ -69,7 +69,7 @@ use crate::{
     error::{SurfpoolError, SurfpoolResult},
     helpers::time_travel::calculate_time_travel_clock,
     rpc::utils::{convert_transaction_metadata_from_canonical, verify_pubkey},
-    surfnet::FINALIZATION_SLOT_THRESHOLD,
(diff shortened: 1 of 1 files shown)


crates/cli/src/tui/simnet.rs
```diff
@@ -343,6 +343,7 @@ impl App {
 
     pub fn epoch_progress(&self) -> u16 {
         let absolute = self.slot() as u64;
+        println!("{}", self.epoch_info.slots_in_epoch);
```
MicaiahReid:
```suggestion
```

MicaiahReid (changes requested):
Good find, thanks @Arrowana! One small change, then let's get this merged!

crates/cli/src/tui/simnet.rs
Arrowana: Done

lgalabru (approved):
Thanks @Arrowana!

PR #527, opened 2026-02-11
fix(geyser): align commitment slots and remove duplicate account re-emits

bonlime:
Hi, submitting a minor fix for geyser. I know that currently the `geyser_plugin` is disabled due to conflict with litesvm, so it's a little hard to test this PR, locally i've resolved the dependancy issues by cloning/updating the deps version, so it's ok for this PR to sit for a while until something proper is done for geyser support in main

**Summary of the changes:**
- Removes duplicate account re-emits during confirmation for `processed` commitment
- Keeps account mutation tracking (account_update_slots) without sending extra synthetic account updates
- Aligns geyser commitment progression to the same slot used by processed tx/account updates:
        * UpdateSlotStatus::Confirmed now emits with slot (not slot + 1)
        * NotifyBlockMetadata now emits with slot and parent derived from that same slot
        * NotifyEntry now emits with slot.
         * AccountUpdates always have proper `txn_signature` set, currently it's NOT set for "confirmed" commitment

**Why:**
The previous flow produced duplicate account updates and inconsistent slot mapping across processed/confirmed paths, which made commitment behavior confusing in gRPC consumers.

**Impact:**
Account streams no longer get duplicate processed-like updates from confirmation re-emits.
Commitment promotion (processed -> confirmed -> finalized) is now based on a consistent slot identity.

### DIFF
--- a/crates/core/src/surfnet/locker.rs
+++ b/crates/core/src/surfnet/locker.rs
@@ -4443,13 +4443,14 @@ mod tests {
 
     #[tokio::test(flavor = "multi_thread")]
     async fn test_v0_transaction_without_alt_emits_geyser_account_updates() {
+        use std::time::Duration;
+
         use crossbeam_channel::{RecvTimeoutError, unbounded};
         use solana_keypair::Keypair;
         use solana_message::{VersionedMessage, v0};
         use solana_signer::Signer;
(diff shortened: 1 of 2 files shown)


MicaiahReid (approved):
Looks good! We can merge once geyser is on main again

PR #533, opened 2026-02-18
feat(cli): add --skip-signature-verification flag

bonlime:
Add a `--skip-signature-verification` CLI flag to `surfpool start` that exposes the existing `skip_signature_verification` field in `SimnetConfig`, which was previously hardcoded to `false` with no way to enable it from the command line.

related to #456


MicaiahReid:
Nice, thanks @bonlime!

PR #535, opened 2026-02-20
added fix to the ls command when txtx.yaml file doesn't exist

0xzrf:
## Description

Running `surfpool ls` on a folder that doesn't have txtx.yaml file in it(or even provided in cli) doesn't output anything

## Steps to Reproduce

1. Run `surfpool ls` on a folder that doesn't have txtx.yaml

## Current Behavior

The process exits with a non-zero value, but doesn't log anything that might be helpful to the user

## Expected Behavior

The command should:
- Show some error why specifically the command not work

## What this PR Solves
This isn't solving a huge problem, but it's still a contribution, as I see it doesn't help the user to know what went wrong. This PR aims to add a simple error handling to the `surfpool_cli/src/cli/mod.rs/FN:handle_list_command` to show error when something goes wrong while executing the `surfpool ls` command

### DIFF
--- a/crates/cli/src/cli/mod.rs
+++ b/crates/cli/src/cli/mod.rs
@@ -7,7 +7,7 @@ use fern::colors::{Color, ColoredLevelConfig};
 #[cfg(not(target_os = "windows"))]
 use fork::{Fork, daemon};
 use hiro_system_kit::{self, Logger};
-use log::{error, info};
+use log::info;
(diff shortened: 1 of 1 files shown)


MicaiahReid:
Thanks for the PR, @0xzrf! I changed slightly to more generally improve the error displaying. Will merge once CI runs!

PR #544, opened 2026-02-28
feat(core): implement transaction signature verification and reject duplicates

MicaiahReid:


lgalabru (approved):
Per @Aursen, this behavior is by default implemented in LiteSVM. Could we please add a comment explaining why we had to disable it?

MicaiahReid:
> Per @Aursen, this behavior is by default implemented in LiteSVM. Could we please add a comment explaining why we had to disable it?

I wanted the same thing, but tbh I couldn't find our reasoning...
I remember it causing issues in the past, but I don't remember the cause and couldn't find issues relating to it.

I was worried about the issue arising again if I enabled it blindly.

I did some digging again and I have a _possible_ reason: the LiteSVM `sigverify` being set to true will verify transaction simulations as well as actual transaction processing. So we'd have to temporarily disable in the `svm.rs` file via:
```rust
fn simulate_transaction(&self, sigverify: bool) ... {
if !sigverify {
   self.sigverify(false);
}

/// simulate

// reset it for future tx processing
if !sigverify {
    self.sigverify(true);
}
```
which would also force `simulate_transaction` to require `&mut self`, and seems odd to have mutable vm for a simulate endpoint.

Don't know if that's a strong enough reason to have this extra code to maintain, though.

PR #545, opened 2026-02-28
fix(cli): update accounts initialization to use validator account from anchor.toml

MicaiahReid:


lgalabru (approved):
Thanks @MicaiahReid!

PR #546, opened 2026-02-28
fix: order signatures for getSignaturesForAddress

MicaiahReid:


crates/core/src/surfnet/locker.rs
```diff
@@ -871,7 +881,19 @@ impl SurfnetSvmLocker {
...
+                            .get(&a.signature)
+                            .copied()
+                            .unwrap_or(usize::MAX);
+                        let b_pos = sig_position
+                            .get(&b.signature)
+                            .copied()
+                            .unwrap_or(usize::MAX);
+                        b_pos.cmp(&a_pos)
+                    })
+                })
```
lgalabru, line 899: I don't know how hot this codepath is, but it looks like we're doing a lot of cloning / copy.
When we started building Surfpool, we had limited resources and our core constraint was time to market, joining Foundation is removing some of this stress - I think we can now spend a few additional cycles on shipping optimized / memory-efficient routines.

lgalabru (approved):
LGTM, thanks @MicaiahReid!

crates/core/src/surfnet/locker.rs
MicaiahReid: Fair enough! All clones were of `Option<u64>`, which isn't too bad, but agreed on updating our general approach to start optimizing. Pushed a fix! 35a1833ba69010c97ab93f9f9f0b555ddd023b95

PR #547, opened 2026-02-28
fix: use the config arg in the get_balance rpc method

0xzrf:
## Summary

This PR refactors `SurfpoolMinimalRpc::get_balance()` to use the `config: Option<RpcContextConfig>` argument, which provides the `commitment: Option<CommitmentConfig>` and `min_context_slot: Option<u64>` variables, in order to customize the get_balance rpc method further

##  What Changed

- Extracted `commitment` and `min_context_slot` variables, and provided `Commitment::confirmed()`(the default earilier) to commitment and `None` to `min_context_slot` in case `None`
-  use the `commitment` provided in config inside the `meta.get_rpc_context()` in case present, or `Commitment::confirmed()` otherwise
- Use the `min_context_slot` value (if present) to validate the `svm_locker.get_account().slot`
- Added a `SurfpoolError::get_balance()` method that returns a `SurfpoolError::internal_error()` with `error.data = "Context slot too old"`

##  Context

This change was motivated by an existing comment inside `SurfpoolMinimalRpc::get_balance()` that suggested using the config in the function. This PR increases the configurability of the rpc's `get_balance` method by allowing to provide the commitment level and min_context_slot as arguments

## Note
I may or may not have used the `min_context_slot` value properly, as I've assumed that it's used to validate the value `svm_locker.get_account().slot` and require it to be bigger then then the argument. I'd be happy to change it's use case in case my assumption is wrong. Thanks


crates/core/src/rpc/minimal.rs
```diff
@@ -608,6 +621,12 @@ impl Minimal for SurfpoolMinimalRpc {
                 ..
             } = svm_locker.get_account(&remote_ctx, &pubkey, None).await?;
 
+            if let Some(min_slot) = min_ctx_slot
+                && slot < min_slot
+            {
+                return Err(SurfpoolError::get_balance().into());
```
MicaiahReid: There's an existing error for this, `RpcCustomError::MinContextSlotNotReached`. You can see it in other RPC methods - can we use that instead?

crates/core/src/rpc/minimal.rs
```diff
@@ -586,17 +587,29 @@ impl Minimal for SurfpoolMinimalRpc {
...
+                commitment: Some(commitment),
+                min_context_slot,
+            }) => (commitment, min_context_slot),
+            Some(RpcContextConfig {
+                commitment: None,
+                min_context_slot,
+            }) => (CommitmentConfig::confirmed(), min_context_slot),
+            None => (CommitmentConfig::confirmed(), None),
+        };
+
```
MicaiahReid: I think it's cleaner to use `unwrap_or_default`, and it will allow the underlying types to drive the default value
```suggestion
        let config = _config.unwrap_or_default();
        let commitment_config = config.commitment.unwrap_or_default().commitment;
        let min_context_slot = config.min_context_slot;
```

0xzrf: got it, will change it right away!
Thanks for the feedback

0xzrf:
@MicaiahReid I've made the changes
lmk if you have any feedbacks. Thanks

crates/core/src/error.rs
```diff
@@ -103,7 +103,11 @@ impl SurfpoolError {
         error.data = Some(json!("Failed to access internal Surfnet context"));
         Self(error)
     }
-
+    pub fn get_balance() -> Self {
+        let mut error = Error::internal_error();
+        error.data = Some(json!("Context slot too old"));
+        Self(error)
+    }
```
MicaiahReid: Can we remove this error that is no longer being used?

0xzrf: Got it

0xzrf:
@MicaiahReid added some tests that check whether the `get_balance` rpc method fail when the `min_context_slot` is bigger then `latest_absolute_slot()`

lmk if anything else is required. Thanks

0xzrf:
@MicaiahReid Added config support for the `SurfpoolFullRpc::get_fee_for_message` in the last commit, ik the PR specifically fixes the get_balance, but theoratically the fixes for both the rpc methods felt the same.

crates/core/src/rpc/minimal.rs
```diff
@@ -972,6 +985,59 @@ mod tests {
...
+        assert!(
+            pass_if_correct_config_result.is_ok(),
+            "Expected the operation to pass"
+        );
+
+        assert!(
+            pass_if_correct_config_result.unwrap().value == airdrop_amount,
+            "Invalid returned lamports for the account"
+        );
```
MicaiahReid: Let's change this to `assert_eq`. With `assert_eq`, if the values are different, the failure logs will tell us the value of the left and right side. With `assert`, we'd only see that the values are different.

crates/core/src/rpc/minimal.rs
```diff
@@ -972,6 +985,59 @@ mod tests {
...
+                Some(setup.context.clone()),
+                Pubkey::new_unique().to_string(),
+                Some(RpcContextConfig {
+                    commitment: None,
+                    min_context_slot: Some(latest_slot + 100),
+                }),
+            )
+            .await;
+
+        assert!(
```
MicaiahReid, line 1043: Can we assert the actual error matches the expected error?

crates/core/src/rpc/full.rs
```diff
@@ -2645,6 +2671,92 @@ mod tests {
...
+            "Invalid return value"
+        );
+
+        let get_fee_with_wrong_commitment_fail_result = rpc_server.get_fee_for_message(
+            Some(runloop_context.clone()),
+            encoded_message.clone(),
+            Some(rpc_ctx_config_with_wrong_commitment),
+        );
+
+        assert!(
```
MicaiahReid, line 2751: Can we assert the actual error matches the expected error?

crates/core/src/rpc/full.rs
```diff
@@ -2645,6 +2671,92 @@ mod tests {
...
+            "expected this txn to fail when min_ctx_slot > slot_for_commitment"
+        );
+
+        let get_fee_with_wrong_mint_slot_fail_result = rpc_server.get_fee_for_message(
+            Some(runloop_context.clone()),
+            encoded_message,
+            Some(rpc_ctx_config_with_wrong_min_slot),
+        );
+
+        assert!(
```
MicaiahReid, line 2773: Can we assert the actual error matches the expected error?

crates/core/src/tests/integration.rs
```diff
@@ -1268,6 +1270,87 @@ async fn test_get_transaction_profile(test_type: TestType) {
     );
     println!("All get_transaction_profile tests passed successfully!");
 }
+#[test_case(TestType::sqlite(); "with on-disk sqlite db")]
+#[test_case(TestType::in_memory(); "with in-memory sqlite db")]
+#[test_case(TestType::no_db(); "with no db")]
+#[cfg_attr(feature = "postgres", test_case(TestType::postgres(); "with postgres db"))]
+#[tokio::test(flavor = "multi_thread")]
+async fn test_get_fee_for_messaage_config_tests(test_type: TestType) {
```
MicaiahReid: Is this test a duplicate of the one you added in full.rs? If so, can we remote it here?

Issue #530, opened 2026-02-17
Support Token-2022 Account Reallocation


R-K-H:
Surfpool does not support Token-2022 account reallocation. 

For our specific use in running simulations for our MetaDAO Launches:
When completeLaunch runs, it creates cpamm pools which require reallocating Token-2022 accounts to a larger size. Surfpool fails on this operation.

This involves the Meteora CPAMM as it's a component so not sure what else we can do.

Let me know if you'd like additional information.

R-K-H:
# Surfpool Bug: Token-2022 `realloc` fails during `TokenMetadataInstruction::Initialize`

## Summary

Token-2022's `TokenMetadataInstruction::Initialize` fails with `"Failed to reallocate account data"` when called via CPI from cpamm's `InitializePoolWithDynamicConfig`. The instruction needs to grow a mint account to store token metadata, which requires a `realloc` syscall that Surfpool does not support.

This same transaction succeeds on mainnet.

## Environment

- Surfpool image: `surfpool/surfpool:latest` (v1.0.0-rc1)
- Running in Docker
- Surfpool datasource: Helius mainnet RPC

## Programs Involved

| Program | Address |
|---------|---------|
| Launchpad (MetaDAO) | `moontUzsdepotRGe5xsfip7vLPTJnVuafqdUWexVnPM` |
| Futarchy | `FUTARELBfJfQ8RDGhg1wdhddq1odMAJUePHFuBYfUxKq` |
| cpamm | `cpamdpZCGKUy5JxQXB4dcpGPiikHawvSWAd6mEn1sGG` |
| Token-2022 | `TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb` |
| SPL Token | `TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA` |
| Associated Token | `ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL` |
| Squads v4 | `SQDS4ep65T869zMMBKyuUq6aD6EgTu8psMjkvj52pCf` |
| System Program | `11111111111111111111111111111111` |
| Compute Budget | `ComputeBudget111111111111111111111111111111` |

## Error

```
Transaction simulation failed: Error processing Instruction 2: Failed to reallocate account data
```

## Exact Failure Point

The transaction executes a complex `CompleteLaunch` instruction that succeeds through many sub-operations before failing at one specific point:

```
[...]

PR #548, opened 2026-03-03
test: add regression test for token 2022 realloc

MicaiahReid:
Adds regression test for #530


lgalabru (approved):
LGTM, thanks @MicaiahReid!

PR #551, opened 2026-03-04
ci: add workflow for building and pushing Docker images

MicaiahReid:


lgalabru (approved):
Let's try this approach, but unless I'm missing some nuances I believe we went down that path in the past, and producing a multi-arch image through buildx may take a few hours. 

If this approach doesn't pan-out, we should probably explore spinning up a Github ARM builder.

MicaiahReid:
> Let's try this approach, but unless I'm missing some nuances I believe we went down that path in the past, and producing a multi-arch image through buildx may take a few hours.
> 
> If this approach doesn't pan-out, we should probably explore spinning up a Github ARM builder.

Gotcha, I wasn't tracking that's what you meant in our previous convo. It looks like arm runners are free for public repos! Updated to try building with an arm runner, I'll merge and test + update if needed.

PR #555, opened 2026-03-05
feat: improve scenario

lgalabru:
With this PR, we're bringing `Scenario` a bit closer to the finish line, but we still have a lot of remaining work to wrap up this epic.


crates/core/src/scenarios/protocols/raydium/v4/overrides.yaml
```diff
@@ -0,0 +1,517 @@
...
+
+constants:
+  openbook_market:
+    label: OpenBook Market
+    description: Select an OpenBook market for the AMM v4 pool
+    options:
+      - id: "sol-pippin"
+        label: "SOL/pippin"
+        value: "4Zm2A7VUfuvudpVpQkJyrg2DiZTcw6ckwbKP9LvizAkx"
+        description: "SOL/pippin AMM v4 pool (30d vol: $593.1M)"
```
MicaiahReid: All of the `(30d vol: )` comments I think are not useful/relevant here

crates/cli/src/cli/simnet/mod.rs
```diff
@@ -68,6 +68,24 @@ pub async fn handle_start_local_surfnet_command(
     let feature_config = cmd.feature_config();
     surfnet_svm.apply_feature_config(&feature_config);
 
+    // Restore from snapshots if specified
+    for snapshot_path in &cmd.snapshot {
+        let resolved_path = super::resolve_path(snapshot_path);
+        let path_str = resolved_path.to_string_lossy();
+        info!("Restoring state from snapshot: {}", path_str);
+        match surfnet_svm.restore_from_snapshot(&path_str) {
```
MicaiahReid: We have a very similar code path to `surfnet_svm.restore_from_snapshot` at https://github.com/txtx/surfpool/blob/9cb2398b2cf5caf1969aa9715bcb5a45e7c4c49e/crates/core/src/surfnet/locker.rs#L481

It is initialized a few lines below this (line 89 of this file) where the snapshot file contents are processed and stored on the simnet config.

That one does a few extra things - allowing useres to provide `<pubkey>: null` in their snapshot to fetch from remote. 

Is this current loop causing a double import? Do we want to go with just one of these approaches?

crates/core/src/scenarios/registry.rs
```diff
@@ -65,11 +73,11 @@ impl TemplateRegistry {
...
-            SWITCHBOARD_ON_DEMAND_IDL_CONTENT,
-            SWITCHBOARD_ON_DEMAND_OVERRIDES_CONTENT,
-            "switchboard-on-demand",
-        );
+        // self.load_protocol_overrides(
+        //     SWITCHBOARD_ON_DEMAND_IDL_CONTENT,
+        //     SWITCHBOARD_ON_DEMAND_OVERRIDES_CONTENT,
+        //     "switchboard-v2",
+        // );
     }
```
MicaiahReid: Remove?

MicaiahReid (dismissed):

crates/cli/src/cli/simnet/mod.rs
lgalabru: Done in the rebase plus 5603219. Removed this early restore loop so snapshots are only parsed into `cmd.accounts.snapshot` and loaded through the existing config/locker path, preserving the `<pubkey>: null` remote-fetch behavior and avoiding double import.

crates/core/src/scenarios/protocols/raydium/v4/overrides.yaml
lgalabru: Done in 5603219. Removed the `(30d vol: ...)` suffixes from the Raydium v4 YAML descriptions and updated `generate-markets.js` so regenerated market options do not reintroduce them.

crates/core/src/scenarios/registry.rs
lgalabru: Done in 5603219. Removed the disabled Switchboard loader call, constants, empty method, and ignored tests instead of keeping the commented-out registration path.

crates/types/src/scenarios.rs
```diff
@@ -36,6 +87,288 @@ pub enum PdaSeed {
...
+            PdaSeed::U16BeRef(prop) => {
+                values?.get(prop).and_then(|v| {
+                    // Handle numeric values - convert to u16 big-endian
+                    if let Some(n) = v.as_u64() {
+                        let n16 = n as u16;
+                        return Some(n16.to_be_bytes().to_vec());
+                    }
+                    None
+                })
+            }
```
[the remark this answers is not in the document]
lgalabru: Fixed in f9e6e38. `U16BeRef` now uses a checked `u16::try_from(n).ok()?` conversion and has tests for both in-range encoding and out-of-range rejection.

crates/types/src/verified_tokens.rs
```diff
@@ -1,20 +1,90 @@
...
+        field[1..field.len() - 1].replace("\"\"", "\"")
+    } else {
+        field.to_string()
+    }
+}
+
+/// Parse a CSV line into fields, handling quoted strings with commas
+fn parse_csv_line(line: &str) -> Vec<String> {
+    let mut fields = Vec::new();
+    let mut current = String::new();
```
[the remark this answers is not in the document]
lgalabru: Fixed in 4a43194. `parse_csv_field` now only trims after `parse_csv_line` handles quoting, and `parse_csv_line_preserves_literal_edge_quotes` covers the literal edge-quote case.

crates/mcp/src/surfpool/mod.rs
```diff
@@ -494,41 +530,200 @@ impl Surfpool {
...
+                                                    serde_json::Value::String(value.clone()),
+                                                );
+                                            }
+                                        }
+                                    }
+                                }
+                            }
+                            _ => {}
+                        }
+                    }
```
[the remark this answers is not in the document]
lgalabru: Fixed in f9e6e38. `create_scenario` now rebuilds `override_instance.account` from the template only for PDA templates; fixed-address templates preserve a caller-supplied pubkey.

lgalabru:
Addressed the concrete Greptile summary findings in f9e6e38: compact template responses now include `constants` from both `get_override_templates` and the `str:///override_templates` resource, and `U16BeRef` rejects out-of-range values instead of wrapping.

crates/types/src/verified_tokens.rs
```diff
@@ -1,20 +1,103 @@
...
     pub logo_uri: Option<String>,
 }
 
+/// Normalize a CSV field after `parse_csv_line` has already handled quoting.
+fn parse_csv_field(field: &str) -> String {
+    field.trim().to_string()
+}
+
+/// Parse a CSV line into fields, handling quoted strings with commas
+fn parse_csv_line(line: &str) -> Vec<String> {
```
BriungRi, line 20: why not use something like `.split` and prefer using TSV over CSV to avoid the comma in quote issue?

BriungRi (approved):
one comment

crates/types/src/verified_tokens.rs
lgalabru: Good idea. created an issue [here](https://github.com/solana-foundation/surfpool/issues/713)

PR #559, opened 2026-03-06
fix: correct bool logic in get_multiple_accounts from remote

MicaiahReid:
[This PR](https://github.com/solana-foundation/surfpool/pull/542) introduced a subtle bug when fetching multiple accounts from the remote:
 - a boolean logic bug (`||` -> `&&`) was preventing some accounts from getting fetched
 - reordering of a result led to an errant `zip` among two vecs

[This PR](https://github.com/solana-foundation/surfpool/pull/545) introduced loading accounts on startup when there's a `test.validator.account` provided in an Anchor.toml. However, it shouldn't pull those accounts in if there's a `Test.toml` provided as well (`Test.toml` overrides)

Both bugs are fixed, with tests added

### DIFF
--- a/crates/cli/src/scaffold/anchor.rs
+++ b/crates/cli/src/scaffold/anchor.rs
@@ -8,6 +8,7 @@ use std::{
 
 use anyhow::{Result, anyhow};
 use convert_case::{Case, Casing};
+use log::debug;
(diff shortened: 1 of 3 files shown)


lgalabru (approved):
Thanks @MicaiahReid!

PR #564, opened 2026-03-10
fix: reject duplicate accounts between static keys and ALT

lgalabru:
### Summary
- Detect when a V0 transaction has the same account in both static keys and Address Lookup Table (ALT) loaded addresses, and reject pre-execution with AccountLoadedTwice — matching Agave behavior
- Route AccountLoadedTwice errors through the SimulationFailure path to produce Agave-compatible JSON-RPC error format `({"err":"AccountLoadedTwice","logs":[],"unitsConsumed":0,...})`
- Add integration test covering the duplicate account rejection

### Context
When a V0 transaction references system_program (or any account) in both its static keys and an ALT, Agave rejects it pre-execution with AccountLoadedTwice (0 CU consumed). Surfpool was letting it through to execution, where it failed with MissingAccount (~13k CU consumed) — a different error, at a different stage, with different CU cost.

Root cause: `get_pubkeys_from_message()` concatenated static keys with ALT-loaded addresses without deduplication.

### Test plan
- Unit tests pass (3 variants: no db, in-memory sqlite, on-disk sqlite)
- End-to-end Go reproducer passes (solana-anchorage-collab repo) — confirms Agave-compatible JSON-RPC error format
- Fix-path transaction (no duplicate accounts) still succeeds

### DIFF
--- a/crates/core/src/surfnet/locker.rs
+++ b/crates/core/src/surfnet/locker.rs
@@ -1064,14 +1064,13 @@ impl SurfnetSvmLocker {
                 // Agave-compatible JSON-RPC error format with structured `err` and `data` fields.
                 let err_str = e.to_string();
                 if err_str.contains("Account loaded twice") {
-                    let _ =
-                        status_tx.try_send(TransactionStatusEvent::SimulationFailure((
-                            TransactionError::AccountLoadedTwice,
-                            surfpool_types::TransactionMetadata::default(),
-                        )));
+                    let _ = status_tx.try_send(TransactionStatusEvent::SimulationFailure((
+                        TransactionError::AccountLoadedTwice,
+                        surfpool_types::TransactionMetadata::default(),
+                    )));
                 } else {
-                    let _ = status_tx
-                        .try_send(TransactionStatusEvent::VerificationFailure(err_str));
+                    let _ =
+                        status_tx.try_send(TransactionStatusEvent::VerificationFailure(err_str));
                 }
                 return Err(e);
             }


crates/core/src/surfnet/locker.rs
```diff
@@ -1059,8 +1059,19 @@ impl SurfnetSvmLocker {
...
                 // account fetching, ALT resolution, or other pre-processing steps.
                 // This is critical for issue #454 where program close stops block production.
-                let _ =
-                    status_tx.try_send(TransactionStatusEvent::VerificationFailure(e.to_string()));
+                //
+                // AccountLoadedTwice errors should go through SimulationFailure to produce
+                // Agave-compatible JSON-RPC error format with structured `err` and `data` fields.
+                let err_str = e.to_string();
+                if err_str.contains("Account loaded twice") {
+                    let _ = status_tx.try_send(TransactionStatusEvent::SimulationFailure((
```
MicaiahReid, line 1067: Are we sure `SimulationFailure` is the correct error here? What if simulation is disabled? (`skip_preflight = true`)

crates/core/src/surfnet/locker.rs
```diff
@@ -1144,6 +1155,18 @@ impl SurfnetSvmLocker {
             .get_loaded_addresses(remote_ctx, &transaction.message)
             .await?;
 
+        // Check for duplicate accounts between static keys and ALT-loaded addresses.
+        // Agave rejects such transactions pre-execution with AccountLoadedTwice.
+        if let Some(ref loaded) = tx_loaded_addresses {
```
MicaiahReid, line 1160: Should this happen before we fetch all of the accounts in the line above?

crates/core/src/surfnet/locker.rs
lgalabru:
```
enum ProcessTransactionResult {
    Success(TransactionMetadata),
    SimulationFailure(FailedTransactionMetadata),
    ExecutionFailure(FailedTransactionMetadata),
}
```
Not success, and the transaction is not executing at this point. We could introduce a new enum but it seems overkill.

crates/core/src/surfnet/locker.rs
lgalabru: I don't understand your feedback / change requests. Could you please ask the question differently?

crates/core/src/surfnet/locker.rs
MicaiahReid: Yeah I'm saying I think to better match with what a node would return, if `skip_preflight = true` we return an `ExecutionFailure` - because we're not simulating the node would (I believe) return with the same error but while actually executing the transaction. Vs if `skip_preflight = false`, on a real node you'd be simulating and you'd return a `SimulationFailure`

crates/core/src/surfnet/locker.rs
MicaiahReid: Ohp no I misread the code - I thought we weren't depending on `tx_loaded_addresses` loaded above and wanted to do this check before making the `.get_loaded_addresses` call to fetch accounts. No action here!

crates/core/src/surfnet/locker.rs
lgalabru: This check is pre-execution / pre-simulation, and since the transaction is not executed (0 CU), simulation seems to be the least worst to me. 
Neither of these variant are semantically correct: with the approach you're suggesting (`ExecutionFailure`), if we use the simulation rpc endpoint and hit that code-path, we'd be returning an `execution` error.
 
I honestly don't think it's a big deal as this enum is not exposed anyway in the RPC response.

MicaiahReid: > this enum is not exposed anyway in the RPC response

That's not true. The `send_transaction` RPC method recives this `TransactionStatusEvent` and reacts differently according to the variant. `VerificationFailure` has no data on the final error and just has a message. `SimulationError` returns a `RpcSimulateTransactionResult` on the data of the error. `ExecutionFailure` is completely ignored - it's processed by the VM and the user would need to fetch the transaction by signature to see the error.

So perhaps your point is shining light on that we're further deviating from how mainnet would actually handle - if you skip the simulation on mainnet do you get the same error contents ("Account loaded twice") but during execution? Why isn't LiteSVM returning the correct error in the first place.

This is feeling more like a bandaid rather than a fix -perhaps this should be at the LiteSVM level.

> if we use the simulation rpc endpoint and hit that code-path
Simulation RPC doesn't use this code path, I don't think that's relevant

MicaiahReid: > This check is pre-execution / pre-simulation

Agreed - what I'm saying is that on mainnet this check _isn't_ pre-execution/pre-simulation - so how can we have surfpool either behave the same way or appear to behave the same way

lgalabru: > on mainnet this check isn't pre-execution/pre-simulation

Wdym? it has nothing to do with mainnet / devnet, it's a client implementation detail.

With this patch, behavior Agave and Surfpool are identical - as explained in the description, the PR was built with a test vector Surfpool / Agave.

MicaiahReid: I just noticed this in the PR description:
>Root cause: get_pubkeys_from_message() concatenated static keys with ALT-loaded addresses without deduplication.

Would fixing that root cause allow LiteSVM to return the correct error?

lgalabru: LiteSVM should not be involved at all.

PR #577, opened 2026-03-18
feat(cli): allow passing feature pubkey instead of name on startup

MicaiahReid:
83ab44ac1d1ff666677ec157d4fd38bdd24e71c2 - feature accounts didn't exist in VM at runtime (when referenced in a program), because litesvm doesn't store them. Temporary fix until litesvm fix is in. [Issue to remove todo's](https://github.com/solana-foundation/surfpool/issues/578)
7ee8fe248d3f32dd7c6093290d8e418076f5eae0 - litesvm references `feature_set`s injected from `with_feature_set` in subsequent `with_builtins`, `with_sysvars`, and `with_default_programs` calls. So, when we update feature sets, call these fns
e5336bf44cb1d1b831d5a5cdaae42db13cf90245 - Update `--feature <kebab-case-feature-name>` to take `--feature <pubkey>`. This should prevent breaking the feature name for now, but we'll be able to remove a bunch of code once we deprecate that feature. I've added the deprecation notice to the CLI flags

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -281,9 +281,9 @@ dependencies = [
 
 [[package]]
 name = "agave-feature-set"
-version = "3.1.6"
+version = "3.1.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "53e0d37e3232be339a6a19a376c7696458b6a8183e95bf806709fad71595010b"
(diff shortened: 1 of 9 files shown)


lgalabru (approved):
Awesome finding, thanks for the clean-up!

PR #579, opened 2026-03-18
fix: improve write_program logic; add tests

MicaiahReid:
This PR:
 - fixes write_program fn in locker.rs
   - to summarize, when you call `set_account` on litesvm, it checks if that account is executable and loads it into a cache of programs. I wasn't aware of this. So when we would 1. write program account, 2. write program data account, the program account wasn't being re-indexed with the new program data. This would cause program updates to fail beyond the initial program written
 - slimmed down the minimum program we deploy
   - when you write am account using set_account, if the account is executable, it's also going to verify that there is an associated program data account, and verify the ELF of the program. So, if a program account is being written without the data account, we're pushing a minimal ELF. I've updated this to be smaller
 - tests for the edge cases that I hit


crates/core/src/surfnet/noop_program.rs
```diff
@@ -0,0 +1,99 @@
...
+/// Layout (352 bytes):
+///   0x000..0x040  ELF64 header (ET_DYN, EM_SBPF=0x107, SBPFv0, entry=0x78)
+///   0x040..0x078  1× PT_LOAD program header (text at 0x78, 16 bytes, RX)
+///   0x078..0x088  .text section: mov64 r0,0 ; exit
+///   0x088..0x099  .shstrtab contents: "\0.text\0.shstrtab\0"
+///   0x099..0x0A0  padding (7 bytes, align section headers to 8)
+///   0x0A0..0x0E0  Section header [0]: SHT_NULL
+///   0x0E0..0x120  Section header [1]: .text
+///   0x120..0x160  Section header [2]: .shstrtab
+pub const NOOP_PROGRAM_ELF: &[u8] = &[
```
lgalabru, line 17: 🤷‍♂️

lgalabru (approved):
LGTM, thanks @MicaiahReid !

Rmrifa:
AHBQEXWWVHQAPIYAAAAI7Z5Y5EZVH5VBFDLGC6PWABDS2STYU24AWLWZ5YOQAHLBODIXKFA

PR #581, opened 2026-03-19
test: add test using skipSigVerify in sendTransaction call

MicaiahReid:


lgalabru (changes requested):
That's useful thanks @MicaiahReid !

lgalabru (approved):

PR #582, opened 2026-03-19
fix(core): make 'until' param exclusive for `getSignaturesForAddress`

MicaiahReid:
Fixes #576

### DIFF
--- a/crates/core/src/surfnet/locker.rs
+++ b/crates/core/src/surfnet/locker.rs
@@ -804,12 +804,6 @@ impl SurfnetSvmLocker {
 
             let limit = limit.unwrap_or(1000);
 
-            let config_before = &before;
-            let config_until = &until;
-
-            let mut before_slot = None;
-            let mut until_slot = None;
-
             let sigs: Vec<_> = svm_reader
                 .transactions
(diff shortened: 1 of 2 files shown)


crates/core/src/surfnet/locker.rs
```diff
@@ -879,32 +865,41 @@ impl SurfnetSvmLocker {
...
                     b.slot.cmp(&a.slot).then_with(|| {
                         let a_pos = sig_position.get(&a.signature).unwrap_or(&usize::MAX);
                         let b_pos = sig_position.get(&b.signature).unwrap_or(&usize::MAX);
                         b_pos.cmp(&a_pos)
                     })
                 })
-                .take(limit)
-                .collect()
+                .collect();
+
```
lgalabru, line 893: Can we please have some comments describing this filtering spec?

lgalabru (approved):

crates/core/src/surfnet/locker.rs
MicaiahReid: Done 9f3827af93e69a273538dbc8b298efff6a5de74a

PR #604, opened 2026-03-31
chore: bump litesvm to v0.11.0

MicaiahReid:
Fixes #578 
 
The new LiteSVM release adds a fix to automatically create feature accounts for activated features. This allows us to remove some hacky code paths where we were previously inserting these accounts manually.

This LiteSVM change also had some downstream changes - the `with_feature_set` initializer combined with the `with_feature_accounts` initializer makes the features active and creates the accounts. Then, later if we were to deactivate a feature, the feature account would still exist and appear active in some cases. This is normal behavior for the VM, but we were instantiating things in a weird order - first with all features, then stripping it down to mainnet features. This was causing some issues. So I've refactored how we initialize the VM - not setting feature accounts until we've got the actual list of feature sets ready.

Finally, the LiteSVM release had some fixes around rent handling, specifically requiring the accurate minimum rent. This broke a bunch of our tests that were not leaving accounts with minimum rent. Those tests are fixed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4779,13 +4779,10 @@ dependencies = [
  "agave-syscalls",
  "ansi_term",
  "bincode",
- "hex",
  "indexmap 2.13.0",
  "itertools 0.14.0",
  "log 0.4.29",
- "qualifier_attr",
  "serde",
- "sha2 0.10.9",
  "solana-account 3.3.0",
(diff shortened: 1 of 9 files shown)


lgalabru (approved):
Looking great, thanks @MicaiahReid!

PR #605, opened 2026-03-31
Fix: Fetch StakeHistory sysvar and Stake program from mainnet

aoikurokawa:
### Description

- Fetch `StakeHistory` sysvar from remote RPC during init/reset — the Stake program accesses it via `get_sysvar` syscall (not as a transaction account), so lazy fetching never picks it up. Without this, `Split` panics with `SBF program Panicked in src/sysvar/stake_history.rs`
- Replace LiteSVM's bundled Stake program (`core_bpf_stake-1.0.1.so`) with the live mainnet version — the bundled ELF was compiled against `solana-vote-interface` v2.x which only knows `VoteStateVersions` variants 0-2, but mainnet vote accounts now use V4 (discriminant 3), causing `DelegateStake` to fail with `InvalidAccountData`
- Preserve all fetched accounts (StakeHistory, StakeConfig, Stake program + programdata) across garbage collection and network resets
- Add `get_raw_account` method to `SurfnetRemoteClient` for fetching accounts without token/program classification

### DIFF
--- a/crates/core/src/surfnet/locker.rs
+++ b/crates/core/src/surfnet/locker.rs
@@ -236,18 +236,59 @@ impl SurfnetSvmLocker {
         };
         epoch_info.transaction_count = None;
 
-        // Fetch StakeHistory sysvar from remote if available. Native programs access
-        // this via get_sysvar syscall, not as a transaction account, so Surfpool must
-        // proactively populate it.
-        let stake_history_account = if let Some(remote_client) = remote_ctx {
-            remote_client
-                .get_raw_account(&solana_sdk_ids::sysvar::stake_history::id())
-                .await
-                .ok()
-                .flatten()
-        } else {
-            None
-        };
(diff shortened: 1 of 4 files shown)


MicaiahReid (changes requested):
It seems like a lot of this should be fixed at the LiteSVM level:
 - the `StakeConfig` sysvar could be loaded during the `set_sysvars` call like the other sysvars are
 - the stake program could be loaded with the other default programs

This would remove a lot of the complexity here.

Could you also add an integration test that is reproducing the issue you're seeing, so we can verify this fixes the issue? With the integration test in place, we can easily try the litesvm fixes to make sure they resolve surfpool upstream

MicaiahReid:
Hey @aoikurokawa, any thoughts from my comments?

aoikurokawa:
Sorry for my slow response, i will take a look at ListSVM

PR #608, opened 2026-04-02
feat(rpc): Adds getBundleStatuses rpc method

0xzrf:
## Summery
This PR adds the `getBundleStatuses` rpc method in the `jito.rs` file to get statuses of the transactions submitted via `sendBundle` rpc method

## Changes
* Adds `jito_bundles: Box<dyn Storage<String, Vec<String>>>` to `SurfnetSvm` to persist jito bundles based on bundle_id
* Adds `get_bundle` and `process_bundle` helper methods to inside `locker.rs`
* Adds `SimnetCommand::SendBundle` to send the bundle to runloop via `simet_command_tx`
* Calls `SimnetCommand::SendBundle` inside `sendBundle` rpc call
* Adds `get_bundle_statuses` method in`SurfpoolJitoRpc` which takes a `bundle_id: String`, looks it up in via `get_bundle` helper, and then calls `SurfpoolFullRpc::get_signature_statuses` and sends back status of each transaction in the same order as originally sent in sendBundle

Closes #595

### DIFF
--- a/crates/core/src/rpc/jito.rs
+++ b/crates/core/src/rpc/jito.rs
@@ -1,10 +1,12 @@
 use std::str::FromStr;
 
-use jsonrpc_core::{Error, Result};
+use jsonrpc_core::{BoxFuture, Error, Result};
 use jsonrpc_derive::rpc;
 use sha2::{Digest, Sha256};
 use solana_client::{rpc_config::RpcSendTransactionConfig, rpc_custom_error::RpcCustomError};
+use solana_rpc_client_api::response::Response as RpcResponse;
 use solana_signature::Signature;
+use solana_transaction_status::TransactionStatus;
 
 use super::{
     RunloopContext,
@@ -59,6 +61,47 @@ pub trait Jito {
         transactions: Vec<String>,
         config: Option<RpcSendTransactionConfig>,
     ) -> Result<String>;
+
+    /// Retrieves the statuses of all transactions in a previously submitted bundle.
+    ///
+    /// This RPC method looks up a bundle by its `bundle_id` (the SHA-256 hash returned by
+    /// [`sendBundle`](#method.send_bundle)) and returns the signature statuses for the bundle's
+    /// transactions in the same order they were recorded.
+    ///
+    /// ## Parameters
+    /// - `bundle_id`: The bundle identifier returned by `sendBundle`.
+    ///
+    /// ## Returns
+    /// A contextualized response containing:
+    /// - `value`: A list of optional transaction statuses corresponding to the bundle signatures.
+    ///   Each entry can be:
+    ///   - `null` if the signature is unknown or not sufficiently confirmed for status reporting
+    ///   - a `TransactionStatus` object if the transaction is found and its status can be returned
+    ///
+    /// ## Example Request (JSON-RPC)
+    /// ```json
+    /// {
+    ///   "jsonrpc": "2.0",
+    ///   "id": 1,
+    ///   "method": "getBundleStatuses",
+    ///   "params": [
+    ///     "bundleIdHere"
+    ///   ]
+    /// }
+    /// ```
+    ///
+    /// ## Notes
+    /// - Bundles are stored locally as a mapping from `bundle_id` to a list of base-58 signatures.
+    /// - If the bundle ID is not known locally, an error is returned.
+    /// - Status resolution is delegated to the same logic used by `getSignatureStatuses`:
+    ///   statuses are computed from locally stored transactions (and may fall back to a remote
+    ///   datasource, if configured).
+    #[rpc(meta, name = "getBundleStatuses")]
+    fn get_bundle_statuses(
+        &self,
+        meta: Self::Metadata,
+        bundle_id: String,
+    ) -> BoxFuture<Result<RpcResponse<Vec<Option<TransactionStatus>>>>>;
 }
 
 #[derive(Clone)]
@@ -83,7 +126,7 @@ impl Jito for SurfpoolJitoRpc {
             )));
         }
 
-        let Some(_ctx) = &meta else {
+        let Some(ctx) = &meta else {
             return Err(RpcCustomError::NodeUnhealthy {
                 num_slots_behind: None,
             }
@@ -102,7 +145,7 @@ impl Jito for SurfpoolJitoRpc {
             let bundle_config = Some(SurfpoolRpcSendTransactionConfig {
                 base: RpcSendTransactionConfig {
                     skip_preflight: true,
-                    ..base_config.clone()
+                    ..base_config
                 },
                 skip_sig_verify: None,
             });
@@ -136,7 +179,40 @@ impl Jito for SurfpoolJitoRpc {
         let mut hasher = Sha256::new();
         hasher.update(concatenated_signatures.as_bytes());
         let bundle_id = hasher.finalize();
-        Ok(hex::encode(bundle_id))
+        let bundle_id = hex::encode(bundle_id);
+
+        let _ = ctx
+            .simnet_commands_tx
+            .send(surfpool_types::SimnetCommand::SendBundle((
+                bundle_id.clone(),
+                bundle_signatures
+                    .iter()
+                    .map(|sig| sig.to_string())
+                    .collect(),
+            )));
+
+        Ok(bundle_id)
+    }
+
+    fn get_bundle_statuses(
+        &self,
+        meta: Self::Metadata,
+        bundle_id: String,
+    ) -> BoxFuture<Result<RpcResponse<Vec<Option<TransactionStatus>>>>> {
+        Box::pin(async move {
+            let Some(ctx) = &meta else {
+                return Err(RpcCustomError::NodeUnhealthy {
+                    num_slots_behind: None,
+                }
+                .into());
+            };
+
+            let signatures = ctx.svm_locker.get_bundle(bundle_id)?;
+
+            SurfpoolFullRpc
+                .get_signature_statuses(meta.clone(), signatures, None)
+                .await
+        })
     }
 }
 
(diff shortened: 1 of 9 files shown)


crates/core/src/rpc/jito.rs
```diff
@@ -59,6 +61,47 @@ pub trait Jito {
...
+    /// - If the bundle ID is not known locally, an error is returned.
+    /// - Status resolution is delegated to the same logic used by `getSignatureStatuses`:
+    ///   statuses are computed from locally stored transactions (and may fall back to a remote
+    ///   datasource, if configured).
+    #[rpc(meta, name = "getBundleStatuses")]
+    fn get_bundle_statuses(
+        &self,
+        meta: Self::Metadata,
+        bundle_id: String,
+    ) -> BoxFuture<Result<RpcResponse<Vec<Option<TransactionStatus>>>>>;
```
MicaiahReid: It looks like the expected response per the [jito docs](https://docs.jito.wtf/lowlatencytxnsend/#id6) is a bit different than this `TransactionStatus`

crates/core/src/rpc/jito.rs
```diff
@@ -136,7 +179,40 @@ impl Jito for SurfpoolJitoRpc {
         let mut hasher = Sha256::new();
         hasher.update(concatenated_signatures.as_bytes());
         let bundle_id = hasher.finalize();
-        Ok(hex::encode(bundle_id))
+        let bundle_id = hex::encode(bundle_id);
+
+        let _ = ctx
```
MicaiahReid: We should be able to just `ctx.svm_locker.process_bundle` here - is there a reason to send this over simnet command instead of directly writing?

crates/core/src/surfnet/locker.rs
```diff
@@ -980,6 +980,18 @@ impl SurfnetSvmLocker {
         }
     }
 
+    pub fn get_bundle(&self, bundle_id: String) -> SurfpoolResult<Vec<String>> {
```
MicaiahReid: Can we have the bundle related helpers next to each other in this file? And can we get some RS docs for these helpers?

crates/core/src/surfnet/locker.rs
```diff
@@ -1108,6 +1120,11 @@ impl SurfnetSvmLocker {
         Ok(())
     }
 
+    pub fn process_bundle(&self, bundle_id: String, signatures: Vec<String>) -> SurfpoolResult<()> {
```
MicaiahReid: I think `store_bundle` seems like a better name here

MicaiahReid (changes requested):
Can we get this PR rebased on main? Currently it's including commits from another PR (sigterm one) as well

crates/core/src/rpc/jito.rs
0xzrf: I thought that processing bundle would be a blocking change which changes the svm state(Like `SurfpoolFullRpc::send_transaction()`, but the changes turned out to be way smaller in the end

WIll revert that

crates/core/src/rpc/jito.rs
```diff
@@ -59,11 +63,61 @@ pub trait Jito {
...
+    ///
+    /// ## Parameters
+    /// - `bundle_id`: The bundle identifier returned by `sendBundle`.
+    ///
+    /// ## Returns
+    /// A contextualized response containing:
+    /// - `value`: A list of optional transaction statuses corresponding to the bundle signatures.
+    ///   Each entry can be:
+    ///   - `null` if the signature is unknown or not sufficiently confirmed for status reporting
+    ///   - a `TransactionStatus` object if the transaction is found and its status can be returned
```
MicaiahReid: These doc comments need to be updated with the latest return type!

crates/core/src/rpc/jito.rs
```diff
@@ -59,11 +63,61 @@ pub trait Jito {
...
 
+// #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+// pub struct JitoBundleStatus {
+//     pub bundle_id: String,
+//     pub transactions: Vec<String>,
+//     pub slot: u64,
+//     pub confirmation_status: TransactionConfirmationStatus,
+//     pub err: std::result::Result<(), TransactionError>,
+// }
+
```
MicaiahReid: Can we remove?

crates/core/src/rpc/jito.rs
```diff
@@ -136,7 +190,89 @@ impl Jito for SurfpoolJitoRpc {
...
+                if let Some(cs) = status.confirmation_status.clone() {
+                    let should_replace = confirmation_status
+                        .as_ref()
+                        .map(|existing| confirmation_rank(&cs) > confirmation_rank(existing))
+                        .unwrap_or(true);
+                    if should_replace {
+                        confirmation_status = Some(cs);
+                    }
+                }
```
MicaiahReid: I believe if we've processed a bundle correctly, they should all land on the same block, so this logic is reduncant, right?

crates/core/src/surfnet/locker.rs
```diff
@@ -984,6 +984,25 @@ impl SurfnetSvmLocker {
...
     }
 
+    /// Stores a bundle's signatures under the given bundle ID.
+    pub fn store_bundle(&self, bundle_id: String, signatures: Vec<String>) -> SurfpoolResult<()> {
+        self.with_svm_writer(|svm_writer| svm_writer.jito_bundles.store(bundle_id, signatures))?;
+        Ok(())
+    }
+
+    /// Retrieves the list of transaction signatures for a previously stored bundle.
+    pub fn get_bundle(&self, bundle_id: String) -> SurfpoolResult<Vec<String>> {
```
MicaiahReid: Here we're returning an error if the bundle ID isn't found. According to the Jito docs, we should just be returning `null` if the bundle isn't found. It's not necessarily that the bundle id is invalid, just that there are no sigs for that bundle id

crates/types/src/types.rs
```diff
@@ -971,6 +971,15 @@ pub enum DataIndexingCommand {
     ProcessCollectionEntriesPack(Uuid, Vec<u8>),
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct JitoBundleStatus {
```
MicaiahReid, line 976: We'll want to add a serde camelCase directive here to make sure `confirmation_status` is serialized as `confirmationStatus`

MicaiahReid (changes requested):

crates/core/src/rpc/jito.rs
MicaiahReid, line 277: Ha, I think if I'm reading this loop correctly we're iterating over every status just to get the data from the first one (and the first one with an error). There's gotta be a more efficient way to do this, right?

crates/core/src/surfnet/locker.rs
```diff
@@ -984,6 +984,23 @@ impl SurfnetSvmLocker {
...
+        self.with_svm_writer(|svm_writer| svm_writer.jito_bundles.store(bundle_id, signatures))?;
+        Ok(())
+    }
+
+    /// Retrieves the list of transaction signatures for a previously stored bundle.
+    ///
+    /// Returns `None` when there is no local entry for `bundle_id` (or the backing store read fails).
+    /// This is not an "invalid id" signal: callers such as Jito `getBundleStatuses` treat `None` as
+    /// “no data” and return a null RPC result rather than an error.
+    pub fn get_bundle(&self, bundle_id: String) -> Option<Vec<String>> {
```
MicaiahReid: This function receives as `String`, then we pass in `&str` to the getter function. The callers of this `get_bundle` function are doing `bundle_id.clone()` to keep a `String`. To avoid clones, can we just pass `&str` here?

MicaiahReid (changes requested):
Good stuff, almost there!

crates/core/src/rpc/jito.rs
0xzrf: Ahh, gpt messed up, Sorry about that. The rpc method will return an error if any of the transactions were to fail. So there's prob. no need for this

Will update it on the upcoming commit

crates/core/src/surfnet/locker.rs
0xzrf: got it

crates/core/src/rpc/jito.rs
MicaiahReid: Still waiting on a commit for this

0xzrf: I'm not sure that I'm following.
I'm using `drain_one_process_transaction` on mempool  since not using the `mempool_rx` drops it early, giving a receiver error in `send_transaction()`:
```rust
ctx.simnet_commands_tx
            .send(SimnetCommand::ProcessTransaction(
                ctx.id,
                unsanitized_tx,
                status_update_tx,
                config.base.skip_preflight,
                config.skip_sig_verify,
            ))
            .map_err(|_| RpcCustomError::NodeUnhealthy {
                num_slots_behind: None,
            })?;
```

Or is there something missing here?

MicaiahReid:
```rust
for status in statuses.value.iter().flatten() {
    if !took_bundle_fields {
        slot = status.slot;
        confirmation_status = status.confirmation_status.clone();
        took_bundle_fields = true;
    }

    if first_err.is_none()
        && let Some(err) = status.err.clone()
    {
        first_err = Some(err);
    }
}
```
So in this loop, the goal is to:
 - get the slot/confirmation status from the first item
 - get the first non-None error

However, this loops over every status.

Say statuses is
```
[
  { slot: 1, status: processed, error: Some("err") }
  { slot: 1, status: processed, error: None }
  { slot: 1, status: processed, error: None }
  { slot: 1, status: processed, error: None }
  { slot: 1, status: processed, error: None }
]
```
In the first iteration of the loop, we have all of the information that we need, but we're still iterating over the rest for no reason. Is there a more efficient way we could be extracting the data we need?

MicaiahReid:
Hey @0xzrf, I was trying to clean up a few things and rebase, and things got messy. I ended up taking your commits and adding them to a different PR. Closing in favor of #660. Thanks for the great work!

PR #610, opened 2026-04-03
fix(jito_rpc): makes jito bundle atomic

0xzrf:
## Summery
Simulates each transaction sent in the `sendBundle` rpc method and early returns if any of them fail.

## Changes
* Adds an iterator inside the `SurfpoolJitoRpc::send_bundle` function before sending transactions
* Adds test to check if it returns the right error message

## Note
This might not entirely replicate jito bundles. But since it's majorly used for testing purposes, it works well to say that the bundle is `atomic`

Closes #594

### DIFF
--- a/crates/core/src/rpc/jito.rs
+++ b/crates/core/src/rpc/jito.rs
@@ -3,10 +3,7 @@ use std::str::FromStr;
 use jsonrpc_core::{Error, Result};
 use jsonrpc_derive::rpc;
 use sha2::{Digest, Sha256};
-use solana_client::{
-    rpc_config::{RpcSendTransactionConfig, RpcSimulateTransactionConfig},
-    rpc_custom_error::RpcCustomError,
-};
+use solana_client::{rpc_config::RpcSendTransactionConfig, rpc_custom_error::RpcCustomError};
 use solana_signature::Signature;
 
 use super::{
@@ -101,13 +98,7 @@ impl Jito for SurfpoolJitoRpc {
         // pre-flight simulate each transaction and bail out early if any of them fails.
         for (idx, txn) in transactions.iter().enumerate() {
             let simulate_result = match hiro_system_kit::nestable_block_on(
-                full_rpc.simulate_transaction(
-                    meta.clone(),
-                    txn.clone(),
-                    Some(RpcSimulateTransactionConfig {
-                        ..Default::default()
-                    }),
-                ),
+                full_rpc.simulate_transaction(meta.clone(), txn.clone(), None),
             ) {
                 Ok(res) => res,
                 Err(e) => {
--- a/crates/core/src/rpc/jito.rs
+++ b/crates/core/src/rpc/jito.rs
@@ -1,15 +1,17 @@
-use std::str::FromStr;
-
 use jsonrpc_core::{Error, Result};
 use jsonrpc_derive::rpc;
 use sha2::{Digest, Sha256};
 use solana_client::{rpc_config::RpcSendTransactionConfig, rpc_custom_error::RpcCustomError};
 use solana_signature::Signature;
+use solana_transaction::versioned::VersionedTransaction;
+use solana_transaction_status::UiTransactionEncoding;
 
 use super::{
     RunloopContext,
     full::{Full, SurfpoolFullRpc, SurfpoolRpcSendTransactionConfig},
+    utils::decode_and_deserialize,
 };
+use crate::surfnet::locker::SurfnetSvmLocker;
 
 /// Maximum number of transactions allowed in a single bundle, matching Jito's limit.
 const MAX_BUNDLE_SIZE: usize = 5;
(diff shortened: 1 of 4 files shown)


MicaiahReid (changes requested):
Sadly this won't cut it. A jito bundle is likely to be a set of transactions that build off each other. For example:
1. Transaction 1 - Create token account, fund it
2. Transaction 2 - Put token from token account into escrow

In this case, since we're just simulating one after the other, tx 1 would succeed for simulation, and tx 2 would fail, because it depends on tx 1 being in affect.

Even a diferent loop like:
 1. Simulate tx 1
 2. Execute tx 1
 3. Simulate tx 2
 4. Execute tx 2
 5. etc...

Wouldn't work, because if there's a third tx that will always fail, txs 1 + 2 have made changes to the VM while tx 3 fails, so this is no longer atomic:

 1. Simulate tx 1
 2. Execute tx 1 (changes VM)
 3. Simulate tx 2
 4. Execute tx 2 (changes VM)
 5. Simulate tx 3 -> fails, return error

This case shows that even though a transaction in the bundle fails, some of the transactions still make changes to the vm, breaking atomicity.

---

I don't know how much of the plumbing is in place to do this from the jito RPC methods, but take a look at how we're executing the instruction profiles.

We essentially clone the whole VM state so we have a disposable sandbox where we can make changes to the cloned VM without changing the real VM.

We execute all transactions against the cloned VM, and if none of them fail, we can be confident that the transactions won't fail on the source VM.

I think we'll have to do the same here

0xzrf:
Hey @MicaiahReid 
Added the disposable vm as suggested, and added some tests for the same

I'm wondering what should happen with the real svm state?
1. Should we execute txns as usual(Time consuming, can miss the block)
2. Just Copy the account state from the vm to the real svm

What're your thoughts on it?

crates/core/src/rpc/jito.rs
```diff
@@ -83,21 +85,100 @@ impl Jito for SurfpoolJitoRpc {
...
+                    idx + 1
+                )));
+            }
+
+            // `process_transaction` can return `Ok(())` even when the tx failed, because the
+            // failure is communicated through the status channel. Treat any non-success status
+            // as a bundle failure.
+            match status_rx.recv_timeout(std::time::Duration::from_secs(2)) {
+                Ok(surfpool_types::TransactionStatusEvent::Success(_)) => {}
+                Ok(other) => {
```
MicaiahReid: Let's match on all other options here - one will be simulation error, one will be execution error

crates/core/src/rpc/jito.rs
```diff
@@ -83,21 +85,100 @@ impl Jito for SurfpoolJitoRpc {
...
+                    return Err(Error::invalid_params(format!(
+                        "Jito bundle couldn't be sent as it's not atomic: simulation failed for transaction {}: {:?}",
```
MicaiahReid:
```suggestion
                        "Jito bundle couldn't be executed: simulation failed for transaction {}: {:?}",
```

crates/core/src/rpc/jito.rs
```diff
@@ -83,21 +85,100 @@ impl Jito for SurfpoolJitoRpc {
...
+                        idx + 1,
+                        other
+                    )));
+                }
+                Err(e) => {
+                    return Err(Error::invalid_params(format!(
+                        "Jito bundle couldn't be sent as it's not atomic: simulation RPC failed for transaction {}: {e}",
```
MicaiahReid: I think that if we have an error here, it's because the `status_tx` was dropped, which would mean something has really gone wrong. I believe
```
return Err(RpcCustomError::NodeUnhealthy {
                num_slots_behind: None,
            }
            .into());
```
would be more appropriate

crates/core/src/rpc/jito.rs
```diff
@@ -83,21 +85,100 @@ impl Jito for SurfpoolJitoRpc {
...
+                    skip_preflight,
+                    sigverify,
+                ));
+
+            let signature = tx.signatures[0];
+            bundle_signatures.push(signature);
+
+            if let Err(e) = process_res {
+                return Err(Error::invalid_params(format!(
+                    "Jito bundle couldn't be sent as it's not atomic {}: {e}",
```
MicaiahReid:
```suggestion
                    "Jito bundle couldn't be executed, failed to process transaction {}: {e}",
```

MicaiahReid (changes requested):
I love the idea of taking the cloned VM and applying the changes onto the real VM, but I don't think it'll work here.

When the vm is cloned for profiling, we're making sure that geyser events + account WS updates aren't being propagated. So we need to be sure to actually execute the changes against the real VM so those updates get sent

crates/core/src/rpc/jito.rs
```diff
@@ -421,4 +500,187 @@ mod tests {
...
         );
     }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_send_bundle_dependent_transaction_failure_aborts_entire_bundle() {
+        let payer = Keypair::new();
+        let recipient = Keypair::new();
+
+        // Use mempool-backed setup so we can assert that a sandbox failure does NOT enqueue any
+        // ProcessTransaction commands (phase 2 commit currently still uses sendTransaction).
```
MicaiahReid:
```suggestion
        // ProcessTransaction commands.
```

crates/core/src/rpc/jito.rs
```diff
@@ -421,4 +500,187 @@ mod tests {
...
+                                    mutated_accounts,
+                                ),
+                            );
+
+                            let _ = status_tx.send(TransactionStatusEvent::Success(
+                                TransactionConfirmationStatus::Confirmed,
+                            ));
+                        }
+                        SimnetCommand::AirdropProcessed => continue,
+                        _ => continue,
```
MicaiahReid:
```suggestion
                        _ => continue,
```

crates/core/src/rpc/jito.rs
```diff
@@ -421,4 +500,187 @@ mod tests {
...
+        stop_drain.store(true, Ordering::SeqCst);
+        let _ = drain_handle.join();
+
+        // If sandbox failure happens as expected, Phase 2 should never run.
+        assert_eq!(
+            observed_process_tx.load(Ordering::SeqCst),
+            0,
+            "Expected zero mempool ProcessTransaction commands; sandbox failure should prevent Phase 2"
+        );
+    }
```
MicaiahReid, line 669: Could we also have some assertions here that tx1 didn't update the VM at all? Since tx 1 sends funds from sender to recipient, maybe we can assert that the recipient has no funds?

0xzrf:
Hey @MicaiahReid 
Could you review this PR?
Thanks.

PR #615, opened 2026-04-14
chore: various cleanups

cds-amal:
I ran [`rust-scope`](https://github.com/cds-rs/rs-scope) against `surfpool`
twice: once hunting for structural duplication, once for complexity
hotspots. Each commit addresses one finding. Collectively about 55 net
lines of DRY come off, and three of the hairier functions pick up phase
helpers. No observable behavior moves.

Commit messages carry the per-commit rationale. This doc is the reviewer's
plan: where to look closely, what I want pushed back on, and what to run.

## How to review

Any order works. If you only have bandwidth for one or two, prioritise
2.2 (hot path, widest behavior surface) and 2.1 (a behavior change
from review feedback, so worth a second look).

Sizes: XS (under a minute), S (a few minutes), M (a careful read), L
(deep dive).

Reviewed commits (unchanged since round one):

| # | Commit | What | Size |
|---|---|---|---|
| 1.1 | `eb878de` | 13-way `\|\|` chain to const needle slice | XS |
| 1.2 | `2c350dc` | Macro-generated `TokenAccount` / `MintAccount` serde | M |
| 1.3 | `2947cf8` | Shared `storage::diesel_common` for SQLite/Postgres | S |
| 2.1 | `d262afe` | `write_and_execute_iac`: three booleans to `RunbookExecutionMode` enum | S |
| 2.2 | `54f6c81` | `update_account_registries` + `remove_from_indexes` phase split | L |
| 2.3 | `080c229` | `simulate_transaction` + `send_transaction` share decode preamble | XS |
|     | `fc3c2d6` | (`docs(cli):` truth table on `from_inputs`; reverted by `1e6aaf1`, see below) | XS |

Follow-ups from round-one review (@MicaiahReid), additive on top of the
reviewed history:

| Commit | What | Size |
|---|---|---|
| `1e6aaf1` | Revert `fc3c2d6` (the maintainer judged the truth table not worth its review cost) | XS |
| `df723fd` | Strip past-state prose from enum doc and `diesel_common` module doc | XS |
| `3e73199` | Collapse `Neither` into `ExistingOnDisk` (behavior change) | S |
| `99fd3e4` | Inline `sync_account_in_db` and `mark_account_offline_and_unindex` | S |

**1.1 (needle slice, XS).** Check that `METHOD_NOT_SUPPORTED_NEEDLES` matches
the original `||` chain at `rpc/utils.rs:222` in both contents and order.
Nothing else has changed.

**1.2 (Token/Mint macro, M).** The macro
`impl_token_program_packable_serde!` at `crates/core/src/types.rs:860`
gets called twice, once for each enum. Verify the substitution by
reading the macro definition and one of the call sites side-by-side: the
four macro parameters (enum name, 2022 inner type, legacy inner type,
error-label string) should end up in obvious places in the generated
impl. If you have `cargo-expand` installed, `cargo expand -p
surfpool-core --lib types 2>/dev/null | sed -n '/impl Serialize for
TokenAccount/,/^}/p'` prints the actual expansion so you can diff it
against `git show 2c350dc^:crates/core/src/types.rs` without squinting.
The macro uses `concat!` to splice error labels at compile time;
confirm the error messages stay byte-identical to the originals.
*Push back on:* the lack of a round-trip test. If you'd rather not merge
without one, I'll add it as an extra commit.

**1.3 (storage helpers, S).** Confirm every deleted helper in `sqlite.rs` /
`postgres.rs` has a corresponding import from `diesel_common`, and the
per-file `NAME` const still reaches every error constructor. *Push back
on:* whether to push further (a `SqlDialect` trait or a SQL-template
macro). I stopped at the dialect-free boundary on purpose, but if the
team wants the next step, it's a reasonable follow-up. *Ask:* if you have
a live Postgres handy, please run
`DATABASE_URL=... cargo test -p surfpool-core --features postgres
storage::postgres`. I couldn't.

**2.1 (mode enum + `3e73199`, S).** Read commits `d262afe` (the
refactor) and `3e73199` (the Neither-collapse behavior change)
together. The final state dispatches `(false, _, true) =>
ExistingOnDisk`, which is a genuine behavior change: users running
`--anchor-compat` with an existing `txtx.yml` will now see that
`txtx.yml` execute instead of nothing. That's the intentional
reading per review feedback.

**2.2 (registries split + `99fd3e4`, L).** Biggest commit, on a hot
path. Review order:

1. The two free-function helpers (`add_pubkey_to_index`,
   `remove_pubkey_from_index`) near the top of `svm.rs`. They encode
   two invariants: dedup on insert, take on empty. If those are right,
   the callers are right.
2. The three phase methods that stayed
   (`index_token_account_variant`, `index_mint_account_variant`,
   `index_token_2022_mint_extensions`) against the original body at
   `git show 54f6c81^:crates/core/src/surfnet/svm.rs:1172`. Ordering
   matters: DB mirror first, then offline-marking-with-early-return on
   delete, then remove stale indexes, then add to new indexes.
3. The trickiest line:
   `if is_supported_token_program(&old_account.owner) && let Some(old_token_account) = self.token_accounts.take(&pubkey_str)?`
   uses `let chain` to avoid re-indenting the original nested `if let`.
   Short-circuit semantics are preserved (if the predicate is false,
   `take` does not run).

`99fd3e4` inlined the two shortest phase methods back per review
feedback. The three remaining phase methods are named by what they
decode and index, so adding another token-related step to any of them
doesn't need a rename. `add_pubkey_to_index` / `remove_pubkey_from_index`
also stay: a single rename couldn't serve all four call sites, which
each carry different semantic meaning (owner, token-owner, token-mint,
token-delegate).

Behavior check: `cargo test -p surfpool-core --lib
surfnet::svm::tests::test_inserting_account_updates` (3 backends).

**2.3 (decode extraction, XS).** The helper is 10 lines in `rpc/utils.rs`.
Verify the sync and async call sites both preserve the error path
end-to-end (same message, same error code). *Push back on:* the name
`decode_rpc_versioned_transaction` if it feels off.

## Tests I ran

```bash
cargo check --workspace --all-targets                               # clean
cargo test -p surfpool-core --lib storage::sqlite                   # 4/4
cargo test -p surfpool-core --lib \
  surfnet::svm::tests::test_inserting_account_updates               # 3/3
cargo test -p surfpool-core --lib rpc::                             # 136/136
```

Postgres integration tests: not run (need a live server). `postgres.rs` is
structurally symmetric with `sqlite.rs` post-refactor, so the same tests
should continue to pass. Hence the ask in 1.3.

### DIFF
--- a/crates/cli/src/cli/simnet/mod.rs
+++ b/crates/cli/src/cli/simnet/mod.rs
@@ -522,9 +522,6 @@ enum RunbookExecutionMode {
     /// `bin_path` injection) or `--anchor-compat` without an existing
     /// `txtx.yml`. Requires framework detection.
     InMemory,
-    /// `--anchor-compat` set AND `txtx.yml` exists: the original code fell
-    /// through without setting either data pointer. Preserved as-is.
-    Neither,
 }
 
 impl RunbookExecutionMode {
@@ -533,9 +530,9 @@ impl RunbookExecutionMode {
             // `--artifacts-path` always wins: the custom bin_path has to be
             // injected, which only the in-memory runbook can do.
             (true, _, _) => Self::InMemory,
+            // An on-disk `txtx.yml` is authoritative regardless of `--anchor-compat`.
+            (false, _, true) => Self::ExistingOnDisk,
             (false, true, false) => Self::InMemory,
-            (false, true, true) => Self::Neither,
-            (false, false, true) => Self::ExistingOnDisk,
             (false, false, false) => Self::ScaffoldOnDisk,
         }
     }
@@ -622,7 +619,7 @@ async fn write_and_execute_iac(
                         cmd.artifacts_path.as_deref(),
                     )?);
                 }
-                RunbookExecutionMode::ExistingOnDisk | RunbookExecutionMode::Neither => {}
+                RunbookExecutionMode::ExistingOnDisk => {}
             }
         }
 
--- a/crates/cli/src/cli/simnet/mod.rs
+++ b/crates/cli/src/cli/simnet/mod.rs
@@ -506,15 +506,11 @@ fn log_events(
 
 /// How a simnet session decides to execute its deployment runbooks.
 ///
-/// The previous implementation encoded this as three independent booleans
-/// (`has_custom_artifacts_path`, `cmd.anchor_compat`, `txtx_manifest_exists`)
-/// combined into ad-hoc predicates at the call sites. The truth table below
-/// lifts those predicates into one place so every decision downstream is a
-/// single `match` instead of a compound boolean. Semantics are byte-identical
-/// to the pre-refactor version, including the `Neither` fallthrough that the
-/// original logic produced for the anchor-compat-with-existing-txtx case
-/// (preserved deliberately; whether that case is a bug is a separate
-/// question).
+/// Classifies the three runtime inputs that steer execution
+/// (`cmd.artifacts_path.is_some()`, `cmd.anchor_compat`, whether a `txtx.yml`
+/// exists at the simnet's base location) into a single variant. Downstream
+/// callers `match` on the variant instead of recomputing compound boolean
+/// predicates at each decision site.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 enum RunbookExecutionMode {
     /// A `txtx.yml` already exists on disk: execute it as-is.
--- a/crates/cli/src/cli/simnet/mod.rs
+++ b/crates/cli/src/cli/simnet/mod.rs
@@ -532,45 +532,6 @@ enum RunbookExecutionMode {
 }
 
 impl RunbookExecutionMode {
-    /// Truth table proving equivalence with the pre-refactor boolean form.
-    ///
-    /// Predicates:
-    /// - `H` = `has_custom_artifacts_path` (i.e. `cmd.artifacts_path.is_some()`)
-    /// - `A` = `anchor_compat`
-    /// - `T` = `txtx_exists`
-    ///
-    /// Pre-refactor action tuple, derived from the three compound booleans
-    /// that used to live at lines ~528, ~531, and ~571 of this function
-    /// (see `git show d262afe^`):
-    /// - `on_disk_pre` = `!H && !A && T`          (on-disk pointer set before framework detection)
-    /// - `do_scaffold` = `!H && !A && !T`         (scaffold-on-disk after framework detection; also sets the on-disk pointer)
-    /// - `do_in_mem`   = `H || (A && !T)`         (in-memory runbook after framework detection)
-    ///
-    /// Post-refactor action tuple, driven by the variant this function returns
-    /// and the downstream `match mode`:
-    /// - `ExistingOnDisk`:  `(on_disk_pre=1, do_scaffold=0, do_in_mem=0)`
-    /// - `ScaffoldOnDisk`:  `(on_disk_pre=0, do_scaffold=1, do_in_mem=0)`  (note: `do_scaffold` also writes the on-disk pointer, separately from `on_disk_pre`)
-    /// - `InMemory`:        `(on_disk_pre=0, do_scaffold=0, do_in_mem=1)`
-    /// - `Neither`:         `(on_disk_pre=0, do_scaffold=0, do_in_mem=0)`
-    ///
-    /// Row-by-row equivalence:
-    ///
-    /// | H | A | T | pre-refactor (on_disk_pre, do_scaffold, do_in_mem) | variant          | post-refactor tuple |
-    /// |---|---|---|-----------------------------------------------------|------------------|---------------------|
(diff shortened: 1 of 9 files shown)


crates/cli/src/cli/simnet/mod.rs
```diff
@@ -504,6 +504,47 @@ fn log_events(
     Ok(())
 }
 
+/// How a simnet session decides to execute its deployment runbooks.
+///
+/// The previous implementation encoded this as three independent booleans
+/// (`has_custom_artifacts_path`, `cmd.anchor_compat`, `txtx_manifest_exists`)
+/// combined into ad-hoc predicates at the call sites. The truth table below
```
MicaiahReid: I don't think the comment talking about past state is necessary here, just talking about the current implementation is sufficient

crates/cli/src/cli/simnet/mod.rs
```diff
@@ -504,6 +504,47 @@ fn log_events(
...
+    ExistingOnDisk,
+    /// No `txtx.yml`: scaffold one on disk from the detected framework and
+    /// execute it. Requires framework detection to take effect.
+    ScaffoldOnDisk,
+    /// Execute an in-memory runbook. Triggered by `--artifacts-path` (custom
+    /// `bin_path` injection) or `--anchor-compat` without an existing
+    /// `txtx.yml`. Requires framework detection.
+    InMemory,
+    /// `--anchor-compat` set AND `txtx.yml` exists: the original code fell
+    /// through without setting either data pointer. Preserved as-is.
```
MicaiahReid: Again this comment is talking about the original code rather than what causes this variant and what this variant means for what the code should do

crates/core/src/surfnet/svm.rs
```diff
@@ -1176,124 +1216,152 @@ impl SurfnetSvm {
...
-            self.offline_accounts.store(
-                pubkey.to_string(),
-                OfflineAccountConfig {
-                    include_owned_accounts: false,
-                },
-            )?;
-            if let Some(old_account) = self.get_account(pubkey)? {
-                self.remove_from_indexes(pubkey, &old_account)?;
-            }
+            self.mark_account_offline_and_unindex(pubkey)?;
```
MicaiahReid: Don't think this refactor is helpful - we're combining two pieces of logic that we need to do when an account is deleted. What happens if we need to do a third thing when account is deleted? We have to update this function name?

crates/core/src/surfnet/svm.rs
```diff
@@ -1176,124 +1216,152 @@ impl SurfnetSvm {
...
-            .accounts_by_owner
-            .get(&owner_key)
-            .ok()
-            .flatten()
-            .unwrap_or_default();
-        if !owner_accounts.contains(&pubkey_str) {
-            owner_accounts.push(pubkey_str.clone());
-            self.accounts_by_owner.store(owner_key, owner_accounts)?;
-        }
+        add_pubkey_to_index(
```
MicaiahReid, line 1253: Why do `index_token_account_variant`, etc get a nice, purty name, but this one doesn't? 🥺👉👈

crates/core/src/surfnet/svm.rs
```diff
@@ -1176,124 +1216,152 @@ impl SurfnetSvm {
...
-        // in the inner SVM. However, the database hasn't been updated yet, so we need to manually update the db.
-        if is_deleted_account {
-            // This amounts to deleting the account from the db if the account is deleted in the SVM
-            self.inner.delete_account_in_db(pubkey)?;
-        } else {
-            // Or updating the db account to match the SVM account if not deleted
-            self.inner
-                .set_account_in_db(*pubkey, account.clone().into())?;
-        }
+        self.sync_account_in_db(pubkey, account, is_deleted_account)?;
```
MicaiahReid: Don't think I love this abstraction either - it's kinda hiding the logic for what we do for deleted accounts with little benefit

crates/cli/src/cli/simnet/mod.rs
```diff
@@ -504,6 +504,47 @@ fn log_events(
...
+    Neither,
```
MicaiahReid: Let's remove the `Neither` variant - this case is doing the same as `ExistingOnDisk`, so it should just be married with that. So in the below `from_inputs` we'll have
```rust
 (false, _, true) => Self::ExistingOnDisk,
```

MicaiahReid (changes requested):

MicaiahReid (approved):
LGTM, thanks @cds-amal!

MicaiahReid:
Hey @cds-amal, can you run `cargo +nightly fmt --all -- --check` and fix linting issues?

cds-amal:
> Hey @cds-amal, can you run `cargo +nightly fmt --all -- --check` and fix linting issues?

@MicaiahReid, i applied the fmt changes. ~I assume you didn't mean `cargo clippy` 😅 , which aren't related to the changes in this PR.~

Edit: read the red ci testing thingy ..

PR #620, opened 2026-04-20
feat: add self-update command to cli

### LINKED ISSUE #623: `surfpool upgrade`: use semver check
We should probably do a semver > check rather than `==` here - if we have a pre-release published, this current logic would actually downgrade the release. (Say latest is 1.0.0, but we publish 1.1.0-beta as a pre-release. If the user has the beta installed, then runs this, they will not be equal, so user will be downgraded to 1.0.0)

_Originally posted by @MicaiahReid in https://github.com/solana-foundation/surfpool/pull/620#discussion_r3117844941_

### LINKED ISSUE #624: `surfpool update`: add confirmation
`surfpool update` immediately downloads and overwrites the running binary. A `dialoguer::Confirm` prompt ("Update 1.1.1 → 1.2.0? [y/N]") — `dialoguer` is already a dep — would prevent fat-finger updates. We can also add a `--yes` flag to skip this prompt

_Originally posted by @MicaiahReid in https://github.com/solana-foundation/surfpool/pull/620#discussion_r3117557365_

### LINKED ISSUE #625: `surfpool update`: allow passing version
It would be nice to be able to pass in `--version` and download at that version.

_Originally posted by @MicaiahReid in https://github.com/solana-foundation/surfpool/pull/620#discussion_r3117479841_

VivekJaiswal18:
Added `surfpool update` command to the cli which updates `surfpool` to the latest version.

fixes #597 
fixes #623 
fixes #624
fixes #625

### DIFF
--- a/crates/cli/src/cli/mod.rs
+++ b/crates/cli/src/cli/mod.rs
@@ -30,9 +30,10 @@ use surfpool_types::{
 use txtx_core::manifest::WorkspaceManifest;
 use txtx_gql::kit::{helpers::fs::FileLocation, types::frontend::LogLevel};
 
+use crate::cli::update::handle_update_command;
 use crate::runbook::handle_execute_runbook_command;
-
 mod simnet;
+mod update;
 
 #[derive(Clone)]
 pub struct Context {
@@ -127,6 +128,9 @@ enum Command {
     /// Start MCP server
     #[clap(name = "mcp", bin_name = "mcp")]
     Mcp,
+    /// Update Surfpool to the latest version
+    #[clap(name = "update", bin_name = "update")]
+    Update,
 }
 
 #[derive(Parser, PartialEq, Clone, Debug)]
@@ -649,6 +653,7 @@ fn handle_command(opts: Opts, ctx: &Context) -> Result<(), String> {
         }
         Command::List(cmd) => hiro_system_kit::nestable_block_on(handle_list_command(cmd, ctx)),
         Command::Mcp => hiro_system_kit::nestable_block_on(handle_mcp_command(ctx)),
+        Command::Update => hiro_system_kit::nestable_block_on(handle_update_command()),
     }
 }
 
--- a/crates/cli/src/cli/mod.rs
+++ b/crates/cli/src/cli/mod.rs
@@ -30,8 +30,7 @@ use surfpool_types::{
 use txtx_core::manifest::WorkspaceManifest;
 use txtx_gql::kit::{helpers::fs::FileLocation, types::frontend::LogLevel};
 
-use crate::cli::update::handle_update_command;
-use crate::runbook::handle_execute_runbook_command;
+use crate::{cli::update::handle_update_command, runbook::handle_execute_runbook_command};
 mod simnet;
 mod update;
 
--- a/crates/cli/src/cli/mod.rs
+++ b/crates/cli/src/cli/mod.rs
@@ -130,7 +130,7 @@ enum Command {
     Mcp,
     /// Update Surfpool to the latest version
     #[clap(name = "update", bin_name = "update")]
-    Update,
+    Update(UpdateCommand),
 }
 
 #[derive(Parser, PartialEq, Clone, Debug)]
@@ -535,6 +535,16 @@ pub struct ExecuteRunbook {
     pub log_dir: String,
 }
 
+#[derive(Parser, PartialEq, Clone, Debug)]
+pub struct UpdateCommand {
+    /// Flag to skip confirmation prompt
+    #[arg(long = "yes", short = 'y')]
+    pub skip_confirm: bool,
+    /// To update to a specific version instead of the latest
+    #[arg(long = "version", short = 'v')]
+    pub version: Option<String>,
+}
+
 impl ExecuteRunbook {
     pub fn default_localnet(runbook_name: &str) -> ExecuteRunbook {
         ExecuteRunbook {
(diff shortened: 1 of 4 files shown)


crates/cli/src/cli/update/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+    println!("Latest version: {}", latest_tag_name);
+    let users_asset = get_asset_name()?;
+    let browser_download_url = latest_version
+        .assets
+        .iter()
+        .find(|a| a.name == users_asset)
+        .map(|a| a.browser_download_url.as_str())
+        .ok_or_else(|| format!("No asset name found matching the users platform"))?;
+    println!("Current version: {}", current_version);
+    println!("Download URL: {}", browser_download_url);
```
MicaiahReid, line 45: Let's log this after the if statement below (so only log if we're actually downloading)

crates/cli/src/cli/mod.rs
```diff
@@ -127,6 +128,9 @@ enum Command {
     /// Start MCP server
     #[clap(name = "mcp", bin_name = "mcp")]
     Mcp,
+    /// Update Surfpool to the latest version
+    #[clap(name = "update", bin_name = "update")]
+    Update,
```
MicaiahReid, line 133: It would be nice to be able to pass in `--version` and download at that version.

crates/cli/src/cli/update/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+    assets: Vec<Asset>,
+}
+
+#[derive(Deserialize, Debug)]
+struct Asset {
+    name: String,
+    browser_download_url: String,
+}
+
+pub async fn handle_update_command() -> Result<(), String> {
```
MicaiahReid, line 23: `surfpool update` immediately downloads and overwrites the running binary. A `dialoguer::Confirm` prompt ("Update 1.1.1 → 1.2.0? [y/N]") — `dialoguer` is already a dep — would prevent fat-finger updates. We can also add a `--yes` flag to skip this prompt

crates/cli/src/cli/update/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+
+fn get_asset_name() -> Result<String, String> {
+    let users_os = std::env::consts::OS;
+    let users_arch = std::env::consts::ARCH;
+
+    match (users_os, users_arch) {
+        ("macos", "aarch64") => Ok("surfpool-darwin-arm64.tar.gz".into()),
+        ("macos", "x86_64") => Ok("surfpool-darwin-x64.tar.gz".into()),
+        ("linux", "x86_64") => Ok("surfpool-linux-x64.tar.gz".into()),
+        _ => Err(format!("Unsupported platform: {users_os}-{users_arch}")),
```
MicaiahReid, line 109: There is a windows build being uploaded, can we add that? And for the windows build, the filename will be "surfpool.exe" rather than just "surfpool", so you'll need to update the name check above accordingly.

crates/cli/src/cli/update/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+    let latest_tag_name = latest_version.tag_name.trim_start_matches('v');
+    let current_version: &str = env!("CARGO_PKG_VERSION");
+    println!("Latest version: {}", latest_tag_name);
+    let users_asset = get_asset_name()?;
+    let browser_download_url = latest_version
+        .assets
+        .iter()
+        .find(|a| a.name == users_asset)
+        .map(|a| a.browser_download_url.as_str())
+        .ok_or_else(|| format!("No asset name found matching the users platform"))?;
```
MicaiahReid, line 43: adding the asset that was checked for in this error message would be helpful

MicaiahReid: Also, should be `user's platform`, not `users platform`

crates/cli/src/cli/update/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+    println!("Current version: {}", current_version);
+    println!("Download URL: {}", browser_download_url);
+
+    if current_version == latest_tag_name {
```
MicaiahReid, line 47: We should probably do a semver > check rather than `==` here - if we have a pre-release published, this current logic would actually downgrade the release. (Say latest is 1.0.0, but we publish 1.1.0-beta as a pre-release. If the user has the beta installed, then runs this, they will not be equal, so user will be downgraded to 1.0.0)

MicaiahReid (changes requested):

crates/cli/src/cli/update/mod.rs
MicaiahReid: Tracking in #623

crates/cli/src/cli/update/mod.rs
MicaiahReid: Tracking in #624

crates/cli/src/cli/mod.rs
MicaiahReid: Tracking in #625

MicaiahReid:
Great work @VivekJaiswal18! I've added some comments for next steps and converted to issues. The remaining comments are bugs or cleanup. Would be great to get this in the next release today or tomorrow!

VivekJaiswal18:
@MicaiahReid please have a look!

crates/cli/src/cli/update/mod.rs
```diff
@@ -33,21 +46,55 @@ pub async fn handle_update_command() -> Result<(), String> {
...
+        return Ok(());
+    }
+
+    if !cmd.skip_confirm {
+        let theme = ColorfulTheme {
+            defaults_style: Style::new().for_stderr(),
+            ..ColorfulTheme::default()
+        };
+        println!("Current version: {}", current_version);
+        println!("Latest version: {}", latest_tag_name);
```
MicaiahReid: These seem redundant, since you're about to ask for confirmation in the next line

MicaiahReid (changes requested):
Looking good, and aplogies for the slow review! Small nit, then let's merge! 💪

MicaiahReid (approved):
Let's go, thanks @VivekJaiswal18!

PR #621, opened 2026-04-20
feat(core): simulate realistic network conditions

### LINKED ISSUE #529: Simulate realistic network conditions
Surfpool always successfully accepts and executes your transactions, which avoids distractions and is great for proving logical correctness of your program. There are a lot of complications in transaction landing that would be difficult to account for. Try again with higher priority fees? Try again with Jito bundles? Harmonic? Are you using a staked connection? But I think it would still be helpful to randomly not accept transactions (maybe the acceptance rate itself can also be configurable), so that developers can mature their "if no work, what do?" strategy.

Additionally, Surfpool executes transactions in exactly the order you submit them no matter how fast you send them. If a transaction later in the sequence depends on a transaction earlier in the sequence, if you send too quickly and don't confirm, they sometimes get out of order on mainnet. We could simulate that by inserting a random delay in the time surfpool actually executes the transactions.

These could either be one flag, or two separate flags. I can see a case for both.

AvhiMaz:
Closes #529
  - Add `--transaction-drop-rate <0.0-1.0>` flag: randomly drops transactions with the given probability, simulating
  packet loss or leader rejection
  - Add `--transaction-execution-delay-ms <ms>` flag: applies a random delay (0 to N ms) before executing each
  transaction, simulating out-of-order processing under high throughput
  - Add `TransactionStatusEvent::Dropped` variant returned to the RPC client when a transaction is dropped

 Usage:
  ```bash
  surfpool start --transaction-drop-rate 0.1 --transaction-execution-delay-ms 200

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11406,6 +11406,7 @@ dependencies = [
  "opentelemetry-prometheus",
  "opentelemetry_sdk 0.28.0",
  "prometheus",
+ "rand 0.8.5",
  "reqwest 0.12.28",
  "serde",
  "serde_derive",
--- a/crates/cli/src/cli/mod.rs
+++ b/crates/cli/src/cli/mod.rs
@@ -64,6 +64,16 @@ lazy_static::lazy_static! {
 /// We set out snap build to set this environment variable to the real home directory,
 /// because by default, snaps run in a confined environment where the home directory is not
 /// the user's actual home directory.
+fn parse_drop_rate(s: &str) -> Result<f64, String> {
+    let v: f64 = s
+        .parse()
+        .map_err(|_| format!("'{}' is not a valid number", s))?;
+    if !(0.0..=1.0).contains(&v) {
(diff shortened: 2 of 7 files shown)


AvhiMaz:
can you review this, @MicaiahReid @lgalabru ? thanks :)

MicaiahReid:
Thanks for the PR, @AvhiMaz. I've taken a look, and I don't think this PR will cut it, for a few reasons:
 - Delaying a transaction from processing as you're introducing here via transaction-execution-delay-ms just delays all transactions but keeps them in the same order
 - A transaction being dropped on mainnet won't tell a user it's dropped - this PR is too helpful/not realistic. The tx should be accepted and then just not be available later
 - To actually address #529, we'd want more sophisticated (probably ComputeBudget-based) ways of dropping txs, not just random

I'll follow up with a PR for #529 eventually, but it's not a huge priority right now.

PR #628, opened 2026-04-21
feat(studio): add STUDIO_UI_DIST env var for sandbox builds

0xgsvs:
- Add STUDIO_UI_DIST environment variable support
This will help in packaging the cli where building stage is done in sandboxed environment like nix
- Preserves previous behavior as fallback


MicaiahReid (approved):
LGTM, thanks @0xgsvs

PR #642, opened 2026-04-30
feat(sdk): add Surfnet.stop() for graceful shutdown

amilz:
## Summary
- Add `Surfnet::stop(&mut self)` to the Rust SDK that sends `SimnetCommand::Terminate` and blocks until both RPC servers (HTTP + WS) acknowledge shutdown via `SimnetEvent::Shutdown`, with a 5s timeout.
- Returns an error when shutdown is not confirmed so callers can treat unfreed ports as fatal instead of silently leaking sockets.
- Surface `stop()` through the napi binding and TypeScript wrapper so consumers can call `surfnet.stop()` from `afterAll` test teardown.
- `Drop` keeps its best-effort `Terminate` send when `stop()` was not called or failed.
- Install a no-op `log::Log` on first surfnet boot in `sdk-node` to silence parity-ws's hardcoded `println!` fallback that fires on every client-side WS reset on macOS (parity-ws only ignores ECONNRESET when `errno == 104` (Linux); macOS ECONNRESET is 54). Only registers when no logger is already installed, leaving host-configured loggers untouched.

## Test Plan
- `cargo test -p surfpool-sdk -p surfpool-sdk-node` — passes (22 + 0 tests).
- `npm test` in `crates/sdk-node/` — passes (6/6) including a new smoke test that calls `stop()` twice (idempotency on success), confirms the RPC fails after, and verifies the port is rebindable.
- Full workspace `cargo test --workspace` — pre-existing flake in `surfpool-core` unrelated to this change (verified by stashing the diff).
- Verified `sdk-node` consumers no longer see `Encountered an error: ... Enable a logger to see more information.` on stderr during normal client disconnects on macOS.

Closes DEV-467

(Replaces #641)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11542,6 +11542,7 @@ name = "surfpool-sdk-node"
 version = "1.2.0"
 dependencies = [
  "hiro-system-kit",
+ "log 0.4.29",
  "napi",
  "napi-build",
  "napi-derive",
--- a/crates/sdk-node/Cargo.toml
+++ b/crates/sdk-node/Cargo.toml
@@ -14,6 +14,7 @@ crate-type = ["cdylib"]
 napi = { version = "2", features = ["napi4", "napi6"] }
 napi-derive = "2.16"
 hiro-system-kit = { workspace = true }
+log = "0.4"
 surfpool-sdk = { workspace = true }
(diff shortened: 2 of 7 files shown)


MicaiahReid (approved):
Very nice, thank you @amilz!

PR #645, opened 2026-05-05
fix: update disabled features from mainnet

jamie-osec:
This feature list is very out of date, meaning that mainnet state was not properly reflected.


jamie-osec:
It might be nice to have a weekly cron GHA to automatically update these.

MicaiahReid (approved):
Thank you, @jamie-osec! Will create an issue for weekly job to do this, good call.

PR #646, opened 2026-05-05
ci: skip release workflows on forks

beeman:
Add fork repository guards to the release binary and crate publishing workflows so release jobs are skipped when the workflow runs from a fork.


MicaiahReid (approved):
Good call, thanks @beeman!

Issue #636, opened 2026-04-24
Postgres storage not compiled into published Docker image or official installer


zekikucukkose:
### Summary                                                                                                                          
                                                                                                                                 
The documentation states that --db accepts a PostgreSQL connection URL, but neither the published Docker image (surfpool/surfpool) nor the binary installed via curl -fsSL https://run.surfpool.run/ | bash include the postgres feature. Passing a postgres URL produces an immediate error.

### Reproduction

```                                                                                                                            
  # Docker
  docker run --rm surfpool/surfpool:1.2.0 start \                                                                                  
    --db "postgres://user:pass@localhost/surfpool" \                                                                               
    --no-deploy --no-studio                                                                                                        
  # → Error: Failed to initialize Surfnet SVM: Internal error:                                                                     
  #          "Storage error: Postgres storage is not enabled in this build"                                                        
                                                                                                                                   
  # Native installer (macOS arm64)                                                                                                 
  surfpool start --db "postgres://user:pass@localhost/surfpool" \                                                                  
    --no-deploy --no-studio                                                                                                        
  # → same error 
```

### Expected behaviour                                                                                                             
                                                                                                                                   
  Either:
  1. The published Docker image and installer include the postgres feature, OR                                                     
  2. The docs clarify that postgres is not supported

subhdotsol:
CC : @MicaiahReid @lgalabru can anyone of you assign me this issue ?

MicaiahReid:
@subhdotsol, assigned!
The postgres feature is a Rust feature flag that needs to be enabled. So if users want to use postgres, they need to build from source.
So the fix here would be to properly document that, both in the CLI help text and in the error messaging when someone tries to start with postgres but it isn't enabled

PR #650, opened 2026-05-07
fix(cli/core): improve error messaging and help text for postgres feature flag (#636)

subhdotsol:
Pre-built binaries and Docker images do not include PostgreSQL support since it requires the `postgres` Rust feature flag. Users who tried passing a postgres:// URL to --db received an unhelpful error with no guidance on how to enable it.

- Update StorageError::PostgresNotEnabled to include build-from-source instructions using `cargo surfpool-install --features postgres`
- Update --db CLI help text to document that PostgreSQL URLs require building from source with the `postgres` feature flag

### DIFF
--- a/crates/core/src/storage/mod.rs
+++ b/crates/core/src/storage/mod.rs
@@ -104,7 +104,9 @@ where
 pub enum StorageError {
     #[error("Sqlite storage is not enabled in this build")]
     SqliteNotEnabled,
-    #[error("Postgres storage is not enabled in this build")]
+    #[error(
+        "Postgres storage is not enabled in this build. To use PostgreSQL, build surfpool from source with the `postgres` feature flag:\n  git clone https://github.com/solana-foundation/surfpool.git && cd surfpool\n  cargo surfpool-install --features postgres\nPre-built binaries and the official Docker image do not include PostgreSQL support."
+    )]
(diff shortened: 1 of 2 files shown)


crates/core/src/storage/mod.rs
```diff
@@ -104,7 +104,9 @@ where
 pub enum StorageError {
     #[error("Sqlite storage is not enabled in this build")]
     SqliteNotEnabled,
-    #[error("Postgres storage is not enabled in this build")]
+    #[error(
+        "Postgres storage is not enabled in this build. To use PostgreSQL, build surfpool from source with the `postgres` feature flag:\n  git clone https://github.com/solana-foundation/surfpool.git && cd surfpool\n  cargo surfpool-install --features postgres\nPre-built binaries and the official Docker image do not include PostgreSQL support."
```
MicaiahReid, line 108: These `\n`s are not rendering as a new line in the error message, it's just displaying `\n` in the text. In general, it's best practice to verify these sorts of things render correctly when you open a PR :)

We can either remove everything after the `:`, or figure out how to get those line breaks to work, I'm okay with either.

MicaiahReid (changes requested):

MicaiahReid:
Closing in favor of #654

PR #653, opened 2026-05-11
feat(sdk-node): expose feature-gate config in Node SDK

amilz:
## Summary

Closes [DEV-526](https://linear.app/solana-fndn/issue/DEV-526/expose-feature-gate-config-in-surfpool-sdk-node-sdk).

- Adds `enableFeatures`, `disableFeatures`, and `allFeatures` to the napi `SurfnetConfig`, parsed via the existing `parse_feature_pubkey()` helper (kebab-case name, snake_case name, or base58 pubkey).
- Threads through new `SurfnetBuilder::{enable_feature, disable_feature, feature_config}` methods into SVM init.
- Builder default now seeds `SvmFeatureConfig::default_mainnet_features()` so single-feature overrides layer on a mainnet baseline instead of flipping every gate on (matches CLI semantics — previously, `.enable_feature(X)` would activate every known gate, not just `X`).
- Bumps `@solana/surfpool` to **1.3.0**.

## Usage

```ts
import { Surfnet } from '@solana/surfpool';

// Enable v1 transactions (SIMD-0385) on top of the mainnet feature baseline.
const surfnet = Surfnet.startWithConfig({
  enableFeatures: ['enable-loader-v4'],
  // Optional: also accepts base58 pubkeys, snake_case names,
  // or a top-level `allFeatures: true` to flip every known gate on.
});

console.log(surfnet.rpcUrl);
// ...send a versioned transaction, etc.
```

Unknown feature names reject at startup with an `InvalidArg` error listing the available gates.

## Test plan

- [x] `cargo test -p surfpool-types --lib features` — 22/22 pass
- [x] `cargo test -p surfpool-sdk --lib` — 19/19 pass
- [x] `cd crates/sdk-node && npm run build && npm test` — 6/6 pass (incl. new feature-gate case + invalid-name rejection)
- [ ] Follow-up: bump `surfpool-sdk` dep in `@solana/kit-plugin-surfpool` from `^1.2.0` → `^1.3.0` once published, and add a kit-tools V1-transaction test that fails without `enableFeatures` and passes with it.

### DIFF
--- a/crates/cli/src/cli/mod.rs
+++ b/crates/cli/src/cli/mod.rs
@@ -583,11 +583,7 @@ impl StartSimnet {
     pub fn feature_config(&self) -> SvmFeatureConfig {
         let mut config = if self.svm.all_features {
             // Enable all SVM features from agave-feature-set (override mainnet defaults)
-            let mut cfg = SvmFeatureConfig::default();
-            for pubkey in agave_feature_set::FEATURE_NAMES.keys() {
-                cfg = cfg.enable(*pubkey);
-            }
-            cfg
+            SvmFeatureConfig::all_features_enabled()
         } else {
             // No overrides → run with the mainnet baseline (applied by LiteSVM
             // when the SVM is constructed).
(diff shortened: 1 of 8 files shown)


MicaiahReid (approved):
Nice, thanks @amilz!

PR #659, opened 2026-05-13
fix(cli): verify SHA256 of downloaded binary on `surfpool update`

### LINKED ISSUE #626: `surfpool update`: check sha of uploaded binary to ensure download integrity

rit3sh-x:
## Summary
`surfpool update` (added in #620) streams the release tarball and self-replaces the running binary with no integrity verification. This PR adds SHA256 verification using the digest that GitHub already publishes for every release asset — no workflow change required.

- **`crates/cli/src/cli/update/mod.rs`** — `Asset::digest: Option<String>` is now deserialized from the GitHub releases API. After the streaming download completes, the in-memory tarball is hashed with `sha2::Sha256` and compared to the expected hash *before* any disk write or `self_replace`. On mismatch, the update aborts with both hashes printed.
- **`crates/cli/Cargo.toml`** — adds `hex` and `sha2` (both already workspace deps elsewhere in the repo).
- Defensive: if the digest is absent or parses to an unknown algorithm, a warning is printed and the update continues. Practically unreachable — every release from v1.0.0 onward exposes `digest: "sha256:..."` — but keeps the code robust to API surprises and forks.
- 5 inline unit tests on `parse_sha256_digest` (valid input, unknown algorithm, wrong hex length, non-hex, full round-trip).

## Order of operations
```
stream download → buffer in memory
  → hash & compare       ← fail-closed here, no disk touched
  → extract tarball
  → write temp file
  → self_replace
```
A checksum mismatch never reaches `self_replace`, never writes to `/tmp`, and never touches the user's installed binary.

## Test plan
- [x] `cargo check -p surfpool-cli` — clean.
- [x] `cargo test -p surfpool-cli --bin surfpool cli::update::tests` — 5/5 passing.
- [x] `cargo clippy -p surfpool-cli --no-deps` — no new lints on `update/mod.rs`.
- [x] Manual end-to-end: `target/debug/surfpool update --version 1.2.0 --yes` from a patched 1.2.1 build completed cleanly.

Closes #626.


crates/cli/src/cli/update/mod.rs
```diff
@@ -116,6 +136,18 @@ pub async fn handle_update_command(cmd: UpdateCommand) -> Result<(), String> {
...
+        let actual = Sha256::digest(&download);
+        if actual.as_slice() != expected {
+            return Err(format!(
+                "Checksum verification failed for {}.\n  expected: sha256:{}\n  actual:   sha256:{}",
+                users_asset,
+                hex::encode(expected),
+                hex::encode(actual),
+            ));
+        }
+    }
```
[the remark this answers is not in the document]
rit3sh-x: Addressed in the new commit:

rit3sh-x:
@MicaiahReid can you have a look?

crates/cli/src/cli/update/mod.rs
```diff
@@ -92,6 +97,23 @@ pub async fn handle_update_command(cmd: UpdateCommand) -> Result<(), String> {
...
+        }
+        Some(d) => match parse_sha256_digest(d) {
+            Ok(bytes) => Some(bytes),
+            Err(e) => {
+                eprintln!("Warning: {e}; integrity will not be verified");
+                None
+            }
+        },
+    };
+
```
MicaiahReid: Are we able to move this check to before the CLI confirmation?
If they see a warning that the integrity of the download can't be verified, we should give them a chance to bail on the install.

crates/cli/src/cli/update/mod.rs
```diff
@@ -92,6 +97,23 @@ pub async fn handle_update_command(cmd: UpdateCommand) -> Result<(), String> {
             return Ok(());
         }
     }
+
+    let expected_digest: Option<[u8; 32]> = match &asset.digest {
+        None => {
+            eprintln!(
+                "Warning: release asset {users_asset} has no checksum; integrity will not be verified"
```
MicaiahReid:
```suggestion
                "Warning: release asset {users_asset} has no checksum, so the integrity of the release cannot be verified"
```

crates/cli/src/cli/update/mod.rs
```diff
@@ -92,6 +97,23 @@ pub async fn handle_update_command(cmd: UpdateCommand) -> Result<(), String> {
...
+            );
+            None
+        }
+        Some(d) => match parse_sha256_digest(d) {
+            Ok(bytes) => Some(bytes),
+            Err(e) => {
+                eprintln!("Warning: {e}; integrity will not be verified");
```
MicaiahReid:
```suggestion
                eprintln!("Warning: {e}; the integrity of the release cannot be verified");
```

crates/cli/src/cli/update/mod.rs
```diff
@@ -116,6 +138,25 @@ pub async fn handle_update_command(cmd: UpdateCommand) -> Result<(), String> {
     }
     progress_bar.finish_with_message("Download complete");
 
+    // Note: the digest and browser_download_url originate from the same
+    // GitHub API response, so this verifies the tarball against what GitHub
+    // *says* the hash is. It guards against CDN corruption and in-flight
+    // tampering of the tarball itself, but not against a tampered API
+    // response (which would require MITM of api.github.com's TLS).
+    // Stronger guarantees (signed SHASUMS, GitHub Attestations) are a
+    // separate concern, tracked outside this PR.
```
MicaiahReid: This comment says we're tracking externally, but we're not 😆 
Can you make an issue with these above details, then slim down the comment to a shorter summary and point to the issue number?

MicaiahReid (changes requested):

crates/cli/src/cli/update/mod.rs
rit3sh-x: @MicaiahReid addressed in latest commit.

MicaiahReid (approved):
Thanks @rit3sh-x!

PR #686, opened 2026-06-04
fix(core): add SlotHashes to sysvar for first 31 slots

### LINKED ISSUE #684: getSlot('finalized') returns a slot below minimumLedgerSlot at startup (phantom pre-genesis slot), breaking CreateLookupTable
## Summary

Right after a surfnet boots, `getSlot` with `finalized` commitment returns a slot **below `minimumLedgerSlot`**, a slot that was never produced and is absent from `SlotHashes`. Any instruction or call that resolves a finalized recent-slot against `SlotHashes` then fails. The concrete failure we hit is `AddressLookupTable::CreateLookupTable`, which rejects the `recent_slot` with `"N is not a recent slot"`.

> Correction from the original report: `SlotHashes` is **not** shallow. It is 512-deep and fills before eviction, as expected. The real issue is the `finalized` slot underflowing the surfnet's genesis slot, described below.

## Environment

- surfpool `1.3.0` (from `solana-foundation/surfpool` releases)
- reports `solana-core` `3.1.10`
- macOS (arm64)

## Steps to reproduce

1. Boot a surfnet (offline boot from a snapshot, or any boot):
   ```
   surfpool start --offline --snapshot base.json -p 8899 -w 8900 --no-tui
   ```
2. A few seconds after start, compare commitments and the ledger floor:
   ```
   getSlot {commitment: processed}  -> 40
   getSlot {commitment: finalized}  -> 10
   minimumLedgerSlot                -> 31
   getFirstAvailableBlock           -> 31
   ```
3. The finalized slot (10) is below the first slot the surfnet ever produced (31). Fetching it confirms it never existed:
   ```
   getBlock(10) -> error -32602: "Requested 10 is before the first local slot, and no remote RPC was provided."
   ```
4. `SlotHashes` (`SysvarS1otHashes111111111111111111111111111`, `jsonParsed`) holds `[31 .. processed]`, so slot 10 is not in it.
5. Create an address lookup table using the finalized slot as `recent_slot` (the standard `getSlot('finalized')` pattern), e.g. `@solana/web3.js` `AddressLookupTableProgram.createLookupTable({ recentSlot })`.

## Expected

`finalized` is always a real, rooted slot `>= minimumLedgerSlot` and present in `SlotHashes` (as on `solana-test-validator` / mainnet), so `CreateLookupTable` succeeds.

## Actual

`finalized` is computed as roughly `processed - 30` with no floor at the genesis slot, so at startup it returns a phantom pre-genesis slot:

```
Program AddressLookupTab1e1111111111111111111111111 invoke [1]
Program log: Instruction: CreateLookupTable
Program log: 10 is not a recent slot
Program log: Error: InvalidInstructionData
Program AddressLookupTab1e1111111111111111111111111 failed: invalid instruction data
```

The problem clears once the surfnet warms up: at `processed=116`, `finalized=86`, `minimumLedgerSlot=31`, the finalized slot is inside `SlotHashes` and `CreateLookupTable` succeeds. The dead zone is the first ~60 slots, where `finalized` (= `processed - ~30`) is still below the genesis slot (~31).

## Suggested fix

Clamp the `finalized` (and `confirmed`) commitment slot to `>= minimumLedgerSlot`, so `getSlot` never returns a slot the surfnet never produced. Then code that takes a finalized recent-slot works at startup, matching a real validator.

## Workaround

Warm up after boot until `getSlot('finalized') >= minimumLedgerSlot` before issuing slot-sensitive instructions. This lets test suites keep using the standard `finalized` commitment unchanged.

MicaiahReid:
Fixes #684

### DIFF
--- a/crates/core/src/surfnet/svm.rs
+++ b/crates/core/src/surfnet/svm.rs
@@ -43,6 +43,7 @@ use solana_pubkey::Pubkey;
 use solana_rpc_client_api::response::SlotInfo;
 use solana_sdk_ids::{bpf_loader, system_program};
 use solana_signature::Signature;
+use solana_slot_hashes::MAX_ENTRIES as MAX_SLOT_HASHES_ENTRIES;
 use solana_system_interface::instruction as system_instruction;
 use solana_transaction::versioned::VersionedTransaction;
 use solana_transaction_error::TransactionError;
@@ -434,6 +435,16 @@ fn synthetic_chain_tip_at_index(index: u64) -> BlockIdentifier {
         index,
         hash: SyntheticBlockhash::new(index - 1).to_string(),
     }
+}
+
+fn synthetic_blockhash_for_slot(slot: Slot, genesis_slot: Slot) -> SyntheticBlockhash {
+    if slot >= genesis_slot {
+        return SyntheticBlockhash::new(slot - genesis_slot);
+    }
+
+    // Pre-genesis slot hashes only exist to cover the finalized warmup window.
+    // Keep them deterministic and distinct from local chain-index hashes.
+    SyntheticBlockhash::new(u64::MAX - (genesis_slot - slot - 1))
 }
 
 impl SurfnetSvm {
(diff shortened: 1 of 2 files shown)


St0rmBr3w:
Confirmed, this fixes #684. Built the branch: a fresh surfnet now seeds SlotHashes down to slot 0, so the early `finalized` slot is in-window (oldest 31 -> 0, getSlot('finalized')=10 now present). Our AddressLookupTable repro passes at startup on the standard `finalized` commitment, no workaround.

One note for the `--offline --snapshot` flow (how we use surfpool): a snapshot boot restores the `SysvarS1otHashes` stored in the snapshot, which overrides the seeding, so a base captured before this fix still has the old window and still fails. Re-exporting the base with the fixed binary captures the seeded SlotHashes and it works. Non-blocking, just flagging that existing snapshots need a re-export, unless you'd want to re-seed/regenerate SlotHashes on snapshot load so old snapshots benefit too.

Minor: `getBlock(finalized)` still returns "before the first local slot" for that early finalized slot (no block exists there). Doesn't affect CreateLookupTable since it only checks SlotHashes, just noting.

lgalabru (approved):
LGTM, thanks @MicaiahReid !

PR #689, opened 2026-06-05
fix(cli): add pinocchio to Framework::FromStr and replace todo!() with unimplemented!()

### LINKED ISSUE #688: fix(cli): Framework::FromStr is incomplete — pinocchio missing, typhoon panics
### Bug

`Framework::FromStr` has two problems that make the `--framework` flag behave
incorrectly for two of the five supported variants.

**1. `--framework pinocchio` fails with "Unknown framework"**

`Framework::Pinocchio` is fully implemented — it has template dispatch, a
`Display` impl, and project auto-detection in `scaffold/pinocchio.rs` — but
the `FromStr` match has no `"pinocchio"` arm. So any code path that parses a
`Framework` from a string (e.g. an explicit `--framework pinocchio` flag) hits
the fallthrough `Err("Unknown framework: pinocchio")`.

Auto-detection (`surfpool run`, which reads `Cargo.toml` and constructs
`Framework::Pinocchio` directly) works fine, which is why this was never caught.

**2. `--framework typhoon` is accepted but panics at runtime**

`"typhoon"` correctly parses to `Ok(Framework::Typhoon)`, but the two template
generation methods (`get_interpolated_program_deployment_template` and
`get_in_memory_interpolated_program_deployment_template`) have `todo!()` in
the `Typhoon` arms. Acceptance at parse time followed by a panic at runtime is
a poor experience.

### Root cause

Both issues trace back to `crates/cli/src/types/mod.rs`. `FromStr` was written
in commit `0f69d35` and simply never got a `"pinocchio"` arm. The `todo!()`
macros in the template methods have been there since the same commit.

gamandeepsingh:
## Problem

`Framework::FromStr` in `crates/cli/src/types/mod.rs` had two issues:

- `"pinocchio"` was missing from the match, so `--framework pinocchio` always
  returned `Unknown framework: pinocchio` even though `Framework::Pinocchio`
  is fully implemented everywhere else.
- The `Typhoon` arms in both template methods used `todo!()`, which gives a
  misleading "not yet implemented" panic with no context.

## Changes

- Add `"pinocchio" => Ok(Framework::Pinocchio)` to `FromStr`
- Replace `todo!()` with `unimplemented!("Typhoon framework templates are not yet implemented")`
  in both `get_interpolated_program_deployment_template` and
  `get_in_memory_interpolated_program_deployment_template`

## Why `unimplemented!()` over `todo!()`

`todo!()` implies the code is actively being worked on.
`unimplemented!()` correctly signals that this feature doesn't exist yet,
giving a clear message if someone calls it instead of a bare panic.

## Test plan

- [x] `cargo test -p surfpool-cli` — 18/18 passing
- [x] `cargo +nightly fmt --all -- --check` — clean
- [x] `cargo check -p surfpool-cli` — clean
- [x] `cargo clippy --all-targets` — 0 new warnings (pre-existing warnings unrelated to this change)

Closes #688

### DIFF
--- a/crates/cli/src/types/mod.rs
+++ b/crates/cli/src/types/mod.rs
@@ -19,7 +19,7 @@ impl Framework {
     pub fn get_interpolated_program_deployment_template(&self, program_name: &str) -> String {
         match self {
             Framework::Anchor => get_interpolated_anchor_program_deployment_template(program_name),
-            Framework::Typhoon => todo!(),
+            Framework::Typhoon => unimplemented!("Typhoon framework templates are not yet implemented"),
             Framework::Native | Framework::Steel | Framework::Pinocchio => {
                 get_interpolated_native_program_deployment_template(program_name)
             }
@@ -34,7 +34,7 @@ impl Framework {
             Framework::Anchor => {
                 get_in_memory_interpolated_anchor_program_deployment_template(program_name)
             }
-            Framework::Typhoon => todo!(),
+            Framework::Typhoon => unimplemented!("Typhoon framework templates are not yet implemented"),
             Framework::Native | Framework::Steel | Framework::Pinocchio => {
                 get_in_memory_interpolated_native_program_deployment_template(program_name)
             }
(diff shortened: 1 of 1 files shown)


MicaiahReid (approved):
Good find, thanks @gamandeepsingh!

PR #690, opened 2026-06-09
ci: gate crate publishing on unpublished versions

MicaiahReid:
Add an unprotected release preparation job that checks the configured crate versions before entering the protected release environment. Skip the publish job when every crate version is already on crates.io so ordinary main merges do not create deployment approval requests.


lgalabru (approved):
❤️

PR #692, opened 2026-06-10
feat(rpc): add sysvar and feature-gate filters to surfnet_exportSnapshot

MicaiahReid:
Adds `excludeSysvars` and `excludeFeatureGates` boolean fields to `ExportSnapshotFilter`. When set, accounts owned by the sysvar program or whose pubkey appears in `agave_feature_set::FEATURE_NAMES` are omitted from the snapshot. An explicit `includeAccounts` entry bypasses both exclusions. Feature gates are resolved against the set compiled into the running binary; gates added upstream after this version are not excluded.

Includes 5 new unit tests in `svm.rs` covering the filter logic and precedence rules. RPC docstring updated with the two new params and full precedence notes.


lgalabru (approved):
Looks good to me, thanks @MicaiahReid!

PR #70, opened 2025-04-24
docs: rpc endpoints

lgalabru:

PR #693, opened 2026-06-18
fix(core): populate memo and block_time in getSignaturesForAddress for local txs

DanBurton:
I hit an issue with an absent `memo` when trying to switch my testing over from the test validator to windsurf. I guided AI (Claude) to whip up a fix, under the assumption that the omission was not intentional. AI-generated PR description follows.

---

  ## Problem

  For locally-executed transactions, `getSignaturesForAddress` hardcoded both
  `memo` and `blockTime` to `null` in `get_signatures_for_address_local`
  (`crates/core/src/surfnet/locker.rs`) — even when the transaction carried an
  spl-memo instruction and a synthetic block time was available. This diverges
  from `solana-test-validator` and a full Agave validator, where both fields are
  populated. It silently breaks downstream tooling that does memo-based
  idempotency/crash-recovery by scanning a funder's signature history and
  substring-matching the `memo` field.

  `getTransaction` already exposed both the memo instruction and a block time for
  the same signature, so the data was present — the summary path simply wasn't
  reconstructing it.

  This also **aligns the implementation with its own documented contract**: the
  RPC trait docstring (added in #70, before the endpoint was implemented in #178)
  specifies both fields and shows an example response with `"memo": "example-memo"`
  and a non-null `"blockTime"`. The fields were stubbed to `None` when the endpoint
  landed and carried forward unchanged through later refactors, with no TODO or
  rationale — i.e. an unfinished stub rather than a deliberate omission.

  ## Changes

  Two focused commits, both confined to `get_signatures_for_address_local`:                                                                           

  - **`memo`** — reconstructed the same way a full validator does, reusing Agave's
    canonical `extract_and_fmt_memos` over a `VersionedTransactionWithStatusMeta`
    built from the stored transaction and meta. The wrapper's `account_keys()`
    folds in `meta.loaded_addresses`, so memos invoked via an address lookup table
    are handled too. Produces the Agave `"[len] memo"` format (e.g.
    `"[10] hello-memo"`), or `None` when no memo instruction is present. No new
    dependency; the owned transaction/meta are moved into the wrapper (no per-row
    clone).
  - **`block_time`** — synthesized from the slot via
    `calculate_block_time_for_slot` (milliseconds → divided by 1000 for
    `UnixTimestamp` seconds), matching `getBlockTime` and a real Agave validator.

  Only the local path is affected; the remote path already returns these fields
  from the upstream RPC.

  ## Verification
  
  - New TDD test `test_get_signatures_for_address_local_populates_memo_and_block_time`
    (red → green): a transfer+memo tx yields `Some("[10] hello-memo")` with a
    non-null `block_time` matching the slot's synthetic time; a plain transfer
    yields `memo: None`.
  - All 21 existing signature/pagination tests pass (incl. the `before`/`until`
    window coverage from #658) — no regression.
  - Black-box repro against a freshly-built binary: `memo` and `blockTime`
    populated on the memo tx, `memo` null on a control transfer, `blockTime`
    equal to `getBlockTime` for the same slot.
  - `cargo fmt` clean.

  ## Out of scope / known

  `getTransaction.blockTime` has a separate pre-existing double-divide (the stored                                                                    
  block-header time is already in seconds, then divided by 1000 again), returning a
  value ~1000x too small. This PR deliberately emits the *correct* value (matching
  `getBlockTime`), so `getSignaturesForAddress.blockTime` will not equal the
  current `getTransaction.blockTime`. Fixing `getTransaction` (and the related
  `getBlock` null) is left for a follow-up.

  Field contract: https://solana.com/docs/rpc/http/getsignaturesforaddress

### DIFF
--- a/crates/core/src/surfnet/locker.rs
+++ b/crates/core/src/surfnet/locker.rs
@@ -49,7 +49,7 @@ use solana_transaction_error::TransactionError;
 use solana_transaction_status::{
     EncodedConfirmedTransactionWithStatusMeta,
     TransactionConfirmationStatus as SolanaTransactionConfirmationStatus, UiConfirmedBlock,
-    UiTransactionEncoding,
+    UiTransactionEncoding, VersionedTransactionWithStatusMeta, extract_and_fmt_memos,
 };
 use surfpool_types::{
     AccountSnapshot, ComputeUnitsEstimationResult, ExecutionCapture, ExportSnapshotConfig, Idl,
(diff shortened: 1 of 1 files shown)


MicaiahReid (approved):
Great find, great fix, and well-written issue! Thanks @DanBurton!

PR #700, opened 2026-06-25
feat: add stop cmd

### LINKED ISSUE #629: No clean way to stop the daemon surfpool instance
when surfpool is started with `--daemon` flag there is no clean way to stop it using the surfpool cli,  user have to stop the instance with kill command .
adding `stop` subcommand would be ideal.

aspnxdd:
Closes #629 

Add stop command that internally does an RPC request with "exit" - better UX

```bash
surfpool stop
```
image

I also tried the `stop` command in the PR #704 and it works with the --daemon flag : it stops the process.

### DIFF
--- a/crates/cli/src/cli/mod.rs
+++ b/crates/cli/src/cli/mod.rs
@@ -190,14 +190,25 @@ pub struct StartSimnet {
 
 #[derive(Parser, PartialEq, Clone, Debug)]
 pub struct StopCommand {
+    /// RPC host of the running Surfpool instance. Defaults to localhost.
+    #[arg(long = "host", short = 'o', value_name = "HOST")]
+    pub host: Option<String>,
+    /// RPC port of the running Surfpool instance. Required unless --rpc-url is provided.
+    #[arg(
+        long = "port",
+        short = 'p',
+        required_unless_present = "rpc_url",
+        value_name = "PORT"
+    )]
+    pub port: Option<u16>,
     /// Full RPC URL of the running Surfpool instance.
     #[arg(
         long = "rpc-url",
         short = 'u',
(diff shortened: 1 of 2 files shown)


MicaiahReid (commented):
Nice, thank you @aspnxdd! I'm back from leave and reviewing PRs. 

Thoughts on requiring the `--rpc-url` flag? My concern is a user starts some other process (i.e. a real validator) on the default port, then also runs surfpool on an alternate port, then they accidentally `surfpool stop` and kill the important process.

You could leave `rpc-url` optional, and a separate (optional, default `localhost`) `rpc-host` + required (if rpc-url not provided) `port` or `rpc-port` flag. This makes the logic slightly heavier, but makes it a bit more ergonomic: `surfpool stop --port 8899`.

My main concern is that we're calling this `surfpool stop`, but it's really stopping any rpc server living on that port 😆

crates/cli/src/cli/mod.rs
```diff
@@ -962,6 +989,75 @@ fn handle_command(opts: Opts, ctx: &Context) -> Result<(), String> {
...
+    if let Some(rpc_url) = &cmd.rpc_url {
+        return rpc_url.trim_end_matches('/').to_string();
+    }
+
+    let host = cmd.host.as_deref().unwrap_or(DEFAULT_NETWORK_HOST);
+    let port = cmd
+        .port
+        .expect("--port is required when --rpc-url is absent");
+    format!("http://{host}:{port}")
+}
```
MicaiahReid: This seems like it would be perfect to live as an `impl` on the `StopCommand` struct, though what you have works fine too.

MicaiahReid (dismissed):
LGTM! One little comment on clean up if you care to implement, though I'm good either way

MicaiahReid (dismissed):

lgalabru (approved):

PR #702, opened 2026-06-26
feat: async transaction execution

MicaiahReid:


cds-amal:
@MicaiahReid, are you still working this? It overlaps with the lifecycle/state-machine work I've been exploring (#788 is the startup half of that thinking), and I'd rather coordinate than collide. If it's parked, would love to take a swing; if not, I'll bring my thoughts here.

PR #704, opened 2026-06-26
fix(cli): remove fork daemon dependency + fix `start --daemon` flag

### LINKED ISSUE #703: `surfpool start --daemon` flag crashes
Image

In this image you can see that the process was not started (ignore the log `Starting surfpool in daemon mode. Logs are written to the log directory.`, I was adding logs to debug). There are no error logs either, unless:

L937 of crates/cli/src/cli/mod.rs

i switched 
```rs
match daemon(false, false) {
```

to see the logs

```rs
match daemon(false, true) {
```

```bash
./target/release/surfpool start --daemon
Jun 26 10:18:51.167 INFO Starting surfpool in daemon mode

thread 'main' (269301) panicked at /home/arnau/.rustup/toolchains/1.95.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/thread/lifecycle.rs:247:14:
threads should not terminate unexpectedly
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

aspnxdd:
Closes #703

- Removes the `fork` crate dependency from CLI daemon startup.
- Uses the libc daemon FFI directly on Linux.
- Returns a proper error when daemon startup fails instead of continuing silently.

I tried passing `1` as the 2nd arg, just to see the logs, and it does not crash as it does in issue #703

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3305,15 +3305,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "00b0228411908ca8685dba7fc2cdd70ec9990a6e753e89b6ac91a84c40fbaf4b"
 
 [[package]]
-name = "fork"
-version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "05dc8b302e04a1c27f4fe694439ef0f29779ca4edc205b7b58f00db04e29656d"
-dependencies = [
- "libc",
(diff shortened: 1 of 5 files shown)


MicaiahReid:
LGTM, and tested on a Linux box, thanks @aspnxdd!

lgalabru (approved):

PR #707, opened 2026-07-01
fix(core): set parent_slot to slot - 1 for transaction blocks

saigowthamch22:
## Summary

Fixes `getBlock` returning `parentSlot == slot` for blocks that contain transactions. It should be `slot - 1`, matching empty blocks and normal Solana block metadata.

Photon and similar indexers require `parent_slot == last_indexed_slot` to advance; with the wrong value they stall on the first tx block.

## Test plan

- [ ] Start surfpool in clock mode, submit a tx, confirm `getBlock` returns `parentSlot == slot - 1`
- [ ] Photon indexer stays caught up after txs land


saigowthamch22:
Closing to reopen with a clean single-commit fix.

PR #708, opened 2026-07-01
fix(core): set parent_slot to slot - 1 for transaction blocks

### LINKED ISSUE #709: getBlock returns parentSlot == slot for transaction blocks
## Description

`getBlock` returns `parentSlot == slot` for blocks that contain transactions. It should be `slot - 1`, consistent with empty blocks (`reconstruct_empty_block`) and normal Solana block metadata.

Indexers such as Photon require `parent_slot == last_indexed_slot` to advance sequentially. With the incorrect value, indexing stalls on the first tx block.

## Root cause

In `crates/core/src/surfnet/svm.rs`, `confirm_current_block()` stores tx blocks with `parent_slot: slot` instead of `slot.saturating_sub(1)`. Empty blocks already use the correct parent.

## Reproduction

1. Start surfpool and submit any transaction at slot `S`.
2. `getBlock(S)` → `parentSlot` equals `S` (should be `S - 1`).
3. Photon stops advancing after the first tx block.

## Fix

#708

saigowthamch22:
Fixes #709

## Summary

Fixes `getBlock` returning `parentSlot == slot` for blocks that contain transactions. It should be `slot - 1`, matching empty blocks and normal Solana block metadata.

Photon and similar indexers require `parent_slot == last_indexed_slot` to advance; with the wrong value they stall on the first tx block.

## Test plan

- [ ] Start surfpool in clock mode, submit a tx, confirm `getBlock` returns `parentSlot == slot - 1`
- [ ] Photon indexer stays caught up after txs land

### DIFF
--- a/crates/core/src/surfnet/svm.rs
+++ b/crates/core/src/surfnet/svm.rs
@@ -2332,7 +2332,7 @@ impl SurfnetSvm {
                     previous_blockhash: previous_chain_tip.hash.clone(),
                     block_time: self.updated_at as i64 / 1_000,
                     block_height: self.chain_tip.index,
-                    parent_slot: slot,
+                    parent_slot: slot.saturating_sub(1),
                     signatures: confirmed_signatures,
                 },
             )?;


MicaiahReid (approved):
LGTM, thanks @saigowthamch22!

saigowthamch22:
@MicaiahReid Done — re-pushed with a verified signature. Should be good to merge now.

PR #710, opened 2026-07-02
fix(core): stop double-dividing block_time in getTransaction/getBlock responses

serejke:
## Summary

`getTransaction` and `getBlock` currently return `blockTime` as **unix seconds divided by 1000** (e.g. `1782985` instead of `1782985020`), disagreeing with `getBlockTime` and with the `Clock` sysvar the programs execute under.

## Root cause

`BlockHeader.block_time` has been stored in **seconds** since #428 ("fix: block time in secs instead of ms"), which moved the ms→s conversion to the write side:

- block store on confirmation: `block_time: self.updated_at as i64 / 1_000`
- sparse-block reconstruction: `calculate_block_time_for_slot(slot) / 1_000`
- (the unit test pins this: "1020000ms = 1020 seconds")

But two read paths still carry the ms→s division that was only correct **before** #428 (added by #401 back when the header held milliseconds):

- `get_transaction_local` (`crates/core/src/surfnet/locker.rs`): `.map(|b| (b.block_time / 1_000) as UnixTimestamp)`
- the `getBlock` response builder (`crates/core/src/surfnet/svm.rs`): `block_time: Some(block.block_time / 1000)`

`getBlockTime` was already corrected (#484) and the Clock sysvar path is computed independently, so today the clock surfaces disagree with each other.

## Reproduction

Against a running surfnet, execute any transaction and compare:

1. `getTransaction(sig).blockTime` → a value ~1000× too small (e.g. `1782985`),
2. `getBlockTime(slot)` and the `Clock` sysvar's `unix_timestamp` → correct unix seconds (e.g. `1782985020`).

Per the Solana RPC spec all three should agree (estimated production time as unix seconds).

## Fix

Remove the two stale divisions so all block-time surfaces (`getTransaction`, `getBlock`, `getBlockTime`, `Clock` sysvar) report unix seconds. The only existing `block_time` assertion targets header storage (seconds) and is unaffected.

### DIFF
--- a/crates/core/src/rpc/full.rs
+++ b/crates/core/src/rpc/full.rs
@@ -3371,6 +3371,16 @@ mod tests {
             block.signatures.is_none() || block.signatures.as_ref().unwrap().is_empty(),
             "Reconstructed empty block should have no signatures"
         );
+
+        // blockTime must be unix SECONDS (the header stores seconds since #428; a
+        // stale ms->s division here once returned seconds/1000 — ~1000x off, which
+        // this window catches without wall-clock flakiness).
+        let block_time = block.block_time.expect("block has a blockTime");
+        let now = chrono::Utc::now().timestamp();
+        assert!(
+            (now - block_time).abs() < 600,
+            "getBlock blockTime must be unix seconds: got {block_time}, now {now}"
+        );
     }
 
     #[tokio::test(flavor = "multi_thread")]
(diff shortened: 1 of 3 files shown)


MicaiahReid (changes requested):
Thank you @serejke! Great PR as always.

serejke:
@MicaiahReid Hey! Thanks, I've set up the commit signing, and force pushed.

MicaiahReid (approved):

Issue #712, opened 2026-07-10
Clock mode runs ~1.33% fast because blockhash expiry produces an extra block


nhanphan:
### Affected versions

Surfpool 1.2.1 (`d529bb4a`) and the current 1.4.0 release/main (`b46c47e0`).

### Symptom

In clock block-production mode, chain time advances approximately 1.33% faster than wall time. A long-lived process therefore accumulates about 48 seconds of positive `Clock.unix_timestamp` skew per hour.

One production sample over 51.4 seconds measured a chain/wall rate of 1.012x and about 2.55 slots/s with `--slot-time 400`.

### Cause

The clock thread sends one `ClockEvent::Tick` per `slot_time`, then sends an additional `ClockEvent::ExpireBlockHash` every `75 * slot_time`:

https://github.com/solana-foundation/surfpool/blob/d529bb4a773a4eea0db35ca573fef0cbbac7f4a2/crates/core/src/runloops/mod.rs#L541-L600

In clock mode, both events produce a block:

https://github.com/solana-foundation/surfpool/blob/d529bb4a773a4eea0db35ca573fef0cbbac7f4a2/crates/core/src/runloops/mod.rs#L348-L375

Every produced block advances `updated_at` by another `slot_time` and writes it to the `Clock` sysvar:

https://github.com/solana-foundation/surfpool/blob/d529bb4a773a4eea0db35ca573fef0cbbac7f4a2/crates/core/src/surfnet/svm.rs#L1924-L2041

The resulting rate is deterministic:

```text
76 credited intervals / 75 wall intervals = 1.013333x
```

Changing `--slot-time` cannot correct the ratio because it scales the normal tick, expiry interval, and credited time together.

### Expected behavior

In clock mode, `Clock.unix_timestamp` and slot-derived block time should remain synchronized with wall time over long uptimes.

### Suggested direction

Closed PR #711 already proposed a wall-anchored scheduler that absorbs an early expiry block by skipping the next scheduled tick:

https://github.com/solana-foundation/surfpool/pull/711

At minimum, the expiry event should not produce a second block in clock mode, where normal ticks already rotate the blockhash. A regression test should cross several 75-slot expiry boundaries and assert a 1.000x average rate.

MicaiahReid:
Thanks for the detailed issue, @nhanphan! This should be resolved by #746

PR #722, opened 2026-07-26
core: emit the unused SimnetEvent::BlockHashExpired variant

michael-moffett:
## Summary
`SimnetEvent::BlockHashExpired` is defined with two consumers but has no producer anywhere in the tree. This emits it from the `ClockEvent::ExpireBlockHash` path in the clock runloop, wiring up the existing variant.

## Background
This PR began as a regression test for #712. As discussed in the thread, a deterministic test for that bug is not achievable from the current observable surface, because every receiver-side signal derives from a single `confirm_current_block()` counter and cannot distinguish a tick-driven confirm from an expiry-driven one. The test has been removed. Emitting `BlockHashExpired` is the piece that stands on its own and is a step toward the observability a real regression test would need.


michael-moffett:
Thanks, this is correct. The assertion reads wall-clock time on the receive side, so under scheduler pressure correctly-spaced events can drain within the window and false-fail a correct implementation.

I tried to make it deterministic and could not, and the reason is worth recording for #712: every receiver-side signal (SystemClockUpdated, absolute_slot, epoch info, the Clock struct, geyser slot updates) derives from a single confirm_current_block() call, so a Tick-driven confirm and an ExpireBlockHash-driven confirm are indistinguishable downstream. The BlockHashExpired variant exists in SimnetEvent but is never emitted, and the internal ClockEvent channel that would separate the two is not exposed to tests. Without one of those, any assertion collapses back into a wall-clock threshold whose margin is narrower than host sleep drift.

I have reduced this PR to the piece that stands on its own: emitting the unused BlockHashExpired variant from the expiry path, which wires up a producer for the two no-op consumers already in the tree. A deterministic regression test for #712 would need either that event consumed alongside a tick count, or a test hook on the ClockEvent channel. Happy to take either further if useful, or to close this out.

crates/core/src/runloops/mod.rs
```diff
@@ -370,6 +370,9 @@ pub async fn start_block_production_runloop(
                         }
                     }
                     ClockEvent::ExpireBlockHash => {
+                        let _ = svm_locker
```
MicaiahReid: @michael-moffett Do you think we should instead emit this after we call `confirm_current_block`?

`ExpireBlockHash` (as well as `Tick`s and `SlotForward`s) set `do_produce_block = true`, and our only read from `do_produce_block` will `confirm_current_block`. This is what actually produces a block/expires a blockhash. Emitting this event before a new block is actually confirmed and a hash is expired seems too early in the process

michael-moffett:
Yes. It was firing before `confirm_current_block()` could fail, so an error would have announced an
expiry that never happened.

One catch in the move: `do_produce_block` has five writers and one read, so the cause is gone by the
time you reach the confirm, and an emit there would fire on every block. I carried a
`blockhash_expired` flag alongside it, so it fires only on the expiry path and only after the confirm
returns.

Pushed.

MicaiahReid (changes requested):
Looking good @michael-moffett!

MicaiahReid:
Actually @michael-moffett, I'm going to close this in favor of #746. We aren't actually using the expiry event anywhere, and there blockhash expiration always happens when a block is produced - the event is completely redundant. Worse than that, it was actually causing #712 because we were creating an extra block.

michael-moffett:
Makes sense, thanks for explaining rather than just closing it. I was working out how to emit that event correctly and never stopped to ask whether it should exist at all. Removing it is the right call.

PR #725, opened 2026-07-29
feat(types,sdk-node): generate kit cheatcode TS types from Rust via ts-rs

amilz:
## Summary

Stacked on #724. Replaces the hand-maintained wire types in `@solana/surfpool/kit` with bindings generated from the Rust source of truth.

- `surfpool-types` gains an off-by-default `ts-bindings` feature adding `ts-rs` derives (pinned `=12.0.1`) to every `surfnet_*` payload type, plus a `ts_bindings` module mirroring the few types that cannot derive `TS` directly (foreign `solana-account-decoder` types and local enums carrying them in tuple-variant position).
- `npm run generate:kit-types` (crates/sdk-node) regenerates the committed bindings in `surfpool-sdk/kit/generated/`; a new `kit_types_fresh` CI job regenerates and fails on any diff.

## Drift guards

Adding, removing, or reshaping a cheatcode now fails CI at multiple layers:

- A `SURFNET_CHEATCODE_METHODS` manifest in `surfpool-types` is asserted against the methods actually registered by the `SurfnetCheatcodes` trait delegate (runs in `rust.yml` on every PR).
- The generated `methods.ts` drives a compile-time assertion that `SurfnetCheatcodesApi` covers exactly the manifest.
- Mirror types are guarded by serialization-equivalence tests (real vs mirror JSON) and compile-time exhaustiveness matches; both gate every regeneration since the regen script runs the full `surfpool-types` suite.
- The wire shapes of the few deliberately hand-written types (`TimeTravelConfig`, and `PdaSeed`'s digit-adjacent casing) are pinned by serde JSON-key tests.

## Fixes the generated types caught

Real drift the hand-port had missed: the new `TokenAccountUpdate.confidential` field, the `OverrideInstance.account` field, `ExportSnapshotFilter`'s exclude options, `UiAccount.space` being nullable rather than optional, and `aesKey` being required per the documented server contract.


lgalabru (dismissed):
Thanks @amilz 🙏

lgalabru (approved):

PR #727, opened 2026-07-30
chore(sdk-node): bump @solana/surfpool to 1.6.0

amilz:
Bumps the `@solana/surfpool` node SDK to **1.6.0**.

**Background:** 
`1.5.0` was never published to npm — its publish approval flow ([run 29422570066](https://github.com/solana-foundation/surfpool/actions/runs/29422570066)) has sat in `waiting` since 2026-07-15 and was never approved (npm latest is still `1.4.0`). The kit plugin + confidential-transfer changes are already on `main` but stuck under the `1.5.0` label, so a bump is needed to actually ship them.

Bumping to `1.6.0` gives them a clean minor release and keeps the SDK aligned with the upcoming CLI/crates `1.6.0`.

This also regenerates `package-lock.json` with the optional platform-package stubs (via `prepare-release.js`) that the current lock is missing — that gap failed the most recent publish at `npm ci` under npm 11.

**Changes:**
- `package.json` — `version` + `optionalDependencies` → `1.6.0`
- `package-lock.json` — regenerated with platform stubs

### DIFF
--- a/crates/sdk-node/package-lock.json
+++ b/crates/sdk-node/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@solana/surfpool",
-  "version": "1.5.0",
+  "version": "1.6.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@solana/surfpool",
-      "version": "1.5.0",
+      "version": "1.6.0",
       "license": "Apache-2.0",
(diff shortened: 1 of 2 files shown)


amilz:
Superseded by the stub-only 1.5.0 fix — publishing 1.5.0 (the version already on main) rather than bumping to 1.6.0.

PR #729, opened 2026-07-31
feat(sdk-node): stop embedded Surfnet on kit client disposal

amilz:
## What

Makes the embedded `surfpool()` kit client `Disposable`: `surfnet.stop()` now runs when the client is disposed, freeing its RPC/WS ports. Recreating the client boots a fresh Surfnet (stop/start), matching the behavior the wallet plugin relies on.

Implements a suggestion from Anza: wrap the client in `withCleanup` so disposal stops the Surfnet, mirroring the existing stop-on-setup-failure path. Attach mode is untouched — it never owns a Surfnet.

## Changes

- `surfpool.ts` — wrap the embedded client in `withCleanup(client, () => surfnet.stop())`.
- `kit-smoke.js` — test that `client[Symbol.dispose]()` stops the Surfnet (subsequent RPC rejects).
- `README.md` — `using` example.
- `package-lock.json` — sync with the now-published `1.5.0` platform packages. The version-less optional stubs from #728 satisfied `npm ci` only while those packages were unpublished; now that they are on the registry, `npm ci` resolves them and rejects the stubs. Regenerated so they carry real version/resolved/integrity entries.

## Testing

`npm ci` + `npm test` → 20/20 pass (includes `typecheck:kit`).

### DIFF
--- a/crates/sdk-node/README.md
+++ b/crates/sdk-node/README.md
@@ -50,6 +50,15 @@ const slot = await client.rpc.getSlot().send();
 client.surfnet.stop();
 ```
 
+The embedded client is `Disposable`: disposing it stops the Surfnet, so you can
+use `using` instead of calling `stop()` yourself and a recreated client boots a
+fresh instance.
+
+```ts
+using client = await createClient().use(surfpool());
+// Surfnet stops automatically when `client` goes out of scope.
+```
+
 Surfnet startup options go under the `surfnet` key; everything else is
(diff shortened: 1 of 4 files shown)


MicaiahReid (changes requested):
LGTM, thanks @amilz! Can you rebase for me, looks like there are some conflicts with the lock file.

MicaiahReid (approved):

PR #730, opened 2026-08-01
test(core): if you spawn it, you stop it!

cds-amal:
# Problem

`cargo test --workspace` leaks one runloop per test in twelve integration tests. Because each runloop owns a Tokio runtime, RPC ports, and storage, the test process continually accumulates resources instead of returning to its baseline.

CI never exercises these tests because they are behind `ignore_tests_ci`, so the leak only affects local runs, where it turns a developer workstation into a small local sun: **6648 threads, 1274% CPU, three and a half hours in, still running.** 

# Changes

The commits are intended to review in order:

1. Lint the entire workspace (`c0be437`), exposing the `never_loop` lint that hid the problem.
2. Introduce `RunloopGuard` so every test shuts down the runloop it starts, even on panic.
3. Replace panic- and hang-based lifecycle failures with `RunloopError`, bounded waits, and `BootedSimnet`.
4. Widen lifecycle deadlines and make `boot_simnet` return a `BootedSimnet` instead of a tuple.

> `RunloopGuard::drop` fails the test when a runloop outlives it, which turns a silent leak into a visible failure and makes a slow shutdown a red test. 😅 

# Result

Running the leaking tests serially, sampling the test process twice a second:

|              | before | after |
| ------------ | -----: | ----: |
| peak threads |   1404 |   479 |
| wall clock   |   ~21s |  ~22s |

The same workload now peaks at roughly one third the thread count, with no measurable change in runtime.

More importantly, the process no longer accumulates runloops: running one test's three variants in sequence produces thread counts of **76, 77, 81** rather than roughly tripling. Each test leaves the process as it found it.

# How to test

* `cargo test --all --features surfpool-core/ignore_tests_ci`

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -33,7 +33,10 @@ jobs:
       - name: Run Cargo fmt
         run: cargo +nightly fmt --all -- --check
       - name: Run Cargo clippy
-        run: cargo clippy --all-targets
+        # --workspace, because default-members is crates/cli: without it the
+        # other eight crates are linted only as libraries, and their test code
+        # not at all.
+        run: cargo clippy --workspace --all-targets
   build:
     runs-on: ubuntu-latest
     services:
--- a/crates/core/src/tests/integration.rs
+++ b/crates/core/src/tests/integration.rs
@@ -1,7 +1,7 @@
 use std::{str::FromStr, sync::Arc, time::Duration};
 
 use base64::Engine;
-use crossbeam_channel::{unbounded, unbounded as crossbeam_unbounded};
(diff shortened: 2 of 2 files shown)


cds-amal:
Rebased against main

MicaiahReid:
@cds-amal thoughts on the "The PR is not yet safe to merge" comment from greptile? Easy enough fix?

MicaiahReid (approved):

cds-amal:
> @cds-amal thoughts on the "The PR is not yet safe to merge" comment from greptile? Easy enough fix?

[Addressed here](https://github.com/solana-foundation/surfpool/pull/730/changes/870882d67419fe3c76b1339d6df791468da87b99)

PR #731, opened 2026-08-01
Fix/datasource deadline

cds-amal:
## Summary
This PR bounds the lifetime of an RPC call instead of only the lifetime of each individual HTTP request attempt.

Today, the HTTP client's 30-second timeout applies to a single request attempt. On a `429`, Solana's sender retries up to five times while honoring `Retry-After` delays of up to 120 seconds, allowing one RPC call to block for roughly ten minutes without any individual attempt timing out.

A surfnet targets mainnet-beta by default, making this reachable in ordinary use. From the caller's perspective, it is indistinguishable from a surfnet that has stopped responding.

The fix wraps the transport in `DeadlineSender`, which bounds the complete RPC call at 60 seconds (`81fb880`). Both `SurfnetRemoteClient` constructors now route through it, so every method inherits the deadline at the `RpcSender` layer, outside the retry loop.

This PR also adds `async-trait` as a direct dependency of `surfpool-core`.  Solana declares `RpcSender` with `#[async_trait]`; the crate was already existed in the lockfile.

With this change, datasource calls that remain throttled or otherwise fail to complete within 60 seconds now return an error instead of leaving the smurfnet apparently hung. The deadline remains longer than the HTTP client's per-attempt timeout, allowing an in-flight request to complete while placing a finite bound on the overall retry sequence.

### Validation

* `cargo test -p surfpool-core --lib -- a_datasource_that_never_answers`
* `cargo test --all --features surfpool-core/ignore_tests_ci`

### Order

Depends on #730 for the `never_loop` fix because CI runs `cargo clippy --workspace --all-targets` after that PR lands.

### Todo

- [x] Rebase against `main` if/when #730 lands. The rebase will make this PR one single commit, as it currently carries 730's work.

### DIFF
--- a/crates/core/src/surfnet/remote.rs
+++ b/crates/core/src/surfnet/remote.rs
@@ -32,6 +32,7 @@ use solana_rpc_client::{
 use solana_rpc_client_api::client_error::{ErrorKind as ClientErrorKind, Result as ClientResult};
 use solana_signature::Signature;
 use solana_transaction_status::UiConfirmedBlock;
+use surfpool_types::sanitized_datasource_url;
 
 use super::GetTransactionResult;
 use crate::{
@@ -67,10 +68,15 @@ impl<S: RpcSender + Send + Sync> RpcSender for DeadlineSender<S> {
     ) -> ClientResult<serde_json::Value> {
         match tokio::time::timeout(self.deadline, self.inner.send(request, params)).await {
             Ok(response) => response,
+            // Scheme and host only. This message reaches a client through
+            // JSON-RPC error data, and a datasource URL carries credentials in
+            // its query, its path, and its userinfo. A URL that will not parse
+            // is named generically rather than printed raw.
             Err(_) => Err(ClientErrorKind::Custom(format!(
                 "{:?} to {} did not answer within {:?}",
                 request,
-                self.inner.url(),
+                sanitized_datasource_url(&self.inner.url())
+                    .unwrap_or_else(|| "the datasource".to_string()),
                 self.deadline
             ))
             .into()),
@@ -542,7 +548,7 @@ mod tests {
     /// A call that never completes, whether because the endpoint went quiet
     /// or because its retry policy never gave control back. The deadline does
     /// not need to know which.
-    struct NeverAnswers;
+    struct NeverAnswers(String);
 
     #[async_trait]
     impl RpcSender for NeverAnswers {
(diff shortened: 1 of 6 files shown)


crates/core/src/surfnet/remote.rs
```diff
@@ -35,6 +42,56 @@ use crate::{
...
+/// solana's sender retries a 429 up to five times and honours `Retry-After`
+/// for as much as 120 seconds each time, so a throttled datasource can hold a
+/// call for ten minutes with no individual attempt ever timing out. The value
+/// sits comfortably above that 30 second per-attempt timeout, since a deadline
+/// at or below it would cut off attempts that were going to succeed.
+const DATASOURCE_DEADLINE: Duration = Duration::from_secs(60);
+
+/// Bounds how long the sender it wraps may take, so a datasource that stops
+/// answering surfaces as an error rather than as a surfnet that appears stuck.
+struct DeadlineSender<S> {
```
lgalabru, line 57:
```suggestion
struct SurfpoolRpcClient<S> {
```
Rationale: Tweaking the RPC client is opening an interesting avenue of insights / features we could provide, we may want to explore more.

lgalabru: Can we add a `new` constructor?

lgalabru: EDIT: we can actually keep your `DeadlineSender`, but let's introduce a SurfpoolRpcClient wrapping the full construction?

cds-amal: Happy to add both. Two questions first:

1. It sounds like you're describing a builder, or were you thinking of something else?
1. `SurfnetRemoteClient` already exists and the new proposal would be a tad confusing.

If it's a builder, what do you think of one public type, `SmurfpoolRpcClient`, with `SurfnetRemoteClient` folded into it? That one is a breaking change for anyone depending on `surfpool-core` directly, so it wants a major bump or a deprecated alias. 

It also reaches further than it looks: `RunloopContext`, `SurfnetRpcContext` and `SurfpoolMiddleware` all carry it in public fields or signatures. Nothing else in the workspace needs it, though by Hyrum's law it's gonna hurt someone. If not, keeping `SurfpoolRpcClient` internal leaves the public surface as is.

`DeadlineSender::new` either way. 🫡

cds-amal: Thinking about it more: we can have the seam without settling any of that. Keep `SurfpoolRpcClient` private, the way `DeadlineSender` already is in this PR, and the public API doesn't move.

Anything we layer in later (tracing, recording, a retry policy) goes inside it and changes behaviour without touching a signature. Making it public is additive whenever we want it; folding `SurfnetRemoteClient` away isn't.

I'll go with that unless you'd rather have it public now.

cds-amal:
Rebased to previous stack + address @lgalabru's feedback.

crates/core/src/surfnet/remote.rs
```diff
@@ -35,6 +41,51 @@ use crate::{
...
+        request: RpcRequest,
+        params: serde_json::Value,
+    ) -> ClientResult<serde_json::Value> {
+        match tokio::time::timeout(self.deadline, self.inner.send(request, params)).await {
+            Ok(response) => response,
+            Err(_) => Err(ClientErrorKind::Custom(format!(
+                "{:?} to {} did not answer within {:?}",
+                request,
+                self.inner.url(),
+                self.deadline
```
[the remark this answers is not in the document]
cds-amal: FYI, [Addressed here](https://github.com/solana-foundation/surfpool/pull/731/changes/83a0c2ce6182ce7c85159a509fad6fc18a30ebed), @MicaiahReid

crates/core/src/surfnet/remote.rs
```diff
@@ -35,6 +42,112 @@ use crate::{
...
 
+/// How long one call to the datasource gets, start to finish.
+///
+/// The HTTP client's timeout bounds a single attempt rather than a call:
+/// solana's sender retries a 429 up to five times and honours `Retry-After`
+/// for as much as 120 seconds each time, so a throttled datasource can hold a
+/// call for ten minutes with no individual attempt ever timing out. The value
+/// sits comfortably above that 30 second per-attempt timeout, since a deadline
+/// at or below it would cut off attempts that were going to succeed.
+const DATASOURCE_DEADLINE: Duration = Duration::from_secs(60);
```
cds-amal, line 53: @MicaiahReid i could see this being a configurable setting, with a sane default.

MicaiahReid: I'm good with this as a constant, a user can open an issue if they need to configure, but this default seems fine

MicaiahReid (approved):
Looking good, thanks @cds-amal!

PR #732, opened 2026-08-02
fix(core): support JSON-RPC batch requests

### LINKED ISSUE #717: JSON-RPC batch requests are rejected
Any batched JSON-RPC request (a top-level array of calls) is rejected with:
WARN Request rejected due to not being a single method call
Breaks @solana/web3.js Connection.getTransactions([...]), which is implemented as a JSON-RPC batch of getTransaction calls; client-side it surfaces as TypeError: unsafeRes.map is not a function.

Repro: POST [{"jsonrpc":"2.0","id":1,"method":"getSlot"},{"jsonrpc":"2.0","id":2,"method":"getSlot"}] → rejected. Or web3.js connection.getTransactions([sigA, sigB]).

Expected: batch array processed, array of results returned (JSON-RPC 2.0 batching is standard and supported by mainnet RPCs).

Note: distinct from getMultipleAccounts (a single method taking an array param — that's fine). The gap is batch request support.

koriyoshi2041:
## Problem

The HTTP middleware rejects every top-level JSON-RPC batch before the handler can dispatch it. This breaks clients such as `@solana/web3.js` when they use batched RPC calls.

## Fix

Move method-level logging and cheatcode lockout checks to the middleware's per-call hook. The JSON-RPC handler can now dispatch batch entries independently while each entry still receives the same policy checks as a single call. Batch notifications are executed without producing response entries, as required by JSON-RPC 2.0.

## Test

- `cargo test -p surfpool-core middleware_tests -- --nocapture` — 2 passed
- `cargo clippy -p surfpool-core --lib --no-deps` — passed with existing repository warnings
- `git diff --check` — passed

The regression tests cover multiple IDs in one batch and a mixed notification/method-call batch.

## Risk

The change is limited to HTTP middleware dispatch. Single method calls keep the same metadata and cheatcode lockout behavior. Request-body size remains bounded by the server's existing 15 MiB limit. WebSocket middleware is unchanged.

Closes #717.

### DIFF
--- a/crates/core/src/rpc/mod.rs
+++ b/crates/core/src/rpc/mod.rs
@@ -6,7 +6,8 @@ use std::{
 use blake3::Hash;
 use crossbeam_channel::Sender;
 use jsonrpc_core::{
-    BoxFuture, Error, ErrorCode, FutureResponse, Metadata, Middleware, Request, Response,
+    BoxFuture, Call, Error, ErrorCode, FutureResponse, Metadata, Middleware, Output, Request,
+    Response,
     futures::{FutureExt, future::Either},
     middleware,
 };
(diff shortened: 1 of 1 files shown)


MicaiahReid:
Thanks for the PR, @koriyoshi2041!! I'm reviewing now! In the meantime, Foundation repos require verified commit signatures to merge. Can you [set up commit signature verification](https://docs.github.com/en/authentication/managing-commit-signature-verification) and repush this change so I can merge?

MicaiahReid (changes requested):
This is awesome, thanks again @koriyoshi2041! Just waiting on the verified commits and CI to be green and we'll merge!

MicaiahReid:
Hey @koriyoshi2041 just following up on if you can set up verified commits and re-push

MicaiahReid:
@koriyoshi2041, closing PR since I can't merge without signature verification. Feel free to reopen once that's taken care of.

PR #733, opened 2026-08-02
fix: gate readiness on a sealed startup plan

### LINKED ISSUE #715: Anchor account cloning seems racey
Recently, Anchor tests that clone programs onto the test validator have been intermittently failing: https://github.com/otter-sec/anchor/actions/runs/29356983948/job/87168653926

It seems that Surfpool is reporting readiness before all accounts are cloned and ready.

cds-amal:
Closes #715. A surfnet now signals ready only after every requested startup task has completed, so the race where Anchor saw ready mid-clone can no longer occur.

### Todo (before merge)

- [x] drop `tests/issue-715/` folder

### The fix

Startup becomes a lifecycle with one rule: ready means the plan is **sealed** (the task list is closed: nothing can be added) and every task in it has succeeded. `SurfnetStartupStatus` carries the rule structurally: its unsealed variants have no task table, so registering a task before sealing, or deriving readiness from an unsealed status, cannot be constructed. Who seals depends on the caller: the CLI seals once it knows the clone list; for embedders (`crates/mcp`, `crates/sdk`) the runloop seals an empty plan immediately, so their behavior is unchanged.

The spec is `crates/types/src/startup-lifecycle.md`. It ships as the enum's rustdoc, and the model check asserts the implementation against an oracle written from it, so CI tests "code matches spec" instead of "code matches itself".

The spec's tables and its diagram are generated, with the regeneration commands aliased in `.cargo/config.toml`. `cargo surfpool-update-startup-spec` rewrites the tables and needs nothing beyond Rust; `cargo surfpool-render-startup-diagrams` re-renders the SVG and rides the standalone diagram commit. Rendering is the one step with an external dependency: `mmdc`, the Mermaid CLI (`npm install -g @mermaid-js/mermaid-cli`). Since the SVG only goes stale when the diagram source changes, that step could live in CI rather than on contributor machines. Either way the ordinary test run fails when an artifact is stale, and the failure message names the alias to run.

<details>
<summary> 👉 <strong>EXPAND: How I mapped this out 👈 </strong> </summary>

## The bug: Work lagged behind readiness

Production before this change had no startup plan. Readiness followed core startup, and clone hydration ran alongside it with nothing gating the answer, so a client could observe readiness before its declared accounts existed. In ordering terms, the bug was a missing edge.

```mermaid
flowchart TB
    core["core startup, RPC bound"]
    info["SurfnetInfo shows<br/>runbookExecutions = []"]
    obs["client poll answers ready"]:::bad
    inspect["inspect Anchor.toml"]
    fetch["fetch dispatched<br/>(fire and forget)"]
    inst["clones installed"]

    core --> info --> obs
    core --> inspect --> fetch -. "HTTPS" .-> inst
    inst -. "MISSING: nothing ordered<br/>installation before observation" .-> obs

    classDef bad fill:#fee2e2,stroke:#dc2626,color:#7f1d1d,stroke-width:2px
```

"Client poll answers ready" was reachable without passing through "clones installed" at all. Everything in the fix exists to add ordering edges until the cut is total.

## The invariant

A client may observe `Ready` if and only if the startup plan has been sealed and every required startup task has completed successfully.

An unsealed plan can never become ready, even when its current task collection is empty: an unsealed status has no task table to derive readiness from, which is why a sealed empty plan is ready and an unsealed one never is. Everything below is evidence for that rule.

## What a client observes

What a client polling `surfnet_getSurfnetInfo` observes.

```mermaid
stateDiagram-v2
    [*] --> Planning
    Planning --> Initializing: plan sealed, clones pending
    Initializing --> Deploying: clones installed
    Deploying --> Ready: every task succeeded
    Planning --> Ready: sealed with an empty plan
    Planning --> Failed: planning failed
    Initializing --> Failed: task failed
    Deploying --> Failed: task failed
```

`Ready` carries `pending = {}`; every other non-terminal phase carries a non-empty pending set. `Failed` is terminal.

Anchor versions predating an explicit startup field infer readiness by checking that every entry in `runbookExecutions` is complete, in a loop with no timeout that never inspects `errors`. So the lifecycle is also projected into that list, recomputed from the current phase on every response:

| Lifecycle phase                   | Projected into `runbookExecutions`                                |
|-----------------------------------|-------------------------------------------------------------------|
| Planning, Initializing, Deploying | one pending `surfpool-startup` entry (`completedAt` null)         |
| Ready                             | no entry                                                          |
| Failed                            | the entry, completed, with `errors` carrying the failure messages |

The Failed row is completed rather than pending because legacy Anchor's loop has no timeout and never reads `errors`: a forever-pending entry would starve it, while a completed one lets it proceed and fail visibly. The entry's `startedAt` is the surfnet's start time, stable across polls, because clients diff the execution list between polls and a fresh timestamp per response read as a new execution every time. The headless CLI additionally aborts the process on `Failed`, which is the only failure signal that loop can perceive.

## Proof: why `Ready` cannot happen early

Every edge below is a happens-before guarantee. If there is a path from A to B, B cannot occur before A. The label names the primitive that provides it.

```mermaid
flowchart TB
    seal["plan sealed"]:::cut
    okC["clones succeeded"]:::cut
    okD["deployment succeeded"]:::cut
    ready["phase = Ready"]:::cut
    obs["client poll answers ready"]

    seal -- "mpsc FIFO" --> okC
    seal -- "mpsc FIFO" --> okD
    okC -- "join" --> ready
    okD -- "join" --> ready
    ready -- "read-guard projection" --> obs

    classDef cut fill:#dcfce7,stroke:#15803d,color:#14532d,stroke-width:2px
```

Every path to the client passes through the seal and through both task successes. That green cut is total: no path reaches a client answer without crossing it. The intermediate steps (accounts arriving, the batch installing, tasks being dispatched) are implementation, not proof.

Five primitives carry those edges, each with a local and mechanical guarantee:

1. **The seal is a fence.** `SealStartupPlan` is a synchronous round trip, and
   every task command is sent after it on the same FIFO mpsc channel. The
   runloop can never observe work for an unsealed plan; the ordering argument
   is one channel's delivery order.
2. **One writer, no interleaving.** Only the runloop mutates startup state, so
   transitions serialize by construction, and the transition table in
   [state-table.md](https://github.com/cds-rs/surfpool/blob/fix/715/tests/issue-715/state-table.md) is checkable by exhaustive enumeration.
   There is no concurrent schedule to miss.
3. **Hydration is atomic.** The clone batch installs under a single write
   guard, so `Ready` can never expose a partially installed clone set.
4. **Readiness is level-triggered.** The watch channel publishes a value, not
   an event, so late subscribers read the current phase instead of hoping they
   subscribed before an edge.
5. **Clients see projections, never internals.** Both the `startup` field and
   the compatibility entry are computed from the lifecycle under a read guard
   at response time, which is what lets unmodified legacy Anchor benefit.

## The machinery

Subgraphs are process boundaries. Solid edges carry requests or commands,
dashed edges carry replies, events or observations, and the label names the
primitive.

```mermaid
flowchart TB
    subgraph remote["Remote RPC process (mainnet / devnet / testnet)"]
        datasource["JSON-RPC endpoint"]:::external
    end

    subgraph surfpool["surfpool process"]
        direction TB
        cli["CLI orchestration<br/>startup planner"]:::thread
        runloop["Block-production runloop<br/>single consumer; the only writer<br/>of startup machine state"]:::thread
        deployer["Deployment executor thread"]:::thread
        watchdog["Startup watchdog thread<br/>headless only"]:::thread
        tui["TUI / log frontend"]:::thread
        locker["SVM locker<br/>RwLock over SurfnetSvm,<br/>owns the startup status<br/>watch sender"]:::state
        rpcsrv["RPC servers<br/>HTTP and WS"]:::thread
    end

    subgraph clients["Client processes"]
        anchor["Anchor CLI"]:::external
    end

    cli -- "SimnetCommand<br/>(mpsc, FIFO)" --> runloop
    runloop -. "seal outcome<br/>(bounded reply channel,<br/>5s timeout)" .-> cli
    deployer -- "CompleteStartupTask<br/>(same mpsc)" --> runloop
    runloop -- "mutations under<br/>one write guard" --> locker
    rpcsrv -- "reads under<br/>read guard" --> locker
    locker -. "startup status<br/>(tokio watch channel)" .-> watchdog
    runloop -. "SimnetEvent<br/>(mpsc)" .-> tui
    deployer -- "deployment transactions<br/>(loopback JSON-RPC)" --> rpcsrv
    runloop -- "getMultipleAccounts<br/>(HTTPS JSON-RPC)" --> datasource
    anchor -- "getSurfnetInfo poll,<br/>getAccountInfo<br/>(HTTP JSON-RPC)" --> rpcsrv

    classDef external fill:#eceff1,stroke:#546e7a,color:#263238;
    classDef thread fill:#f3e8ff,stroke:#9333ea,color:#581c87;
    classDef state fill:#dcfce7,stroke:#16a34a,color:#14532d;
```

Two edges cross a process boundary, and both only emit requests: the runloop pulls clone accounts from the remote, and clients pull the read model from surfpool. Neither can push state into the lifecycle. All the pushing happens inside the surfpool process, on channels with exactly one consumer.

Where each piece lives:

| Piece                                                      | Location                           |
|------------------------------------------------------------|------------------------------------|
| Startup machine (`SurfnetStartupStatus`)                   | `crates/types/src/startup.rs`      |
| Planner, seal, watchdog                                    | `crates/cli/src/cli/simnet/mod.rs` |
| Command handling, remote fetch, atomic install             | `crates/core/src/runloops/mod.rs`  |
| Watch channel (`subscribe_startup_status`)                 | `crates/core/src/surfnet/svm.rs`   |
| Compat projection (`GetSurfnetInfoResponse::with_startup`) | `crates/types/src/startup.rs`      |

Three processes participate: the remote RPC that clones are fetched from, the surfpool process, and the clients. SDK and MCP embedders run the core in their own process instead, sealing an empty plan at construction, so the same lifecycle applies with no tasks to wait for.

The full documents carry more:
[event-model-simple.md](https://github.com/cds-rs/surfpool/blob/fix/715/tests/issue-715/event-model-simple.md)
folds each pipeline into a composite state and shows the fix as a change of
topology;
[event-model.md](https://github.com/cds-rs/surfpool/blob/fix/715/tests/issue-715/event-model.md)
adds the per-pipeline command and event flows;
[process-model.md](https://github.com/cds-rs/surfpool/blob/fix/715/tests/issue-715/process-model.md)
adds the client journeys and the per-task and watchdog state machines;
[state-table.md](https://github.com/cds-rs/surfpool/blob/fix/715/tests/issue-715/state-table.md)
enumerates every transition with its guards.

</details>

### Design choices to be reviewed

1. **`Failed` completes the legacy compatibility entry** instead of leaving it
   pending. A pending entry would park a timeout-free legacy client forever;
   completing it lets the client proceed and fail visibly, with the reason in
   `errors`.
2. **A failed startup aborts a headless run and is displayed by the TUI.**
   Headless (`--no-tui` or `--daemon`) spawns a watchdog that terminates the
   process at `Failed`: legacy Anchor perceives only `completed_at` and
   process death, so a surfnet nobody is watching must die rather than serve
   forever. The TUI spawns no watchdog; killing an interactive session felt
   wrong, so it stays alive and shows the failure.
3. **The new `startup` field on `GetSurfnetInfoResponse` is `ts(skip)`.** The
   wire format comes from a hand-written serialize impl, and ts-rs derives
   from the Rust shape, so exporting it would mean hand-maintaining a mirror.
   TypeScript clients keep reading `runbookExecutions`.
4. **The diagram commit is standalone by design**: declining the rendered SVG
   is one revert of `docs(types): render the startup lifecycle diagram in
   rustdoc`.

### Background & development notes you might find interesting.

See the two model documents in `tests/issue-715/` 

1. [event-model.md](https://github.com/cds-rs/surfpool/blob/fix/715/tests/issue-715/event-model.md)
1. [process-model.md](https://github.com/cds-rs/surfpool/blob/fix/715/tests/issue-715/process-model.md)

> the folder drops before merge, the PR description keeps this copy

The original reproduction: a mock RPC that delays `getMultipleAccounts`, a mirror of Anchor's readiness algorithm, and `compare.sh` with bisect-compatible exit codes, so `git bisect run` can answer "does revision X have the bug" for any revision, including ones predating this test suite. The model check proves the property, but only for this implementation; the harness demonstrates the bug anywhere. Kept in the branch for review; drops before merge.

### Validation

* `cargo test -p surfpool-types --lib` (and again with `--features ts-bindings`)
* `cargo test -p surfpool-core --lib --features ignore_tests_ci`
* `cargo +nightly fmt --all -- --check`
* `tests/issue-715/compare.sh` (compares `main` against `HEAD`)
* Read docs: `cargo doc -p surfpool-types --no-deps --open`, then open `surfpool_types::startup::SurfnetStartupStatus` - the tables should be populated.


crates/types/build.rs
```diff
@@ -0,0 +1,46 @@
+use std::{env, fs, path::Path};
+
+/// Produces the rustdoc variant of `startup-lifecycle.md`: each mermaid
+/// region is replaced by its pre-rendered SVG from `src/diagrams/`, so
+/// rustdoc shows the drawing while the source file keeps the editable
+/// fence, which GitHub and editors render natively. The test
+/// `the_diagrams_match_their_renderings` holds the SVGs to their sources;
+/// this script only splices.
+fn main() {
```
cds-amal, line 9: 😅

crates/cli/src/cli/simnet/mod.rs
```diff
@@ -617,11 +879,135 @@ impl RunbookExecutionMode {
...
+    response_rx
+        .recv_timeout(SEAL_STARTUP_PLAN_TIMEOUT)
+        .map_err(|e| SealFailure::Unreachable(format!("Timed out sealing startup plan: {e}")))?
+        .map_err(SealFailure::Refused)
+}
+
+/// Everything the planner learned about the project before sealing: the task
+/// list to seal, the work to dispatch, and the context the artifact watcher
+/// needs to re-execute runbooks later.
+struct DeploymentPlan {
```
MicaiahReid: Thoughts on renaming to `StartupPlan`? Because it includes additional startup tasks beyond the deployment.

I also think this could make sense to move, with associated functions (`plan_and_dispatch_startup`, `plan_deployment`) and structs (`SealFailure`, `SealFailure`, etc) to a separate mod in `simnet/`?

crates/cli/src/cli/simnet/mod.rs
```diff
@@ -644,171 +1031,323 @@ async fn write_and_execute_iac(
...
-                            hiro_system_kit::nestable_block_on(join_all(futures))
-                        }));
+    if !futures.is_empty() {
+        let _ = simnet_commands_tx.send(SimnetCommand::StartStartupTask(
+            SurfnetStartupTask::Deployment,
+        ));
+
+        let startup_commands_tx = simnet_commands_tx.clone();
+        if let Err(error) =
+            hiro_system_kit::thread_named("Deployment Runbook Executions").spawn(move || {
```
MicaiahReid: This wasn't you, it was existing, but can we change to `"Startup Runbook Executions"`? And as a whole steer runbook things to runbooks rather than deployments, if you happen to notice any others? Because a user could be executing any number of runbooks for any number of purposes on startup.

crates/cli/src/cli/mod.rs
```diff
@@ -725,6 +725,10 @@ impl StartSimnet {
             subgraph: self.subgraph_config(),
             studio: self.studio_config(),
             plugin_config_path,
+            // The CLI inspects the project (clones, deployment runbooks)
+            // after the runloop starts and seals the plan itself; the
+            // runloop must not seal an empty one out from under it.
+            startup_planner: StartupPlanner::External,
```
MicaiahReid: It's not super clear to me how this comment is related to the code `startup_planner: StartupPlanner::External`. Could the comment be changed or augmented to say what the `External` variant will do here?

crates/cli/src/tui/simnet.rs
```diff
@@ -487,6 +487,10 @@ fn run_app<B: Backend>(terminal: &mut Terminal<B>, mut app: App) -> io::Result<(
                 match oper.index() {
                     0 => match oper.recv(&app.simnet_events_rx) {
                         Ok(event) => match &event {
+                            // The startup panel reads the watch channel; the
+                            // sequence matters to tests, not to the display.
```
MicaiahReid: Not sure what a "startup panel" or a "watch channel" are 😅 
AI halucinations? 


```suggestion
                            // The startup sequence matters to other listeners, not to the TUI
```

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -2202,9 +2214,43 @@ impl SurfnetCheatcodes for SurfnetCheatcodesRpc {
...
+
+        // The predicate is the client's own conclusion from this response, not
+        // the surfnet's internal phase: Anchor reads readiness as "every
+        // execution reports complete", and an empty list satisfies it. Deriving
+        // it from what was answered keeps the event meaningful for any build,
+        // including one whose readiness rule differs.
+        // try_send: this runs on every poll, and the bounded events channel
+        // may belong to an embedder that never drains it. A full channel
+        // costs one ordering-evidence event; a blocking send would park an
+        // RPC worker for as long as the channel stays full.
```
MicaiahReid: To me the first half of this comment is implying that our `AnsweredClient` should be carrying more information than the response to the RPC request. In reality, this `Readiness` struct just has a bool.

Or maybe I'm misreading this comment, which I've read 100 times and don't fully grok 😅 

Think we can simplify and shorten this comment in the least? Very verbose.

I also think the `try_send` portion is unnecessary

crates/core/src/runloops/mod.rs
```diff
@@ -501,16 +571,61 @@ pub async fn start_block_production_runloop(
...
+                                    // batch, so Ready cannot expose a partially installed clone set.
+                                    svm_locker
+                                        .write_multiple_account_updates(&account_updates.inner);
+
+                                    // A declared account the datasource does not have is not a
+                                    // failure: hydration did complete, and some workflows clone
+                                    // addresses that do not exist yet. It is worth saying out
+                                    // loud, though, because the surfnet reaches Ready with the
+                                    // account absent, which looks exactly like the readiness race
+                                    // this plan exists to prevent.
```
MicaiahReid: Ideally let's not reference "this plan" lingo in code. 
Also: 

Image

```suggestion
                                    // Let the user know if the account they requested to clone
                                    // doesn't exist at the datasource
```

MicaiahReid (changes requested):
Will continue my review in the AM. 

In the meantime, could you clean up some of the stuff in the `/issue-715` folder? I don't think we need the bisect/compare scripts. We don't need the Anchor.toml, assert-anchor-ready.mjs (or do we?).

Accross just that folder we have about 2k lines of new code to maintain, I think that's just too much. _Maybe_ a single simple md file explaining the arch, but I'm not sure. I think the code + comments speaks for itself.

Also, maybe just a note for me as I review in the morning, but I found it interesting that `AnsweredClient/ClientAnswer` for the ready state carries potentially relevant information that is only seeded/available if the user happens to call `surfnet_getSurfnetInfo`. That seems potentially off to me, but I haven't followed the full flow of that use yet.

crates/cli/src/tui/simnet.rs
cds-amal: tl;dr; `Other listeners` is correct! 

Watch channel is the tokio primitive (`tokio::sync::watch::Receiver`).

I can't blame the LLM for `Startup panel`. I'm pretty sure it was me picking the wrong hunk and not checing during a rebase. Startup panel was the diagram name that held all valid listeners/clients (tui/headless) that could interact with a surfpool.

crates/core/src/runloops/mod.rs
cds-amal: Guilty .. kilroy got through! I'll have to do another scan. Sorry about that!

crates/cli/src/cli/mod.rs
cds-amal: clarified; it's an important variant. The default seals an empty plan, and this variant allows an external planner (CLI) to seal what said planner needs from surfnet for testing to begin.

> // `External` makes the runloop wait for someone else to seal the
> // startup plan instead of sealing an empty one at boot. Here that
> // someone is the CLI: it inspects the project (clones, startup
> // runbooks) after the runloop starts and seals the plan itself.
>

crates/cli/src/cli/simnet/mod.rs
cds-amal: This makes sense, run-books are the get-on-chain mechanism

crates/cli/src/cli/simnet/mod.rs
cds-amal: That's a better name.

crates/core/src/rpc/surfnet_cheatcodes.rs
cds-amal: Sorry, it's better shorter. 

> Record the readiness verdict the client would read from this
> response (Anchor: every execution complete; an empty list
> qualifies). try_send: an embedder may never drain the bounded
> events channel, and a ful one must not park an RPC worker.

where embedder is the host program that embeds the smurfnet.

crates/cli/src/cli/simnet/mod.rs
```diff
@@ -255,17 +251,45 @@ pub async fn handle_start_local_surfnet_command(
...
-                let _ = simnet_events_tx.send(SimnetEvent::warn(format!(
-                    "Automatic protocol deployment failed: {e}"
-                )));
+        match plan_and_dispatch_startup(&cmd, &simnet_events_tx, &simnet_commands_tx_copy).await {
+            Ok(rx) => runbook_progress_rx.push(rx),
+            // Planning failed before the plan was sealed. Drive the machine
+            // to Failed (from Planning-unsealed this always applies) and
+            // keep going: whether a failed startup is fatal is the
+            // watchdog's decision, headless aborts while the TUI stays
+            // alive and displays it.
```
MicaiahReid:
```suggestion
            // watchdog's decision
```
It's the watchdog's decision, so the watchdog's job to document :)

(future changes to watchdog would require changes to this comment as well, seems like we'd want to avoid that)

crates/cli/src/cli/simnet/mod.rs
```diff
@@ -74,7 +66,8 @@ pub async fn handle_start_local_surfnet_command(
 ) -> Result<(), String> {
     // Local plugin loading is handled directly by `surfpool-core`.
 
-    // We start the simnet as soon as possible, as it needs to be ready for deployments
+    // We start the simnet as soon as possible; startup planning waits for
+    // its `Ready` event before sealing.
```
MicaiahReid:
```suggestion
    // We start the simnet as soon as possible; startup planning (runbook
    // executions + account cloning) waits for the simnet's `Ready`
    // event before being finalized
```
Tbh, this "sealing" language is confusing, particularly in comments outside of the actual state machine for the startup plan.

Also to me "startup planning" in the context of starting the simnet can be ambiguous - it's not a plan for starting the actual simnet, it's for what to do immediately after starting.

Starting to wonder about `StartupHooks` 🤔 but I don't want to bog things down on this PR

crates/types/src/types.rs
```diff
@@ -576,6 +623,23 @@ pub enum SimnetCommand {
         Option<bool>,
     ),
     Terminate(Option<(Hash, String)>),
+    /// Fixes the startup task list; `Ready` is unreachable until a plan is
+    /// sealed. The only startup command with a reply channel: the planner
+    /// must not dispatch tasks against an unsealed plan, so it blocks on the
+    /// outcome. The startup commands below are fire-and-forget
+    /// because no caller decision hangs on them; their failures are machine
+    /// rejections, which the runloop reports as error events.
```
```suggestion
    /// Fired by the startup planning process. Once the block production
    /// runloop receives this event, the startup plan is sealed. This means
    /// that the VM won't be marked `Ready` until each task in the plan is
    /// completed.
    /// The payload includes a reply channel (`Sender<>`) because the planner
    /// must not start tasks until the plan is sealed, so it blocks on the
    /// outcome.
```

MicaiahReid: Not necessarily better, but typing it out helped me to understand the purpose better

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -2202,9 +2214,38 @@ impl SurfnetCheatcodes for SurfnetCheatcodesRpc {
...
+            .try_send(SimnetEvent::AnsweredClient {
+                method: "surfnet_getSurfnetInfo",
+                answer: ClientAnswer::Readiness {
+                    ready: value
+                        .runbook_executions
+                        .iter()
+                        .all(|execution| execution.completed_at.is_some()),
+                },
+            });
+
```
MicaiahReid: I'm still not following why we need to record the readiness verdict? This `AnsweredClient` is not used in any loop. It's read from in one test, but as far as I can tell, the purpose of that test is to validate... that this event is fired?

crates/cli/src/cli/simnet/mod.rs
```diff
@@ -255,17 +251,45 @@ pub async fn handle_start_local_surfnet_command(
...
+            // watchdog's decision, headless aborts while the TUI stays
+            // alive and displays it.
+            Err(StartupPlanFailure::Planning(e)) => {
+                let _ = simnet_commands_tx_copy.send(SimnetCommand::FailStartupPlanning(e.clone()));
+                let _ = simnet_events_tx
+                    .send(SimnetEvent::warn(format!("Startup planning failed: {e}")));
+            }
+            // The command loop is dead or wedged, so the machine is
+            // unreachable and no session can ever become ready. Nothing left
+            // to display; exit.
```
MicaiahReid: Sometimes AI comes up with the weirdest word to describe things 😆 

What does "wedged" mean hear? Blocked?

Also, who is "machine" here?

crates/cli/src/cli/simnet/mod.rs
```diff
@@ -255,17 +251,45 @@ pub async fn handle_start_local_surfnet_command(
...
+            Err(StartupPlanFailure::Sealing(SealFailure::Unreachable(e))) => return Err(e),
+            // The machine answered and declined. The loop is alive and the
+            // state is knowable, but a plan the CLI could not seal is one it
+            // cannot dispatch against, so this is fatal too. The reason names
+            // which rule declined rather than reading as unreachability.
```
MicaiahReid: Who is machine? What loop? I think answering those would help me with this comment

crates/cli/src/cli/simnet/mod.rs
```diff
@@ -255,17 +251,45 @@ pub async fn handle_start_local_surfnet_command(
...
+            Err(StartupPlanFailure::Sealing(SealFailure::Refused(error))) => {
+                return Err(format!("Startup plan refused: {error}"));
             }
         }
-    };
+    } else {
+        // Same policy as the Sealing arm above: an unresponsive command
+        // loop is fatal in both the deploy and no-deploy paths.
```
MicaiahReid:
```suggestion
        // There are no startup tasks to execute, so we must mark our 
        // startup plan as sealed so the listener can mark the simnet `Ready`
```

crates/cli/src/cli/simnet/mod.rs
```diff
@@ -255,17 +251,45 @@ pub async fn handle_start_local_surfnet_command(
...
+        seal_startup_plan(&simnet_commands_tx_copy, vec![])
+            .map_err(|failure| failure.to_string())?;
+    }
+
+    spawn_startup_watchdog(
+        cmd.runtime.daemon || cmd.runtime.no_tui,
+        startup_status_rx,
+        simnet_events_tx.clone(),
+        simnet_commands_tx.clone(),
+    )?;
```
MicaiahReid, line 295:
```suggestion
    
    let is_headless_surfnet = cmd.runtime.daemon || cmd.runtime.no_tui;
    spawn_startup_watchdog(
        is_headless_surfnet,
        startup_status_rx,
        simnet_events_tx.clone(),
        simnet_commands_tx.clone(),
    )?;
```
Naming the bool here helps the reader know the meaning of `cmd.runtime.daemon || cmd.runtime.no_tui`

crates/cli/src/cli/simnet/mod.rs
```diff
@@ -294,11 +318,14 @@ pub async fn handle_start_local_surfnet_command(
 
     let runloop_terminator = Arc::new(AtomicBool::new(false));
 
-    let _ = start_service(
+    // Propagated after the join below: an Aborted event (startup failure
+    // included) must reach the caller so the process exits nonzero, which is
+    // the only failure signal a legacy Anchor readiness loop can perceive.
```
MicaiahReid:
```suggestion
    // service_result contains Aborted event for startup failures, so caller
    // can exit nonzero accordingly
```
Lots of references to "legacy Anchor" in comments which raises two concerns:
 - are we sure this is just legacy anchor's readiness loops? I wasn't aware that code had changed on the anchor side, so I'm curious
 - the caller knowing about failures is relevant with or without Anchor, so no need to mention IMO (I feel like AI will always add context about _this plan it's working on_, when in general it's unnecessary context. Like, this Anchor bit to me makes sense as a comment on the PR to explain how the change pertains to _this fix_, but the code comment should ust explain the code without that context. If the code was to capture anything about a specific fix, that just belongs in a regression test with a reference to the issue and/or plan... Maybe this would be a cool skill to outline for an agent. Anyways /end rant)

crates/cli/src/cli/simnet/startup.rs
```diff
@@ -0,0 +1,593 @@
...
+        .map(|_| ())
+        .map_err(|e| format!("Thread to watch startup status exited: {e}"))
+}
+
+/// Blocks until the startup lifecycle reaches a terminal phase. On `Failed`,
+/// sends `Aborted` first (so the event loop exits with the error before the
+/// shutdown reaches it) and then `Terminate` (so the runloop shuts down
+/// gracefully, WAL checkpoint included). `Ready`, or a dropped sender because
+/// the surfnet is already shutting down, requires nothing.
+pub(super) fn watch_startup_until_terminal(
```
MicaiahReid: I know this is a me problem, but I keep reading "terminal" in this fn name as like, a Terminal (noun, a computer terminal), rather than the state of being terminal (adjective, dead). 

So I'm like "watch startup until terminal what? what will the terminal do???"

How about `watch_startup_until_completion`? `block_on_startup_completion`?

crates/cli/src/cli/simnet/mod.rs
```diff
@@ -579,275 +584,238 @@ fn log_events(
...
+    /// runbook needs, so a project whose `Anchor.toml` or `Cargo.toml` cannot be
+    /// read still starts. Treating that read as fatal would strand a working
+    /// project on a file it does not depend on.
+    #[test]
+    fn an_existing_runbook_does_not_depend_on_framework_detection() {
+        assert!(
+            !RunbookExecutionMode::ExistingOnDisk.requires_framework_detection(),
+            "an on-disk runbook is authoritative; detection only adds clones"
+        );
+    }
```
MicaiahReid: Test belongs in `/startup.rs`

crates/cli/src/cli/simnet/mod.rs
```diff
@@ -579,275 +584,238 @@ fn log_events(
...
+    #[test]
+    fn scaffolded_and_in_memory_runbooks_require_framework_detection() {
+        for mode in [
+            RunbookExecutionMode::ScaffoldOnDisk,
+            RunbookExecutionMode::InMemory,
+        ] {
+            assert!(
+                mode.requires_framework_detection(),
+                "{mode:?} has no runbook without detection"
+            );
```
MicaiahReid: Test belongs in `/startup.rs`

crates/cli/src/cli/simnet/startup.rs
```diff
@@ -0,0 +1,593 @@
...
+    }
+}
+
+/// Spawns the startup watchdog when running headless (`--no-tui` or daemon).
+/// Legacy Anchor's readiness loop can only perceive `completed_at` and
+/// process death (it has no timeout and never reads `errors`), so a headless
+/// surfnet whose startup failed must abort rather than serve forever. The
+/// TUI instead stays alive and displays the failure; killing an interactive
+/// session would be wrong, so in TUI mode this spawns nothing and drops the
+/// receiver.
```
MicaiahReid: Again, not sure if "Legacy Anchor" here is actually specifically to Legacy Anchor vs just Anchor. I could do without the comment, or would appreciate this to be verified at least
```suggestion
/// Spawns the startup watchdog when running headless (`--no-tui` or `daemon`).
/// The watchdog will send a [SimnetCommand::Terminate] if the startup process
/// fails, killing the surfnet. If not headless, no watchdog is spawned; the
/// TUI is responsible for displaying errors.
```

crates/cli/src/cli/simnet/startup.rs
```diff
@@ -0,0 +1,593 @@
...
+        )));
+        let _ = commands_tx.send(SimnetCommand::Terminate(None));
+    }
+}
+
+/// Bounded wait for the seal round-trip. Generous rather than tight: the
+/// runloop emits `SimnetEvent::Ready` immediately before entering the
+/// command loop on the same task, and the CLI waits for Ready before
+/// planning, so the loop is provably alive at every call site. A timeout
+/// here means the loop died or wedged, not that it is merely slow.
```
MicaiahReid: wedged ???
Is this CS terminology I'm not aware of?? 😂

crates/cli/src/cli/simnet/startup.rs
```diff
@@ -0,0 +1,593 @@
...
+    /// must) drive it to Failed; the watchdog policy then decides fatality.
+    Planning(String),
+    /// The seal did not take. Which half failed decides what is knowable
+    /// afterwards, so the two arrive separately.
+    Sealing(SealFailure),
+}
+
+/// The two ways sealing fails, which differ in what they leave known.
+pub(super) enum SealFailure {
+    /// The round trip did not complete: the command loop is dead or wedged,
```
MicaiahReid: wedged

crates/types/src/startup.rs
```diff
@@ -0,0 +1,702 @@
...
+
+use serde::{Deserialize, Serialize};
+
+use crate::types::{GetSurfnetInfoResponse, RunbookExecutionStatusReport};
+
+impl GetSurfnetInfoResponse {
+    /// Runbook ID of the synthetic startup entry in `runbook_executions`.
+    ///
+    /// The value is part of the legacy Anchor wire contract and must not
+    /// change; the constant's name carries no such constraint.
```
MicaiahReid: Questioning the validigy of the "legacy" Anchor wire contract, and that the value must not change. Not seeing this value "surfpool-startup" anywhere in the Anchor source?

crates/types/src/startup.rs
```diff
@@ -0,0 +1,702 @@
...
+/// by the entire clone-and-deploy window.
+///
+/// Each variant's meaning is defined by the phase projection table in
+/// [`SurfnetStartupStatus`]'s documentation.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub enum SurfnetStartupPhase {
+    #[default]
+    Planning,
+    Initializing,
```
MicaiahReid: Am I correct in saying that `Initializing` means that remote accounts are being cloned? If so, should we rename accordingly, e.g. `CloningRemoteAccounts`?

crates/types/src/startup.rs
```diff
@@ -0,0 +1,702 @@
...
+        if !matches!(
+            status.state,
+            SurfnetStartupTaskState::Pending | SurfnetStartupTaskState::Running
+        ) {
+            return false;
+        }
+        status.state = SurfnetStartupTaskState::Failed;
+        status.error = Some(error);
+        true
+    }
```
MicaiahReid, line 564: Doc comments for these transitions would be nice. (When does it return false vs true)

crates/types/src/types.rs
```diff
@@ -430,10 +431,50 @@ pub mod pubkey_option_account_map {
...
     Ready(u64),
+    /// The surfnet answered a client, carrying what the answer conveyed.
+    ///
+    /// Reports what a client was told, so a test can assert on the answer a
+    /// client received rather than infer it from internal state. Issue 715 is
+    /// defined by client observation, which is why the answer is recorded.
+    AnsweredClient {
+        method: &'static str,
+        answer: ClientAnswer,
+    },
```
MicaiahReid: Still not convinced at this one being necessary

crates/types/src/types.rs
```diff
@@ -430,10 +431,50 @@ pub mod pubkey_option_account_map {
...
+    /// Core startup completed: RPC servers are bound and stored transactions
+    /// have been replayed (the payload is their count). This is *not* public
+    /// readiness: with an external startup planner (the CLI), it fires while
+    /// the startup phase is still `Planning`, before clones are hydrated or
+    /// startup runbooks run. Public readiness is
+    /// [`SurfnetStartupPhase::Ready`], observable via
+    /// `surfnet_getSurfnetInfo` or the startup watch channel. With no
+    /// external planner the two coincide: the runloop seals the empty plan
+    /// immediately before emitting this event.
     Ready(u64),
```
MicaiahReid: Considering the comment, should we rename to `RpcReady`? `RpcBound`?

MicaiahReid: I don't think I like `RpcBound`. `SvmInitialized`? Why imply readiness, seems confusing

crates/types/src/startup/spec.rs
MicaiahReid, line 1: I'm a bit confused about this file - it's a test, correct? Or at least, resources for tests? If so, can it be in a tests folder? Perhaps `crates/types/src/tests/startup`?

MicaiahReid: Also, I'm not entirely clear on its value - the state machine _is_ the spec? Why do we need a separate spec?

For example, `expected_phase` here encodes the same logic as `SealedStartupPlan::phase()` - why double the code?

MicaiahReid (changes requested):
Most of these changes are just for comments. This is a big PR that took a lot for me to grok. Perhaps these comment changes are just how they helped me to understand it better as I worked through things. Push back if you think it's worse. But as a whole it feels like there was a good bit of jargon in the comments that it helped me to clarify.

I love how explicit the state machine is. I don't see any issues in terms of what the code actually does, but I'll be spending some time to test things today.

Thanks again @cds-amal!

cds-amal:
> Most of these changes are just for comments. This is a big PR that took a lot for me to grok. Perhaps these comment changes are just how they helped me to understand it better as I worked through things. Push back if you think it's worse. But as a whole it feels like there was a good bit of jargon in the comments that it helped me to clarify.
> 
> I love how explicit the state machine is. I don't see any issues in terms of what the code actually does, but I'll be spending some time to test things today.
> 
> Thanks again @cds-amal!

I haven't reviewed you suggestions yet, but I'm glad the breadcrumb-style, somewhat verbose comments gave you a foothold to grok the changes. Moving to an explicit state machine is a bit of a mental shift, but I found it made the reasoning much simpler once everything had a place in the model. I needed those breadcrumbs too as I bounced between coding and testing; I even had `cargo doc` open in a browser the whole time.

I'll review the comment edits later today and suspect they'll be clearer now that you've walked the path <3.

One thing to weigh while you test: how the machine should be tested now, and as it changes later. I wasn't sure myself, which is why the incident histories live in `startup_regression_scenarios` as named scenarios rather than plain tests: each records a sequence that once misbehaved, so a change that reintroduces the shape fails with the story attached.

crates/cli/src/cli/simnet/mod.rs
cds-amal: I created an issue to understand StartupHooks(#756)

crates/types/src/types.rs
cds-amal: 🤔

crates/cli/src/cli/simnet/mod.rs
cds-amal: Oops, referring the startup state machine.

A wedged system is alive but can't make forward progress because some component, resource, or comms path it depends on is blocked. Here it describes the command loop, not the state machine: the loop is dead or wedged, so the state machine is unreachable. Its state is therefore unknowable, and no session can become ready.

That's why the variant is `Unreachable` rather than `Wedged`. The state machine may be perfectly healthy; we simply can't reach it to find out.

crates/core/src/rpc/surfnet_cheatcodes.rs
cds-amal: Good catch! This is left over from the abandoned VCR (recording engine).

crates/types/src/types.rs
cds-amal: I went with `CoreStarted`

crates/types/src/startup/spec.rs
cds-amal: Fair questions. I want to argue for keeping it because I expect this machine to evolve, and the spec is what makes those changes safe.

`spec.rs` is the canonical specification. The implementation, tests, and generated documentation all relate back to it, but they each serve different purposes. I don't think of `spec.rs` as test code. I think of it as the separation between design, implementation, and verification.

- The specification (`spec.rs`) expresses the intended behavior in a declarative form.
- The implementation (`startup.rs`) is optimized for executing transitions.
- Tests are then written against the specification rather than against the
  implementation.

Three things read from that specification:

* the generated lifecycle documentation
* the verification tests
* anyone trying to understand the **intended** behavior

Keeping those consumers reading from the same spec means they evolve together, while remaining independent of the implementation. If we instead derive the tests or docs from the implementation, they become much less useful at detecting unintended behavior changes, because they change with the code they are meant to validate. 

For a state machine, whose behavior is a set of rules rather than a code path, this is the model-based testing arrangement: an independent model of the rules, and a check that the implementation agrees with it.

That is also why I would rather keep it beside the state machine than under `tests/`. It is not test scaffolding; it is a first-class specification that happens to be consumed only by tests and documentation, so `#[cfg(test)]` is enough to keep it out of the shipped code.

Perhaps a name-change to `startup/specification.rs` ?

crates/types/src/types.rs
cds-amal: You're right! 86'd

crates/types/src/startup/spec.rs
MicaiahReid: > Tests are then written against the specification rather than against the
implementation

Are there tests that ensure that the implementation matches the spec?

crates/types/src/startup.rs
```diff
@@ -0,0 +1,714 @@
...
+        if self
+            .tasks
+            .iter()
+            .any(|status| status.state == SurfnetStartupTaskState::Failed)
+        {
+            SurfnetStartupPhase::Failed
+        } else if self
+            .tasks
+            .iter()
+            .all(|status| status.state == SurfnetStartupTaskState::Succeeded)
```
cds-amal, line 470: ## Mutation testing only

```suggestion
            .any(|status| status.state == SurfnetStartupTaskState::Succeeded)
```

Then run `cargo test -p surfpool-types every_reachable` . 
This is the original error class.

crates/types/src/startup/spec.rs
cds-amal: tl;dr; Yes, the sweep is specifically what enforces implementation matches the specification.

Prod and tests exercise the same transition function, `apply`. The sweep enumerates ever reachable state and command, verifying `apply` accepts exactly the legal transitions the spec permits (and rejects everything else).

The quickest way to see if the implementaiton / spec works is to break the state-machine and run the sweep function. It fails with a description that shows where the impl diverges from the spec.

crates/types/src/startup.rs
cds-amal: @MicaiahReid , this is a test you can run locally. Here's the result from my mutation run.

```console
git --no-pager diff
```
```diff
@@ -467,7 +467,7 @@ impl SealedStartupPlan {
         } else if self
             .tasks
             .iter()
-            .all(|status| status.state == SurfnetStartupTaskState::Succeeded)
+            .any(|status| status.state == SurfnetStartupTaskState::Succeeded)
         {
             SurfnetStartupPhase::Ready
         } else if self.tasks.iter().any(|status| {
```

The sweep refuses it:

<details><summary>:point_right: <code>cargo test -p surfpool-types every_reachable</code>:point_left:  (Click to see result) </summary>

```rust
running 1 test
test startup::surfnet_startup_reachability_tests::every_reachable_state_upholds_the_startup_invariants ... FAILED

---- startup::surfnet_startup_reachability_tests::every_reachable_state_upholds_the_startup_invariants stdout ----

thread '...' panicked at crates/types/src/startup/surfnet_startup_reachability_tests.rs:125:5:
assertion `left == right` failed: phase disagrees with the spec oracle: Sealed(SealedStartupPlan { tasks: [SurfnetStartupTaskStatus { task: RunbookExecutions, state: Succeeded, error: None }, SurfnetStartupTaskStatus { task: RemoteAccounts, state: Pending, error: None }] })
  left: Ready
 right: CloningRemoteAccounts

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 71 filtered out; finished in 0.00s
```

</details>

Reading the result:

- The sweep reached a sealed two-task plan with `RunbookExecutions:
  Succeeded` and `RemoteAccounts: Pending`.
- `left: Ready` is the mutated machine: one task succeeded, so `any` is
  satisfied.
- `right: CloningRemoteAccounts` is the spec oracle: by the projection
  table, a sealed plan whose clone task is unfinished is still cloning.

The disagreement is exactly the incident that opened #715: the machine wrongly says ready while remote accounts are still being hydrated. The sweep catches the mismatch with what the spec declared.

MicaiahReid (approved):
I'm good to go on this, thanks @cds-amal!!

I've built locally and tested some deployments and clones from an Anchor.toml manually, and tested failed deployments/clones not interfering with startup.

I also ran the `surfpool-anchor-smoke` skill, which runs the entire anchor test suite against a local surfpool build, and things worked as expected.

Great work!

PR #734, opened 2026-08-03
fix(sdk-node): record the platform package versions in the lockfile

cds-amal:
`npm ci` rejects the lockfile on `main` because it is out of sync with `package.json`. Issue #728 added bare `{"optional": true}` stubs for the three platform packages to unblock `npm publish`, but entries without a version cannot satisfy the `optionalDependencies` version pins.

Replace the stubs with complete `1.5.0` lockfile entries. Each integrity hash matches the registry tarball, `npm ci` accepts the lockfile again, and installs the appropriate platform package.

### DIFF
--- a/crates/sdk-node/package-lock.json
+++ b/crates/sdk-node/package-lock.json
@@ -1025,6 +1025,54 @@
         }
       }
     },
+    "node_modules/@solana/surfpool-darwin-arm64": {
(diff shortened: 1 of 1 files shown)


MicaiahReid:
Closing since I beat you to it! `npm ci` is passing now :)

PR #736, opened 2026-08-04
fix(core): report the LiteSVM-computed fee instead of 5000 × signatures

serejke:
`TransactionWithStatusMeta::new` and `::from_failure` hardcode `fee = 5000 * signatures`, ignoring the priority fee LiteSVM computes and exposes as `TransactionMetadata.fee`.

Any transaction that sets a compute-unit price therefore reports a fee that contradicts its own pre/post balances: the payer is debited base + priority, but `meta.fee` shows only the base. Failed transactions are affected too, since they still pay fees.

Both paths now read the fee from the metadata.

Regression test: a transfer with `set_compute_unit_price(1_000_000)` over a 100,000 CU limit. `meta.fee` reports 5,000 before the fix and 105,000 after, matching the payer's balance delta.

### DIFF
--- a/crates/core/src/surfnet/locker.rs
+++ b/crates/core/src/surfnet/locker.rs
@@ -4699,6 +4699,79 @@ mod tests {
     }
 
     #[tokio::test(flavor = "multi_thread")]
+    async fn test_transaction_fee_includes_priority_fee() {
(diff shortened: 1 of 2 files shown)


MicaiahReid (approved):
Great find, and great PR as always, @serejke!

PR #737, opened 2026-08-04
fix: get account with base64+zstd

Arrowana:
This should work out of the box, that's also the rust `solana-rpc-client` default

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,7 +108,9 @@ serde_with = { version = "3.14.0", default-features = false }
 strum = { version = "0.26", default-features = false, features = ["derive"] }
 solana-account = { version = "4.3", default-features = false }
 solana-account-decoder = { version = "4.0", default-features = false }
-solana-account-decoder-client-types = { version = "4.0", default-features = false }
+solana-account-decoder-client-types = { version = "4.0", default-features = false, features = [
+    "zstd",
+] }
 solana-address-lookup-table-interface = { version = "3.0", default-features = false }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -108,7 +108,9 @@ serde_with = { version = "3.14.0", default-features = false }
 strum = { version = "0.26", default-features = false, features = ["derive"] }
 solana-account = { version = "4.3", default-features = false }
 solana-account-decoder = { version = "4.0", default-features = false }
-solana-account-decoder-client-types = { version = "4.0", default-features = false }
+solana-account-decoder-client-types = { version = "4.0", default-features = false, features = [
+    "zstd",
```
MicaiahReid, line 112: I removed this feature and the test above still passed. Are we sure it's needed?

MicaiahReid (changes requested):
Thanks for the PR, @Arrowana! Was this an issue you had? It's great to have the test, but it passes without the associated dep change, so I want to be sure we're fixing any underlying issue you had

Cargo.toml
Arrowana: Both `solana-account-decoder` and solana-rpc-client already enable `solana-account-decoder-client-types/zstd`

https://github.com/anza-xyz/agave/blob/d32fc31eb1717f534466fca42eea7848e335ef4d/account-decoder/Cargo.toml#L30

so yes that's indeed incorrect, i think the symptoms are similar to #738 but the root cause was misdiagnosed wildly

PR #750, opened 2026-08-07
docs(rpc): make the documented surface match the wire

92Infinitus92:
Four corrections to documentation a client can follow into failure, each checked against a running surfnet.

`surfnet_timeTravel` documented neither its units nor its direction: `absoluteTimestamp` is milliseconds while the Clock sysvar reports seconds, and the clock only moves forward, so a value read from the clock and passed back is rejected as a past target.

Six field names were documented in snake_case while the wire is camelCase. `fetch_before_use` is the harmful one — it is `#[serde(default)]`, so posting an override the way the docs spelled it returns HTTP 200 and stores `fetchBeforeUse: false`, which means no fresh fetch and a skipped override when the account is not already in the SVM.

The `surfnet_registerScenario` schema omitted the required `tags` (a request following it fails with `missing field tags`) and typed `account` as a string, though `AccountAddress` needs `{"pubkey": "…"}` or `{"pda": {…}}`. The clock cheatcodes were missing from the catalog entirely, although the MCP tool description points models at that resource as the list of available methods, and their `EpochInfo` examples showed snake_case.

### DIFF
--- a/crates/core/src/rpc/surfnet_cheatcodes.rs
+++ b/crates/core/src/rpc/surfnet_cheatcodes.rs
@@ -735,12 +735,15 @@ pub trait SurfnetCheatcodes {
         limit: Option<u64>,
     ) -> BoxFuture<Result<RpcResponse<Vec<RpcLogsResponse>>>>;
 
-    /// A cheat code to jump forward or backward in time on the local network.
+    /// A cheat code to jump forward in time on the local network. The clock only moves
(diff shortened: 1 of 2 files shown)


MicaiahReid (approved):
This looks good, but to clarify, these doc comments don't directly affect any of the online docs. Do these also need to be updated on our doc site?

PR #752, opened 2026-08-07
feat(mcp): compact the template payload and add a constant search tool

92Infinitus92:
`get_override_templates` inlined every constant option. The verified-token list alone is 3920 entries, so the payload runs to megabytes — enough to exhaust an LLM client's context and rate limits before a single scenario is authored.

Options are now summarised to `optionsCount`, and a new `search_constant_options` tool resolves concrete values: case-insensitive matching over option id, label, description and value, capped at 20 results with `totalMatches` and `truncated` so a model knows to narrow its query. Constants iterate in sorted order, so paging is stable. The same compaction is shared with the `str:///override_templates` resource, which previously still inlined everything.

The tool takes `templateId`, matching the name its own response returns. Six tests cover the error paths, case-insensitivity, truncation, and that no compacted payload inlines options.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12081,6 +12081,7 @@ dependencies = [
  "spl-token-interface 2.0.0",
  "surfpool-core",
  "surfpool-types",
+ "tokio",
  "tracing",
 ]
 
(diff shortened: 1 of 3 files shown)


MicaiahReid (approved):
lgtm!

Issue #706, opened 2026-06-27
Confidential-transfer DX: cheatcodes to close the ZK test loop (inspect, derive keys, mints, diagnostics)


lgalabru:
## Summary

Confidential transfers (Token-2022) are some of the hardest things to build and test on Solana: balances are encrypted, account setup is a 3+ transaction `Configure → Deposit → ApplyPendingBalance` dance that requires the ZK ElGamal Proof program, and failures surface as opaque `InvalidInstructionData`. Surfpool is uniquely positioned to make this painless via cheatcodes.

#705 added the first piece — `surfnet_setTokenAccount` can now configure (and optionally fund) a confidential account directly. This issue tracks the follow-ups that would **close the confidential test loop entirely inside surfpool**, so a builder can set up, act, and assert on confidential state without pulling in an external SDK or hand-rolling ElGamal/AES crypto.

These leads come from driving a real confidential transfer end-to-end against a local surfnet (configure both parties → fund sender → `TransferWithFee` → finalized, recipient credited). Each one maps to a concrete wall I hit.

---

### 1. `surfnet_getConfidentialBalance` — decrypt & inspect  *(highest leverage)*
The missing half of #705. You can now *write* a confidential balance, but to *verify* a transfer there's no way to read it — I had to fall back to checking `pendingBalanceCreditCounter` and "is the ciphertext non-zero" instead of asserting "recipient received exactly 10."

A cheatcode that takes an account + the owner's keys (and optionally the auditor key) and returns the **decrypted** available + pending amounts turns confidential test assertions into `assert_eq!(balance, 10)`.

- **Scope:** small. Reuses the `solana-zk-sdk` crypto already wired into `crates/core/src/types.rs` by #705. AES fast-path for `decryptable_available_balance`; ElGamal `decrypt_u32` over the `lo`/`hi` split for pending.
- **Bonus:** an `as: "auditor"` mode (decrypt with the mint's auditor ElGamal key) demonstrates the compliance/oversight story.

### 2. `surfnet_deriveConfidentialKeys` — make the cheatcodes self-contained  *(cheap, synergistic)*
To even *call* the confidential `setTokenAccount` today, you must derive `elgamalPubkey` / `aesKey` somehow — which currently means depending on an external client (e.g. pay-kit). A tiny RPC that derives them from a keypair (or a signature) + token-account seed makes the whole confidential cheatcode suite usable by any builder with **zero crypto dependencies**.

[...]

PR #755, opened 2026-08-07
feat(core): confidential balance read and key derivation cheatcodes

michael-moffett:
## Summary

Covers leads 1 and 2 of #706. #705 made it possible to *write* a confidential balance
with `surfnet_setTokenAccount`, but there was still no way to read one back, so a
confidential-transfer test has to fall back on `pendingBalanceCreditCounter` and "is the
ciphertext non-zero" rather than asserting on an amount. These two cheatcodes close that
half of the loop.

Leads 3 (`surfnet_setMint` with confidential config) and 4 (simulation-time diagnostics)
are not attempted here.

## `surfnet_getConfidentialBalance`

Takes a token account plus the owner's keys and returns the decrypted balances:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "surfnet_getConfidentialBalance",
  "params": [
    "4EXSeLGxVBpAZwq7vm6evLdewpcvE2H56fpqL2pPiLFa",
    { "aesKey": "<base58, 16 bytes>", "elgamalSecretKey": "<base58, 32 bytes>" }
  ]
}
```

```json
{ "available": 10, "pending": 0, "pendingBalanceCreditCounter": 0 }
```

The two balances need different keys, so both are independently optional and a caller
holding only one still gets back the half it can read:

- **available** is read from `decryptable_available_balance` with the AES key. That is the
  fast path the issue describes, and it is also the only path: the authoritative
  `available_balance` is an ElGamal ciphertext over a full u64, and `decrypt_u32` recovers
  only values below 2^32.
- **pending** is read from the `lo` / `hi` ElGamal ciphertexts with the ElGamal *secret* key
  (not the public key stored on the account). Each half is inside the u32 decode range;
  they are recombined as `hi << PENDING_BALANCE_LO_BIT_LENGTH | lo`.

`pendingBalanceCreditCounter` comes back alongside them so that a zero pending balance is
distinguishable from a missing `ApplyPendingBalance` without a second call.

Supplying neither key is rejected rather than returning two nulls, and a wrong `aesKey` fails
the ciphertext's authentication tag rather than quietly decrypting to another number. A wrong
`elgamalSecretKey` also fails against a non-zero pending balance, but not against a zero one:
the all-zero ciphertext a freshly configured account carries is the identity point, and it
decodes to 0 under any key.

## `surfnet_deriveConfidentialKeys`

Takes two signatures and returns base58 `elgamalPubkey` (for `surfnet_setTokenAccount`),
`elgamalSecretKey` (for `surfnet_getConfidentialBalance`), and `aesKey` (for both), so the
confidential cheatcodes can be driven with no client-side crypto dependency.

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "surfnet_deriveConfidentialKeys",
  "params": [
    "<base58, 64-byte signature over \"ElGamalSecretKey\" || token account>",
    "<base58, 64-byte signature over \"AeKey\" || token account>"
  ]
}
```

No keypair crosses the wire. The caller signs the two seed messages locally and sends only
the signatures, so a hardware signer that never exposes its signing key can drive this.
Two signatures rather than one because the SDK domain-separates the two seed messages, and
a single signature cannot produce both keys.

Derivation goes through `ElGamalKeypair::new_from_signature` and `AeKey::new_from_signature`
over the seed `token_account.as_ref()`, with no prefix. A test asserts the derived keys are
byte-identical to what `new_from_signer` produces for the same owner and account, so these
are the keys a reference client derives, not a look-alike. The all-zero signature is
rejected, matching the check both `seed_from_signer` paths carry in the SDK.

Worth stating plainly, because the wire shape can suggest more than it delivers: because the
ElGamal secret is a hash of the signature, `elgamalSignature` is exactly as sensitive as the
`elgamalSecretKey` it derives. What changed is that the material on the wire no longer
confers transaction-signing power, only confidential-balance decryption. The derived
`elgamalSecretKey` and `aesKey` do still travel back in the response, and
`surfnet_getConfidentialBalance` takes them back as inputs. That is what removes the
client-side crypto dependency, and it also means these keys are only as private as the RPC
connection. A simnet testing convenience, not a key-management pattern for a live cluster.

## Tests

Three round trips run against a live surfnet, plus unit coverage on the derivation and the
decrypt path.

`test_confidential_balance_deposit_round_trip` is the assertion the issue asks for: derive,
set, deposit, read, with no crypto in the test itself.

```rust
let balance = result.available.expect("available balance should decrypt");
assert_eq!(balance, 10);
```

The doc comments on each test name which assertions bind to the derived keys and which do
not, because the distinction is not obvious. A `Deposit` adds to the commitment and passes
the decrypt handle through untouched, so a deposit onto the all-zero pending ciphertext a
freshly configured account carries opens to the same value under any secret key. That test
seeds the pending balance with a real encryption first, so the assertion binds; a random
ElGamal secret key is asserted to recover nothing.

Also covered: the `lo` / `hi` recombination against a pending balance that spans both halves,
that derived keys are deterministic and scoped per token account, that signature-derived keys
match signer-derived keys byte for byte, and the rejection cases above.

### The confidential `Transfer` leg is not covered

`test_confidential_balance_transfer_round_trip` builds a real confidential `Transfer` with
all three proofs and ships `#[ignore]`d. It does not run, and nothing else in this change
asserts that a confidential `Transfer` completes. Treat that path as uncovered rather than
working.

The reason is a version skew in the dependency graph rather than anything about the
cheatcodes. Proof *generation* here resolves `solana-zk-sdk` 4.0.0 through
`spl-token-2022-interface` and `spl-pod`, while the proof *verifier* the runtime runs is
5.0.1, and transcript construction moved layer between those majors. The instruction builds
and Token-2022 accepts it; the proof program rejects it with
`SigmaProof(Equality, AlgebraicRelation)`. Nothing published aligns the two sides, so the
test carries that reason in its doc comment rather than being deleted or weakened into
passing. Un-ignore it once the two agree on a major, either `litesvm` linking a 4.x proof
program or `spl-token-2022-interface` moving to the 5.x proof stack.

## Other files touched

`SURFNET_CHEATCODE_METHODS` gains the two method names, which the manifest guard test
requires. The kit bindings under `crates/sdk-node/surfpool-sdk/kit/generated` are
regenerated from it, and `kit/types/api.ts` gains the two matching entries that its
compile-time manifest-coverage assertion requires.

The kit type generator now declares `prettier` as a dependency and runs it over `index.ts`
along with the per-type files, so its output is formatted at the source and stays formatted
across regenerations.


crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -20,7 +20,11 @@ use surfpool_types::{
     GetStreamedAccountsResponse, GetSurfnetInfoResponse, Idl, OfflineAccountConfig,
     ResetAccountConfig, RpcProfileResultConfig, Scenario, SimnetCommand, SimnetEvent,
     StreamAccountConfig, StreamAccountsEntry, UiKeyedProfileResult,
-    types::{AccountUpdate, SetSomeAccount, SupplyUpdate, TokenAccountUpdate, UuidOrSignature},
+    types::{
```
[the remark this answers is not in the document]
MicaiahReid: bad bot

crates/sdk-node/surfpool-sdk/kit/generated/ConfidentialBalanceKeys.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * The owner's AES key (base58 or base64, 16 bytes). Decrypts the available
+ * balance.
+ */
+aesKey?: string, 
+/**
+ * The owner's ElGamal *secret* key (base58 or base64, 32 bytes) — not the
+ * public key stored on the account. Decrypts the pending balance.
+ */
+elgamalSecretKey?: string, };
```
MicaiahReid, line 22: Formatting here is weird

crates/sdk-node/surfpool-sdk/kit/generated/DeriveConfidentialKeysResponse.ts
```diff
@@ -0,0 +1,23 @@
...
+elgamalPubkey: string, 
+/**
+ * The ElGamal secret key — `surfnet_getConfidentialBalance`'s
+ * `elgamalSecretKey`.
+ */
+elgamalSecretKey: string, 
+/**
+ * The AES key — the `aesKey` of both of the above.
+ */
+aesKey: string, };
```
MicaiahReid, line 23: Formatting here is weird

crates/sdk-node/surfpool-sdk/kit/generated/GetConfidentialBalanceResponse.ts
```diff
@@ -0,0 +1,24 @@
...
+ * The pending (credited but not yet applied) balance, or `null` if no
+ * `elgamalSecretKey` was supplied.
+ */
+pending: number | bigint | null, 
+/**
+ * How many confidential credits are sitting in the pending balance. Non-zero
+ * means an `ApplyPendingBalance` is required before they show up in
+ * `available`.
+ */
+pendingBalanceCreditCounter: number | bigint, };
```
MicaiahReid: Formatting here is weird

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -1198,6 +1202,133 @@ pub trait SurfnetCheatcodes {
...
+    ///
+    /// # Notes
+    /// The derivation matches what a confidential-transfer client computes from the same keypair, so
+    /// keys produced here interoperate with keys derived off-chain. The keypair is used only to sign
+    /// the derivation seeds; it is not stored.
+    #[rpc(meta, name = "surfnet_deriveConfidentialKeys")]
+    fn derive_confidential_keys(
+        &self,
+        meta: Self::Metadata,
+        keypair: String,
```
MicaiahReid: Tbh I'm a bit wary of training our users to ever send a keypair over the wire.

I see there is an `ElGamalKeypair::new_from_signature` function - would you be able to experiment with accepting a signature instead? I've never tried it, I'm not sure how it works or if you get the same results, or what kind of signature you need to provide.

Let me know what you think!

MicaiahReid (changes requested):
Thanks @michael-moffett, great stuff!

MicaiahReid (changes requested):
Also, could we get some integration tests?

A round trip of derive keys -> confidential transfer -> get new balance would be a cool test

crates/core/src/types.rs
```diff
@@ -1339,6 +1345,118 @@ pub fn build_confidential_token_account_data(
...
+pub fn derive_confidential_keys(
+    keypair: &str,
+    token_account: &Pubkey,
+) -> Result<DeriveConfidentialKeysResponse, String> {
+    let keypair_bytes =
+        decode_confidential_key(keypair, 64).map_err(|e| format!("keypair: {e}"))?;
+    let keypair = Keypair::try_from(keypair_bytes.as_slice())
+        .map_err(|e| format!("keypair: invalid Solana keypair ({e})"))?;
+
+    let public_seed = [CONFIDENTIAL_KEY_SEED_PREFIX, token_account.as_ref()].concat();
```
MicaiahReid: If I understand correctly, we're doing a non-standard seed here by adding `CONFIDENTIAL_KEY_SEED_PREFIX` (rather than just `token_account.as_ref()`. I think that would create some compatibility issues with keys derived here vs those already on-chain

crates/core/src/rpc/surfnet_cheatcodes.rs
michael-moffett: It works, and you get the same results. Two signatures, because the SDK domain-separates the two seed messages, so one signature can't produce both keys:

- `elgamalSignature`: sign `"ElGamalSecretKey" || token_account`
- `aeSignature`: sign `"AeKey" || token_account`

Each is a normal 64-byte ed25519 signature from the owner, sent base58 or base64. The signing key never leaves the client, so a hardware signer can drive it.

On "the same results": `signatures_reproduce_the_keys_the_signer_path_derived` asserts the derived keys are byte-identical to what `new_from_signer(&owner, token_account.as_ref())` produces for the same owner and account. The all-zero signature is rejected, matching the check both `seed_from_signer` paths carry.

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -1213,6 +1217,165 @@ pub trait SurfnetCheatcodes {
...
+    /// `"AeKey" || token_account` here yields byte-identical keys to
+    /// `ElGamalKeypair::new_from_signer(&owner, token_account.as_ref())` and
+    /// `AeKey::new_from_signer(&owner, token_account.as_ref())`. An account configured off-chain by a
+    /// client on that derivation opens with these keys. The one input where the two paths would
+    /// otherwise part is the all-zero default signature, which `new_from_signer` refuses as key
+    /// material and `new_from_signature` would hash into a publicly computable key pair; this
+    /// method rejects it too, so the equivalence holds without exception.
+    ///
+    /// That SDK marks the KDF behind these functions as non-standard and intends to replace it, so
+    /// what these keys interoperate with is the SDK version, not a frozen standard.
```
MicaiahReid: I think this whole comment can be removed, wdyt?

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -1213,6 +1217,165 @@ pub trait SurfnetCheatcodes {
...
+    ///
+    /// The owner's *signing* key never crosses the wire: the caller signs the two seed messages
+    /// locally and sends only the signatures, so a hardware signer, which never exposes its signing
+    /// key, can drive this. The signatures are used only as key material and are not stored.
+    ///
+    /// What that does and does not buy is worth stating plainly. Because the ElGamal secret is a
+    /// hash of the signature, `elgamalSignature` is exactly as sensitive as the `elgamalSecretKey`
+    /// it derives: anyone who sees it recomputes that key offline. What changed is that the
+    /// material on the wire no longer confers transaction-signing power, only confidential-balance
+    /// decryption; it is not that nothing sensitive is transported.
```
MicaiahReid: The "what that does and does not buy" is helpful.

However, the "What changed is that" isn't - it's referencing an earlier reviesion of this same PR

crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -2236,6 +2399,74 @@ impl SurfnetCheatcodes for SurfnetCheatcodesRpc {
...
+
+        Box::pin(async move {
+            let SvmAccessContext {
+                slot,
+                inner: account_result,
+                ..
+            } = svm_locker
+                .get_account(&remote_ctx, &token_account, None)
+                .await?;
+            svm_locker.write_account_update(account_result.clone());
```
MicaiahReid: This PR will need a rebase - this function doesn't exist anymore. You can remove this call, `get_account` will now automatically update from mainnet after fetch!

MicaiahReid (changes requested):
This is awesome and really close, @michael-moffett, thanks for the clean up!

MicaiahReid (approved):

PR #759, opened 2026-08-12
fix(core): impl StorageBackend to fix sqlite connection leaks

cds-amal:
This PR fixes a SQLite connection leak that presents with exhausted file-descriptors over the course of multiple surfnet lifetimes, when the number of file-descriptors open crosses the system limit and file opens fail. This happens on stock linux machines under `ulimit -Sn 1024`.

Every SQLite store builds or borrows its pool from a process-global cache. A process accumulates connections for each surfnet because its associated connections are owned by the process-global cache. Further, every in-memory SQLite backed surfnet built 19 separate pools; one per store, each a cache miss (19 x 10).

The fix is to let `StorageBackend` give each surfnet an owner for its database connections, specifically one pool, built when the surfnet's storage initializes, shared by every store the surfnet opens. Stores hold pool clones so the pool-clones are bound to the store's lifetime, and when the `StorageBackend` goes away, it will release the one pool it owns; RAII ftw.

PostgreSQL holds a lease on the process-level pool for its database URL. Unlike SQLite, the pool stays shared across surfnets, since pooling exists to amortize the network connection and the server caps total sessions.

## Measurements

Counter movement for ten surfnets built and dropped per phase, recorded by a temporary in-process census that this branch carried for the measurement and removes before merge (the history has it, the squash lands only the fix):

| counter                                       | baseline                  | this branch |
|-----------------------------------------------|--------------------------:|------------:|
| on-disk: pools created                        | 10, plus 180 cache reuses | 10          |
| on-disk: connections opened / closed          | 100 / 0                   | 100 / 100   |
| on-disk: connections live after the last drop | 100                       | 0           |
| in-memory: pools created                      | 190                       | 10          |
| in-memory: connections opened / closed        | 1,900 / 1,900             | 100 / 100   |
| peak live connections                         | 290                       | 10          |

The suite-level effect: `cargo test -p surfpool-core --lib --features
ignore_tests_ci` under `ulimit -Sn 1024`, same machine, same session,
both builds:

|                           | baseline               | this branch          |
|---------------------------|------------------------|----------------------|
| result                    | 536 passed, 105 failed | 642 passed, 0 failed |
| wall clock                | 204.7s                 | 36.1s                |
| peak threads (3s samples) | 4,954                  | 1,079                |
| peak fds (3s samples)     | 1,167                  | 269                  |

- The wall-clock improvement is a consequence I didn't chase (number go down)
- The residual 1,079 threads belong to r2d2 per-pool schedulers; I didn't go
  into that API, the ~5x drop was enough for this PR.

## Verifying from outside the process

The on-disk result reproduces externally, with no code from this branch: the workload is 31 pre-existing tests that each build and drop an on-disk surfnet.

```sh
BIN=$(cargo test -p surfpool-core --lib --no-run --message-format=json 2>/dev/null \
  | jq -r 'select(.reason=="compiler-artifact" and .executable != null) | .executable')
"$BIN" with_on_disk_sqlite_db --skip integration --test-threads=1 &
PID=$!
while kill -0 $PID 2>/dev/null; do
  # Linux; on macOS use: lsof -p $PID | grep -c sqlite
  find /proc/$PID/fd -lname '*sqlite*' 2>/dev/null | wc -l
  sleep 0.2
done
```

On main the count climbs a staircase to 434 and holds it until process exit; on this branch it peaks at 23 (one live surfnet's worth) and returns to the floor between tests. In-memory databases hold no descriptors, so that half of the improvement is visible only in the counter table above.


MicaiahReid (approved):
Nice, thanks @cds-amal!

PR #760, opened 2026-08-13
refactor: route simnet events through a sender that owns delivery policy

cds-amal:
While working #731 I kept meeting the same statement, at over a hundred call sites across core and cli:

```rust
let _ = svm_locker
    .simnet_events_tx()
    .try_send(SimnetEvent::error(format!("Failed to seal startup plan: {e}")));
```

Each site combines three decisions:

1. the transport (`try_send` vs `send`, so whether the caller may block);
2. the failure policy (`let _ =`, so what a refused event costs); 😰 
3. the construction (`SimnetEvent::error(format!(...))`, the wire shape
   inline).

The channel is `crossbeam_channel::bounded(1024)`, so on a full buffer `try_send` fails and `let _ =` discards the failure. The result is two policies nobody chose: core drops events silently under backpressure, cli blocks the sending thread. `Aborted`, `Shutdown`, and `CoreStarted` go through the same silent drop as log lines; a frontend that misses one holds a wrong model of the surfnet.

### Decision table

Since the review is circling the delivery contract, here it is as a decision table: given a delivery policy and the channel's state, each cell is what happens to the event and to the sender. `log` sacrifices the event under contention; `emit` sacrifices the sender's progress. Neither decision is made at the call site, which is what the type brings to the design. 

<table>
  <thead>
    <tr>
      <th rowspan="2">Delivery policy</th>
      <th colspan="3">Channel state</th>
    </tr>
    <tr>
      <th>buffer has room</th>
      <th>buffer full</th>
      <th>receiver disconnected</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td><code>log</code></td>
      <td>deliver</td>
      <td>drop the incoming event, by contract</td>
      <td>drop (no audience)</td>
    </tr>
    <tr>
      <td><code>emit</code></td>
      <td>deliver</td>
      <td>block until the reader drains</td>
      <td>drop (no audience)</td>
    </tr>
  </tbody>
</table>

<details><summary> 👉  How to Read it 👈 </summary>

Pick the policy row, then the channel-state column. The contrast cell pair: `log` on a full buffer drops the incoming event, while `emit` on the same full buffer blocks the sender until the reader drains. Same channel state, different action; the delivery policy is the only input that changed.

Two rules sit outside the grid:

1. **Lock context, a legality rule rather than a runtime cell:** `emit`
   while holding the SVM guard is illegal. The reader may need that lock
   to drain, so blocking on a full channel while holding the guard
   deadlocks the very reader that would make progress possible; exactly
   the deadlock the measurement tests caught. The
   classification rule for call sites is therefore lock context, not event
   importance: anything sent while holding the SVM guard uses `log`.
2. **The unbounded constructor collapses the table:** in buffering contexts
   (the bundle sandbox), the "buffer full" column is unreachable and `log`
   and `emit` coincide. The distinction only exists on a bounded channel,
   which is where every long-lived reader lives.

The type owns the delivery policy; channel state determines its consequence; lock context determines which policy is legal.

</details>

## The change

`SimnetEventsTx` (in `surfpool-types`, next to `SimnetEvent`) owns the delivery policy:

- `log(event)` is lossy by contract: telemetry may drop under backpressure.
  `info` / `warn` / `error` / `debug` conveniences construct and log in one
  call.
- `emit(event)` is lossless: lifecycle events block until the reader drains.

Call sites state intent only:

```rust
// telemetry, may drop under load
svm_locker.simnet_events_tx().error(format!("Failed to seal startup plan: {e}"));

// lifecycle, may not be lost
simnet_events_tx.emit(SimnetEvent::Aborted(e.to_string()));
```

The classification is auditable: `grep 'emit('` lists every send that claims lossless-ness (17 sites), and a reviewer can dispute any claim.

One rule bounds the classification: `emit` is never called while holding the SVM lock, because the reader may need that lock to drain, and a blocked send under the guard wedges every thread behind it (the test suite demonstrates this: an `emit` of `SystemClockUpdated` from `confirm_current_block` under the write guard deadlocked the full run). Guard-context sends use `log`; readers recover missed state from the SVM, not the event stream. The rule is documented on `emit`.

## Observed Before / after

The first commit adds tests against the real `SurfnetSvm` channel and pins the old behavior; the last state asserts the contract. Both runs 3/3.

| behavior                         | before                                | after                                                            |
|----------------------------------|---------------------------------------|------------------------------------------------------------------|
| 1,025th event on a full buffer   | silently refused                      | `log`: refused by declared contract; `emit`: blocks, delivers    |
| `Aborted` behind 1,024 log lines | dropped, indistinguishable from noise | delivered; queued behind the noise, received as event 1,025      |
| 4,096-event log burst            | 3,072 lost, newest first (accident)   | 3,072 shed, newest first (contract; lifecycle goes through emit) |

## Behavior changes

- Lifecycle events sent from runloop level and cli (`Connected`,
  `EpochInfoUpdate`, `CoreStarted`, `ClockUpdate`, `Aborted`, `Shutdown`,
  `RunbookStarted`, `RunbookCompleted`) are now lossless.
- cli's blocking sends of plain log lines become droppable: a slow TUI reader
  sheds telemetry instead of stalling the sending thread.
- Guard-context sends keep their lossy behavior, now stated as `log` instead
  of implied by `let _ = try_send`.

## Breaking change

`SurfnetSvm.simnet_events_tx` is a `pub` field and its type changes from `Sender<SimnetEvent>` to `SimnetEventsTx`, a semver major on `surfpool-core`.

## Reference

The sender-owns-policy shape follows Stefan Baumgartner's "Refactoring in Rust" (https://www.youtube.com/watch?v=wuBkzT_3CDU): mint a type that names the thing, let methods carry the policy, and call sites state intent.

## Testing

1. `cargo test -p surfpool-types simnet_events_tx`
1. `cargo test -p surfpool-core --lib simnet_events`

### DIFF
--- a/crates/cli/src/cli/simnet/mod.rs
+++ b/crates/cli/src/cli/simnet/mod.rs
@@ -80,15 +80,15 @@ pub async fn handle_start_local_surfnet_command(
                 &cmd.observability.metrics_addr,
             ) {
                 Err(e) => {
-                    let _ = surfnet_svm
+                    surfnet_svm
                         .simnet_events_tx
-                        .send(SimnetEvent::warn(format!("Metrics init failed: {}", e)));
+                        .warn(format!("Metrics init failed: {}", e));
                 }
                 Ok(_) => {
-                    let _ = surfnet_svm.simnet_events_tx.send(SimnetEvent::info(format!(
+                    surfnet_svm.simnet_events_tx.info(format!(
                         "Metrics available at http://{}/metrics",
                         cmd.observability.metrics_addr
-                    )));
+                    ));
                 }
             }
         }
@@ -129,11 +129,11 @@ pub async fn handle_start_local_surfnet_command(
                 Option<surfpool_types::AccountSnapshot>,
             > = serde_json::from_str(&content)
                 .map_err(|e| format!("Failed to parse snapshot JSON '{}': {}", snapshot_path, e))?;
-            let _ = simnet_events_tx.send(SimnetEvent::info(format!(
+            simnet_events_tx.info(format!(
                 "Loaded {} accounts from snapshot file: {}",
                 snapshot_data.len(),
                 snapshot_path
-            )));
+            ));
 
             // Merge into the combined snapshot (later files override earlier ones)
             merged_snapshot.extend(snapshot_data);
@@ -197,11 +197,11 @@ pub async fn handle_start_local_surfnet_command(
         Ok(explorer_handle) => Some(explorer_handle),
         Err(e) => {
             error!("Failed to start subgraph and explorer server: {}", e);
-            let _ = simnet_events_tx.send(SimnetEvent::warn(format!(
+            simnet_events_tx.warn(format!(
                 "Failed to start subgraph and explorer server: {}",
                 e
-            )));
-            let _ = simnet_events_tx.send(SimnetEvent::info("Continuing with simnet startup..."));
+            ));
+            simnet_events_tx.info("Continuing with simnet startup...");
             None
         }
     };
@@ -221,7 +221,7 @@ pub async fn handle_start_local_surfnet_command(
             );
             if let Err(e) = hiro_system_kit::nestable_block_on(future) {
                 // Send the error through the event channel so the main thread can handle it
-                let _ = simnet_events_tx_for_thread.send(SimnetEvent::Aborted(e.to_string()));
+                simnet_events_tx_for_thread.emit(SimnetEvent::Aborted(e.to_string()));
             }
             Ok::<(), String>(())
         })
@@ -244,11 +244,11 @@ pub async fn handle_start_local_surfnet_command(
 
     // Re-send early events (like snapshot loading messages) so the TUI receives them
     for event in early_events {
-        let _ = simnet_events_tx.send(event);
+        simnet_events_tx.emit(event);
     }
 
     for event in airdrop_events {
-        let _ = simnet_events_tx.send(event);
+        simnet_events_tx.emit(event);
     }
 
     let simnet_commands_tx_copy = simnet_commands_tx.clone();
@@ -262,8 +262,7 @@ pub async fn handle_start_local_surfnet_command(
             // the watchdog's decision.
             Err(StartupPlanFailure::Planning(e)) => {
                 let _ = simnet_commands_tx_copy.send(SimnetCommand::FailStartupPlanning(e.clone()));
-                let _ = simnet_events_tx
-                    .send(SimnetEvent::warn(format!("Startup planning failed: {e}")));
+                simnet_events_tx.warn(format!("Startup planning failed: {e}"));
             }
             // The command loop is dead or wedged, so the startup state machine
             // is unreachable and no session can ever become ready.
@@ -309,8 +308,7 @@ pub async fn handle_start_local_surfnet_command(
         if let Ok(response) = response {
             if let Ok(body) = response.json::<CheckVersionResponse>().await {
                 if let Some(deprecation_notice) = body.deprecation_notice {
-                    let _ =
-                        simnet_events_tx.send(SimnetEvent::warn(deprecation_notice.to_string()));
+                    let _ = simnet_events_tx.warn(deprecation_notice.to_string());
                 }
             }
         }
--- a/crates/cli/src/cli/simnet/mod.rs
+++ b/crates/cli/src/cli/simnet/mod.rs
@@ -629,9 +629,19 @@ pub(crate) fn replay_early_events(
     early_events: Vec<SimnetEvent>,
     airdrop_events: Vec<SimnetEvent>,
 ) {
-    for event in early_events.into_iter().chain(airdrop_events) {
-        simnet_events_tx.emit(event);
-    }
+    // On a separate thread: the caller owns the only receiver and does not
+    // drain again until the frontend consumer starts, so a blocking emit on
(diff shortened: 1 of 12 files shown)


crates/cli/src/cli/simnet/mod.rs
```diff
@@ -629,3 +621,68 @@ mod tests {
...
+    // On a separate thread: the caller owns the only receiver and does not
+    // drain again until the frontend consumer starts, so a blocking emit on
+    // the caller's thread wedges startup once the buffer fills (a warm
+    // start can buffer a full channel of replayed transactions before
+    // CoreStarted). The replay thread blocks until the consumer drains,
+    // intentionally.
+    let tx = simnet_events_tx.clone();
+    let _ = hiro_system_kit::thread_named("early-event-replay").spawn(move || {
+        for event in early_events.into_iter().chain(airdrop_events) {
+            tx.emit(event);
```
MicaiahReid: Isn't this converting warn/info events onto an `emit`? Shouldn't we match on event and route accordingly?

crates/types/src/types.rs
```diff
@@ -569,6 +569,67 @@ impl SimnetEvent {
...
+    }
+
+    /// Lossless delivery for lifecycle events: blocks on a full buffer. A
+    /// disconnected receiver means shutdown, and the event no longer has an
+    /// audience.
+    ///
+    /// Never call this while holding the SVM lock: the reader may need that
+    /// lock to drain, so a blocked emit under the guard wedges every thread
+    /// behind it. Guard-context code sends with `log` and the receiver
+    /// recovers state it missed by reading the SVM, not the event stream.
```
```suggestion
    /// Lossless delivery for lifecycle events: blocks on a full buffer. A
    /// disconnected receiver means shutdown, and the event no longer has an
    /// audience.
    ///
    /// Never call this while holding the SVM lock: the reader may need that
    /// lock to drain, so a blocked emit under the guard wedges every thread
    /// behind it.
```
last sentence feels like noise

crates/types/src/types.rs
```diff
@@ -569,6 +569,67 @@ impl SimnetEvent {
...
+    pub fn emit(&self, event: SimnetEvent) {
```
MicaiahReid: Considering the needed considerations on having to beware of holding the SVM lock, and this sender driving usage policy, rather than having public general-purpose emit, should we have a function for each of the pre-ordained events that should actually emit? i.e. `pub fn abort`, `pub fn runbook_completed`, etc, etc. Perhaps overkill, just a thought

crates/core/src/tests/simnet_events.rs
```diff
@@ -0,0 +1,102 @@
+//! Pins the delivery contract of the simnet events channel.
+//!
+//! The channel is bounded(1024) and the sender is a [`SimnetEventsTx`],
+//! which owns the two delivery policies: `log` is lossy by contract
+//! (telemetry may drop under backpressure), `emit` is lossless (lifecycle
+//! events block until the reader drains). The first commit on this branch
+//! pinned the old behavior, where every core send was a silent `try_send`
+//! and a full buffer dropped an `Aborted` exactly like a log line; these
+//! tests assert the split that replaced it.
```
MicaiahReid: Remove references to "this branch", probably don't need to mention old behavior at all

crates/core/src/surfnet/svm.rs
```diff
@@ -3911,15 +3887,13 @@ impl SurfnetSvm {
...
-            .try_send(SimnetEvent::StartupStatusChanged(
-                self.startup_status.clone(),
-            ));
+        // log, not emit: callers hold the SVM write guard, and the bounded
+        // events channel may belong to an embedder that never drains it. The
+        // watch channel above is the reliable path; the event stream is a
+        // best effort sequence, and dropping an entry must not wedge the SVM.
+        self.simnet_events_tx.log(SimnetEvent::StartupStatusChanged(
+            self.startup_status.clone(),
+        ));
```
MicaiahReid: again, this being a `log` seems like a decision the caller of the SimnetEventsTx perhaps should not have to make? i.e., there's a `startup_status_changed` fn that logs appropriately for you, which has this comment inside explaining why log + not emit?

MicaiahReid (changes requested):

crates/cli/src/cli/simnet/mod.rs
cds-amal: good catch.

crates/types/src/types.rs
cds-amal: Pushing further is the right call. It got rid of more boilerplate. This is more aligned with Stefan Baumgartner's presentation. Thanks for getting me there!

crates/core/src/surfnet/svm.rs
cds-amal: I like this. It's a good way to evaluate the policy decision.

crates/types/src/types.rs
```diff
@@ -627,6 +629,105 @@ impl SimnetEventsTx {
     pub fn emit(&self, event: SimnetEvent) {
```
MicaiahReid: Should we remove pub on this now? I searched and there's just one more usage of `emit`, and it's a dangling `SimnetEvent::Aborted` that didn't get converted

cds-amal: I can do that!

MicaiahReid (approved):
LFG, thanks @cds-amal!

PR #761, opened 2026-08-13
feat(cli): ask before installing the Solana dev skill on first scaffold

michael-moffett:
Implements #567.

On the first `surfpool start` that scaffolds a `txtx.yml`:

    npx -y skills@1.5.23 add https://github.com/solana-foundation/solana-dev-skill --skill "*" --agent universal -y

Behind a prompt, not automatic. An existing `.agents/skills/solana-dev` in the project or `$HOME` skips it silently. Otherwise the answer is written to `~/.config/surfpool/dev-skill-{accepted,declined}`; only a yes installs, a decline is never asked twice, and `--yes` alone skips and says so. On a yes the child is spawned detached; its outcome reaches the user as a simnet event once it exits, so the scaffold never blocks on it.

The version is pinned so a first scaffold runs a known `skills` rather than whatever is latest that day. `--agent universal` is the installer's own key for the canonical `.agents/skills` copy; without it it detects none of its 77 registered agents and fans the skill out to all of them, dropping `.claude/`, a non-hidden `agent/` and `skills-lock.json` into a project that asked for none.

Nine tests cover it: the already-installed skip, `--yes` deciding nothing on its own and using a recorded answer, both prompt answers surviving a re-entry, an undisplayable prompt recording nothing, the two install outcomes being reported, and both reaching the user.

+370 / -8, two files.

### DIFF
--- a/crates/cli/src/scaffold/mod.rs
+++ b/crates/cli/src/scaffold/mod.rs
@@ -237,7 +237,7 @@ fn prompt_for_dev_skill(theme: &ColorfulTheme) -> Option<bool> {
         .with_prompt(format!(
             "Install the solana-dev skill? This will write to the {DEV_SKILL_DIR} directory."
         ))
-        .default(false)
+        .default(true)
         .interact()
         .ok()
 }
--- a/crates/cli/src/scaffold/mod.rs
+++ b/crates/cli/src/scaffold/mod.rs
@@ -153,7 +153,6 @@ const DEV_SKILL_DECLINED_MARKER: &str = ".config/surfpool/dev-skill-declined";
 
 const DEV_SKILL_ACCEPTED_MARKER: &str = ".config/surfpool/dev-skill-accepted";
 
-/// Returns the command that installs the solana-dev skill into `base_location`.
 fn dev_skill_install_command(base_location: &FileLocation) -> Command {
     let mut command = Command::new("npx");
     command.args([
@@ -222,7 +221,6 @@ fn record_dev_skill_answer(home: &Path, consented: bool) {
     let _ = File::create(marker);
 }
 
-/// Returns the command to install the solana-dev skill, or `None` if it should not be installed.
 fn dev_skill_install_if_wanted(
     base_location: &FileLocation,
     base: &Path,
@@ -686,7 +684,6 @@ mod tests {
         }
     }
 
-    /// A recorded answer is used even under `--yes`, and is never re-prompted.
     #[test]
     fn the_yes_flag_uses_the_recorded_answer() {
         for declined in [true, false] {
--- a/crates/cli/src/scaffold/mod.rs
+++ b/crates/cli/src/scaffold/mod.rs
@@ -227,8 +227,11 @@ fn record_dev_skill_answer(home: &Path, consented: bool) {
     let _ = File::create(marker);
 }
 
-/// A recorded answer and an install on disk both outrank `--yes`. Paths are arguments so the
-/// gate is testable without a real home directory.
+/// Only a recorded answer, or an install already on disk, decides this; `--yes` decides nothing
+/// and alone never installs. It pre-answers surfpool's own prompts, and nobody has been asked
+/// about a third-party repo whose contents we do not pin. What it still does is suppress the
+/// prompt, so a non-interactive run skips rather than blocks. Paths are arguments so the gate is
+/// testable without a real home directory.
 fn dev_skill_install_if_wanted(
     base_location: &FileLocation,
     base: &Path,
@@ -241,7 +244,17 @@ fn dev_skill_install_if_wanted(
     }
     let consented = match dev_skill_answer(home) {
         Some(recorded) => recorded,
-        None if auto_accept => true,
+        // An absent answer is not a yes, and a silent skip is the same invisible install by
+        // another name, so say it and name the run that does ask. Nothing is recorded: the
+        // question is still open.
+        None if auto_accept => {
+            println!(
+                "{} {} (run `surfpool start` without --yes to be asked)",
+                green!("Skipped the installer for"),
+                DEV_SKILL_DIR
+            );
+            false
+        }
         None => match ask() {
             Some(answer) => {
                 record_dev_skill_answer(home, answer);
@@ -662,9 +675,9 @@ mod tests {
     use tempfile::TempDir;
 
     use super::{
-        DEV_SKILL_AGENT, DEV_SKILL_DECLINED_MARKER, DEV_SKILL_DIR, DEV_SKILL_INSTALLER,
-        DEV_SKILL_REPO, FileLocation, dev_skill_answer, dev_skill_install_command,
-        dev_skill_install_if_wanted, spawn_dev_skill_install,
+        DEV_SKILL_ACCEPTED_MARKER, DEV_SKILL_AGENT, DEV_SKILL_DECLINED_MARKER, DEV_SKILL_DIR,
+        DEV_SKILL_INSTALLER, DEV_SKILL_REPO, FileLocation, dev_skill_answer,
+        dev_skill_install_command, dev_skill_install_if_wanted, spawn_dev_skill_install,
     };
 
     /// Every gate test runs against these, never a real home directory.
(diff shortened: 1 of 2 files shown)


MicaiahReid (changes requested):
I wouldn't want a CLI I use to automatically install a skill without asking me. At a minimum this should:
 1. Check if the skill is already installed and not prompt anything if so
 2. If not installed, ask the user if they want it instealled
 3. Install if yes, never ask again if no

If you want to make these changes, lmk. If this is beyond the scope you originally signed up for, lmk and I'll add these notes to the issue

crates/cli/src/scaffold/mod.rs
```diff
@@ -134,6 +139,134 @@ impl ProgramMetadata {
...
+        let _ = child.wait();
+    });
+}
+
+fn dev_skill_installed(base: &Path, home: &Path) -> bool {
+    base.join(DEV_SKILL_DIR).exists() || home.join(DEV_SKILL_DIR).exists()
+}
+
+/// Both answers are ours to keep: `DEV_SKILL_DIR` is written by an installer we neither wait on
+/// nor control, so a yes inferred from it is a yes asked again forever. A no wins a split pair.
```
MicaiahReid: A doc comment should explain what the function does, can this be rewritten?

crates/cli/src/scaffold/mod.rs
```diff
@@ -134,6 +139,134 @@ impl ProgramMetadata {
...
+        let _ = fs::create_dir_all(parent);
+    }
+    let _ = File::create(marker);
+}
+
+/// Only a recorded answer, or an install already on disk, decides this; `--yes` decides nothing
+/// and alone never installs. It pre-answers surfpool's own prompts, and nobody has been asked
+/// about a third-party repo whose contents we do not pin. What it still does is suppress the
+/// prompt, so a non-interactive run skips rather than blocks. Paths are arguments so the gate is
+/// testable without a real home directory.
```
MicaiahReid: Again, doc comments should explain what the function is doing. Please rewrite

crates/cli/src/scaffold/mod.rs
```diff
@@ -134,6 +139,134 @@ impl ProgramMetadata {
...
+        Some(recorded) => recorded,
+        // An absent answer is not a yes, and a silent skip is the same invisible install by
+        // another name, so say it and name the run that does ask. Nothing is recorded: the
+        // question is still open.
+        None if auto_accept => {
+            println!(
+                "{} {} (run `surfpool start` without --yes to be asked)",
+                green!("Skipped the installer for"),
+                DEV_SKILL_DIR
+            );
```
MicaiahReid: Kind of a strange log. I think a silent skip/return false is better here

crates/cli/src/scaffold/mod.rs
```diff
@@ -134,6 +139,134 @@ impl ProgramMetadata {
...
+        Some(true)
+    } else {
+        None
+    }
+}
+
+fn prompt_for_dev_skill(theme: &ColorfulTheme) -> Option<bool> {
+    Confirm::with_theme(theme)
+        .with_prompt(format!(
+            "Install the solana-dev skill? It writes {DEV_SKILL_DIR} and skills-lock.json here"
```
MicaiahReid:
```suggestion
            "Install the solana-dev skill? This will write to the {DEV_SKILL_DIR} directory."
```

crates/cli/src/scaffold/mod.rs
```diff
@@ -512,5 +645,227 @@ pub fn scaffold_iac_layout(
...
+        command
+    }
+
+    /// #567 supplied this invocation literally; we depart from its text twice. The pin is
+    /// spelled out here so that dropping it has to fail this. `--agent universal` is the
+    /// second: without it the installer detects none of its 77 registered agents, falls
+    /// through to fanout, and drops `.claude/`, a non-hidden `agent/` and `skills-lock.json`
+    /// into a project that asked for none of them.
+    #[test]
+    fn the_install_is_the_command_issue_567_asked_for() {
```
MicaiahReid: This is a very strange test, so specific to the issue. I don't think the issue was meant to be an exact prescription of what should be installed, and this doesn't need to be codified in a test.

crates/cli/src/scaffold/mod.rs
```diff
@@ -512,5 +645,227 @@ pub fn scaffold_iac_layout(
...
+        assert!(
+            exact,
+            "the installer must be pinned to an exact version, not a range: {version}"
+        );
+    }
+
+    /// `surfpool start -m ../elsewhere/txtx.yml` scaffolds a tree the caller is not standing in,
+    /// so the manifest's directory is the target and the shell's is not.
+    #[test]
+    fn the_install_runs_in_the_scaffolded_project_not_the_process_cwd() {
```
MicaiahReid: The install runs in the passed `base`, not necessarily the scaffolded project... this naming is misleading. And in general seems like a very low-value test

crates/cli/src/scaffold/mod.rs
```diff
@@ -512,5 +645,227 @@ pub fn scaffold_iac_layout(
...
+        let base = FileLocation::from_path_string("/tmp/surfpool-567-scaffold").unwrap();
+        assert_eq!(
+            dev_skill_install_command(&base).get_current_dir(),
+            Some(Path::new("/tmp/surfpool-567-scaffold"))
+        );
+    }
+
+    /// A skill already on disk is the reviewer's first item: say nothing, do nothing.
+    /// Either scope counts, since a globally installed skill is already available here.
```
MicaiahReid: irrelevant comment

crates/cli/src/scaffold/mod.rs
```diff
@@ -512,5 +645,227 @@ pub fn scaffold_iac_layout(
...
+                "installed in base={in_base} home={in_home} still produced an install"
+            );
+        }
+    }
+
+    /// Under `--yes`, the marker decides and only the marker does. The remembered answer is read
+    /// before the flag rather than after, so a CI machine neither relitigates a no hourly nor
+    /// re-asks a user who already said yes. Neither case may reach the prompt.
+    #[test]
+    fn under_the_yes_flag_the_marker_decides() {
```
MicaiahReid: Please distill comment and name to give only the needed information and to make clear sense on first read

MicaiahReid (changes requested):
Again I ask you to please self-review your code before submitting for review. This is especially critical if you are using AI to do a majority of the work. Self-review for: tests that are low-value but a maintenance burden, doc comments that don't actually explain what the function does but explains why the agent wrote it or some detailed inner workings of the function, doc comments that use technical jargon that require the reader 10-rereads to really understand.

crates/cli/src/scaffold/mod.rs
```diff
@@ -134,6 +139,113 @@ impl ProgramMetadata {
...
+        None
+    }
+}
+
+fn prompt_for_dev_skill(theme: &ColorfulTheme) -> Option<bool> {
+    Confirm::with_theme(theme)
+        .with_prompt(format!(
+            "Install the solana-dev skill? This will write to the {DEV_SKILL_DIR} directory."
+        ))
+        .default(false)
```
MicaiahReid:
```suggestion
        .default(true)
```
Let's encourage install by making default true here. We're giving them a chance to opt out and won't bug them about it later.

MicaiahReid (changes requested):

crates/cli/src/scaffold/mod.rs
```diff
@@ -134,6 +139,144 @@ impl ProgramMetadata {
...
+    command.args([
+        "-y",
+        DEV_SKILL_INSTALLER,
+        "add",
+        DEV_SKILL_REPO,
+        "--skill",
+        "*",
+        "--agent",
+        DEV_SKILL_AGENT,
+        "-y",
```
MicaiahReid, line 166:
```suggestion
        "-y",
        "--global",
```

MicaiahReid (changes requested):
This is looking great! I tested and everything seems to be working, accept I think we want a global install for this skill. Incorporate that suggested commit and I'm ready to merge!

MicaiahReid (approved):
Great, thanks @michael-moffett!

PR #762, opened 2026-08-13
fix(core): honor commitment for remote account batches

xternet:
## Summary

`SurfnetRemoteClient::get_multiple_accounts` accepts a caller-selected commitment, but its primary account batch used `RpcClient::get_multiple_accounts`, which silently applied the client's default commitment. Companion mint/program-account fetches in the same method already used the requested commitment.

Use `get_multiple_accounts_with_commitment` for the primary batch as well, then extract the contextual response value. This keeps every remote account fetched by the operation at the same requested commitment.

## Regression test

The new test records the outgoing JSON-RPC request through a custom `RpcSender`. It requests `confirmed` and asserts that the `getMultipleAccounts` config actually contains `"commitment": "confirmed"`; before this patch it observed `"finalized"`.

## Validation

- `cargo test -p surfpool-core surfnet::remote::tests -- --test-threads=1` — 3 passed
- `cargo clippy -p surfpool-core --lib --no-default-features` — completed with existing repository warnings only
- `git diff --check`

I also attempted the complete `surfpool-core` lib suite locally. It reached 469 passing tests before the shared runner exhausted its open-file limit; the remaining failures reported `Too many open files`, rather than assertion failures from this change. GitHub CI can run the full suite in its normal environment.

### DIFF
--- a/crates/core/src/surfnet/remote.rs
+++ b/crates/core/src/surfnet/remote.rs
@@ -311,9 +311,10 @@ impl SurfnetRemoteClient {
 
         let remote_accounts = self
             .client
-            .get_multiple_accounts(pubkeys)
+            .get_multiple_accounts_with_commitment(pubkeys, commitment_config)
             .await
-            .map_err(SurfpoolError::get_multiple_accounts)?;
+            .map_err(SurfpoolError::get_multiple_accounts)?
+            .value;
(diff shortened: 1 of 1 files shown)


MicaiahReid (changes requested):
@xternet I'll need commit signature verification here as well.

MicaiahReid (approved):

PR #764, opened 2026-08-13
fix(core): preserve remote transaction errors

xternet:
## Summary

Surfpool's remote `getTransaction` path previously treated every client error as a missing transaction. This conflated two different RPC outcomes:

- a valid JSON-RPC `null`, meaning the transaction is unavailable; and
- transport, HTTP, provider, or decoding failures, meaning the lookup did not complete reliably.

The Solana RPC client convenience method deserializes directly into a concrete transaction, so a valid `null` arrives through its error path too.

## Change

- Deserialize the raw `getTransaction` response as `Option<EncodedConfirmedTransactionWithStatusMeta>`.
- Keep valid `null` mapped to `GetTransactionResult::None`.
- Propagate genuine provider failures through the locker's existing `SurfpoolResult` boundary.
- Preserve the published `SurfnetRemoteClient::get_transaction` return type with a compatibility wrapper; the new fallible method is crate-private.
- Sanitize transport failures before putting them in client-visible JSON-RPC error data. Reqwest errors can include the full datasource URL, so callers receive the useful failure category plus scheme/host, never userinfo, path, or query credentials.

## Regression coverage

- A recorded `null` response remains `None` and verifies the exact outgoing `GetTransaction` request/config.
- A provider failure reaches the locker caller instead of becoming `None`.
- A real local HTTP 401 response is requested through a URL carrying credentials in userinfo, path, and query; the propagated error retains `401 Unauthorized` and the sanitized host but omits every secret.

## Validation

- `cargo test -p surfpool-core surfnet::remote::tests::a_ -- --test-threads=1` — 5 passed
- `cargo test -p surfpool-core --no-default-features surfnet::remote::tests::a_ -- --test-threads=1` — 5 passed
- `cargo clippy -p surfpool-core --lib --no-default-features` — completed with existing repository warnings only
- `git diff --check`

This branch and #762 both add tests to `surfnet/remote.rs`; whichever lands second may need a trivial rebase, but the production changes are independent.

### DIFF
--- a/crates/core/src/surfnet/remote.rs
+++ b/crates/core/src/surfnet/remote.rs
@@ -12,7 +12,7 @@ use solana_client::{
         RpcSignaturesForAddressConfig, RpcTokenAccountsFilter, RpcTransactionConfig,
     },
     rpc_filter::RpcFilterType,
-    rpc_request::{RpcRequest, TokenAccountsFilter},
+    rpc_request::{RpcError, RpcRequest, TokenAccountsFilter},
     rpc_response::{
         RpcAccountBalance, RpcConfirmedTransactionStatusWithSignature, RpcKeyedAccount, RpcResult,
         RpcTokenAccountBalance,
@@ -81,10 +81,20 @@ fn sanitized_client_error(error: &ClientError, datasource_url: &str) -> String {
             error.line(),
             error.column()
         ),
-        ClientErrorKind::RpcError(error) => error.to_string().replace(datasource_url, &endpoint),
+        ClientErrorKind::RpcError(error) => match error {
+            RpcError::RpcRequestError(_) | RpcError::ForUser(_) => {
+                format!("datasource RPC request failed for {endpoint}")
+            }
+            RpcError::RpcResponseError { code, .. } => {
+                format!("datasource RPC response error {code} from {endpoint}")
+            }
+            RpcError::ParseError(_) => {
+                format!("failed to parse the RPC response from {endpoint}")
+            }
+        },
         ClientErrorKind::SigningError(error) => error.to_string(),
         ClientErrorKind::TransactionError(error) => error.to_string(),
-        ClientErrorKind::Custom(error) => error.replace(datasource_url, &endpoint),
+        ClientErrorKind::Custom(_) => format!("datasource client error for {endpoint}"),
     }
 }
 
@@ -647,6 +657,8 @@ where
 mod tests {
     use std::sync::{Arc, Mutex};
 
+    use solana_client::rpc_request::RpcResponseErrorData;
+
     use super::*;
     use crate::surfnet::{locker::SurfnetSvmLocker, svm::SurfnetSvm};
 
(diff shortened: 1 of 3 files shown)


crates/core/src/surfnet/remote.rs
```diff
@@ -52,6 +54,40 @@ use crate::{
...
+        ClientErrorKind::Middleware(_) => format!("datasource middleware failed for {endpoint}"),
+        ClientErrorKind::Io(error) => {
+            format!("datasource I/O error ({:?}) for {endpoint}", error.kind())
+        }
+        ClientErrorKind::SerdeJson(error) => format!(
+            "invalid JSON response from {endpoint} at line {}, column {}",
+            error.line(),
+            error.column()
+        ),
+        ClientErrorKind::RpcError(error) => error.to_string().replace(datasource_url, &endpoint),
```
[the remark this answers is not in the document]
xternet: Fixed in b1262b9. Client-visible RpcError and Custom output no longer forwards provider-controlled text; it preserves only the safe category/code and sanitized endpoint. Added regression coverage for reflected userinfo, path, and query fragments. Thanks for catching this.

koriyoshi2041 (approved):
Validated exact head `b1262b94f0` locally. `cargo test -p surfpool-core surfnet::remote::tests::a_ -- --test-threads=1` passed 5/5: valid JSON-RPC `null` remains a missing transaction, provider failures reach the locker caller, and the 401/timeout paths retain useful failure categories without disclosing datasource userinfo, path, or query credentials. The post-review change also removes provider-controlled RpcError/Custom text rather than trying to redact only a full URL. `git diff --check` is clean; the only build output was the existing unrelated dead-code warning in the integration test module.

crates/core/src/surfnet/remote.rs
```diff
@@ -394,15 +440,53 @@ impl SurfnetRemoteClient {
...
+            Ok(result) => result,
+            Err(_) => {
+                // Preserve the published method's return type. Error details may
+                // contain datasource credentials, so this compatibility path
+                // logs only the sanitized endpoint; fallible callers should use
+                // `try_get_transaction`.
+                error!(
+                    "Failed to fetch transaction {signature} from {}",
+                    sanitized_datasource_url(&self.client.url())
+                        .unwrap_or_else(|| "the datasource".to_string())
```
MicaiahReid: Should we just log `Err(e)` here? Doesn't the fn we called (`try_get_transaction`) already give us a nice error with sanitized datasource?

xternet: Yes — `try_get_transaction` already maps the raw `ClientError` through `sanitized_client_error` before wrapping it in `SurfpoolError`. Simplified the compatibility path in 8785c15 to bind `Err(e)` and log `e` directly. Both focused five-test suites still pass.

MicaiahReid (changes requested):
@xternet I can't merge without verified commit signatures: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

MicaiahReid (approved):

PR #767, opened 2026-08-19
feat(scenarios): add Pump support - state templates & presets

92Infinitus92:
Adds first-class Pump and PumpSwap state preparation through Surfpool's scenario pipeline.

The integration includes versioned IDLs and five templates covering Pump bonding curves, Pump global configuration, arbitrary PumpSwap pools, canonical migrated WSOL pools, and PumpSwap global configuration. It derives the relevant PDAs automatically, provides a verified Pump token catalog, supports direct pool addresses when derivation is not possible, and preserves untouched account bytes during mutation.

It also adds validated builders and HTTP and MCP entry points for preparing Token-2022 graduation and canonical PumpSwap price-shock scenarios. The graduation path calculates the completing buy and migration reserve through production code, while the price-shock path updates the economic state used by a real PumpSwap sell.

The PR includes focused coverage for template derivation, byte-preserving state mutation, graduation calculations, and price-shock preparation. The focused core suite passes with 12 tests. The checked-in live-fork lifecycle test is ignored by default because it requires external programs and network access, but it was also executed separately and passed. The two CLI validation tests and formatting checks pass.


crates/core/src/rpc/surfnet_cheatcodes.rs
```diff
@@ -1712,8 +1719,30 @@ impl SurfnetCheatcodes for SurfnetCheatcodesRpc {
...
+                    token_account_data
+                        .patch_amount_preserving_extensions(token_account.expected_data())
+                        .map_err(|e| {
+                            Error::invalid_params(format!(
+                                "Failed to patch token account amount: {}",
+                                e
+                            ))
+                        })?,
+                    initial_lamports,
+                )
```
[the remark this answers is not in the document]
92Infinitus92: @greptileai check commit f519a4e for solution of the reported bug

crates/cli/src/http/mod.rs
```diff
@@ -54,14 +70,163 @@ fn configure_api(cfg: &mut web::ServiceConfig) {
...
+        accounts[3].account(),
+        global_account,
+    )
+    .map_err(|error| actix_web::error::ErrorBadRequest(error.to_string()))
+}
+
+#[post("/v1/scenarios/pump-graduation")]
+async fn post_pump_graduation_scenario(
+    request: web::Json<PumpGraduationScenarioRequest>,
+    source: Data<PumpScenarioDataSource>,
```
MicaiahReid: I'm not understanding why we require a remote data source, and why we use it for all of the pump account fetches.

What if the user has created some of these pump accounts locally already in the survnet vm? Or if they've already been fetched from mainnet, cached locally, and modified? With this, we're bypassing all existing local account data

92Infinitus92: Hey @MicaiahReid fixed the branch with the mismatching commits

crates/core/src/scenarios/protocols/spl-token/overrides.yaml
```diff
@@ -8,6 +8,13 @@ tags:
   - vault
   - balance
 
+constants:
```
MicaiahReid: I'm surprised to see protocol-specific overrieds on spl-token. Is this something that should be made more generic? Both the new `pump-token-2022-curve-balance` and `spl-token-account-balance` are overriding the `TokenAccount.amount` field

crates/mcp/src/surfpool/mod.rs
```diff
@@ -345,6 +374,95 @@ impl TokenAddressResponse {
...
 }
 
+impl Surfpool {
+    /// Reads through the surfnet's own RPC: local state wins, only missing
+    /// accounts fall back to its remote source.
+    async fn fetch_surfnet_accounts(
+        &self,
+        pubkeys: &[Pubkey],
+    ) -> Result<Vec<Option<Account>>, String> {
+        let client = SurfnetRemoteClient::new(format!("http://127.0.0.1:{DEFAULT_RPC_PORT}"));
```
MicaiahReid: Only using `DEFAULT_RPC_PORT` here seems a bit frail - what if user is running surfpool on a different port?

MicaiahReid (changes requested):

crates/cli/src/http/mod.rs
92Infinitus92: Right — the original version bypassed local state. Fixed in the current head: reads go through the surfnet's own RPC, so local state wins and only accounts missing from the fork fall back to the remote source.

crates/core/src/scenarios/protocols/spl-token/overrides.yaml
92Infinitus92: Went one further than keeping it — dropped the pump vault template entirely. The graduation builder now derives the vault itself (curve PDA → its Token-2022 ATA via the SDK helpers) and emits the override directly on `spl-token-account-balance` with the resolved address. So the write goes through the generic spl template, exactly as you suggested, and spl-token carries nothing pump-specific. The vault/curve coupling rule moved into the curve template's llm_context so composing models still learn it.

crates/mcp/src/surfpool/mod.rs
92Infinitus92: Agreed — the tool now takes an optional `surfnetPort` (default 8899), same as `call_surfnet_rpc` already does. Matters doubly since `start_surfnet` itself spawns surfnets on 18899, 28899 and so on. Thanks.

92Infinitus92: resolved, thanks @MicaiahReid

crates/core/src/scenarios/protocols/spl-token/overrides.yaml
92Infinitus92: resolved, thanks @MicaiahReid

MicaiahReid (approved):
Nice work @92Infinitus92!

PR #768, opened 2026-08-19
feat(mcp): split template discovery into index and detail tools

92Infinitus92:
The current template discovery tool returns every template with its complete property descriptions and constant metadata. As protocol coverage grows, that payload spends model context on templates the model will not use.

Template discovery now returns a compact index, while a separate detail tool retrieves one complete template on demand. Constant catalogs remain summarized and searchable through the existing option search tool, and unknown template ids return the available alternatives.

Formatting passes and the complete MCP suite passes with 17 tests.

Rebased onto main after #767 merged: the diff is now the single discovery commit, with nothing stacked underneath.

### DIFF
--- a/crates/mcp/src/surfpool/mod.rs
+++ b/crates/mcp/src/surfpool/mod.rs
@@ -100,6 +100,18 @@ fn compact_template_json(template: &surfpool_types::OverrideTemplate) -> serde_j
     obj
 }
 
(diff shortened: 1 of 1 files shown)


MicaiahReid (approved):
Very nice, thanks @92Infinitus92!

PR #769, opened 2026-08-19
fix(cli): merge partial scenario updates

92Infinitus92:
PATCH /v1/scenarios/{id} currently replaces the complete stored scenario. A client that sends only a changed name or description can therefore erase overrides, tags, and backend-only data.

PATCH now merges supplied fields into an existing scenario, rejects unknown fields, and keeps the path id authoritative. Complete scenario  documents retain the existing upsert behavior for compatibility.

### DIFF
--- a/crates/cli/src/http/mod.rs
+++ b/crates/cli/src/http/mod.rs
@@ -286,11 +286,22 @@ fn scenario_from_full_patch(patch: &serde_json::Value, path_id: &str) -> Result<
     let obj = scenario
         .as_object_mut()
         .ok_or_else(|| "PATCH body must be a JSON object".to_string())?;
+    let supplied_fields = obj.keys().cloned().collect::<Vec<_>>();
     obj.insert(
         "id".to_string(),
         serde_json::Value::String(path_id.to_string()),
     );
-    serde_json::from_value(scenario).map_err(|e| e.to_string())
+    let scenario: Scenario = serde_json::from_value(scenario).map_err(|e| e.to_string())?;
+    let serialized = serde_json::to_value(&scenario).map_err(|e| e.to_string())?;
+    let known_fields = serialized
+        .as_object()
+        .ok_or_else(|| "Failed to serialize the scenario".to_string())?;
+    for key in supplied_fields {
+        if !known_fields.contains_key(&key) {
+            return Err(format!("Unknown scenario field '{key}'"));
+        }
+    }
+    Ok(scenario)
 }
 
 fn json_error(status: actix_web::http::StatusCode, message: String) -> HttpResponse {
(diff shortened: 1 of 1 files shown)


crates/cli/src/http/mod.rs
```diff
@@ -256,10 +256,64 @@ async fn delete_scenario(
...
+        .as_object()
+        .ok_or_else(|| "PATCH body must be a JSON object".to_string())?;
+    let mut merged = serde_json::to_value(existing).map_err(|e| e.to_string())?;
+    let obj = merged
+        .as_object_mut()
+        .ok_or_else(|| "Failed to serialize the stored scenario".to_string())?;
+    for (key, value) in patch {
+        if key != "id" && !obj.contains_key(key) {
+            return Err(format!("Unknown scenario field '{key}'"));
+        }
```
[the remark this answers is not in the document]
92Infinitus92: @greptileai check commit [66e589a](https://github.com/solana-foundation/surfpool/pull/769/commits/66e589a52f0d3f6454d83624dcf33c026bf84fd8) for a solution, review it

crates/cli/src/http/mod.rs
```diff
@@ -256,10 +256,86 @@ async fn delete_scenario(
...
+    }
+    obj.insert(
+        "id".to_string(),
+        serde_json::Value::String(path_id.to_string()),
+    );
+    deserialize_scenario_strictly(merged)
+}
+
+fn scenario_from_full_patch(patch: &serde_json::Value, path_id: &str) -> Result<Scenario, String> {
+    let mut scenario = patch.clone();
```
MicaiahReid: We are immediately cloning the reference here, are we able to just pass in `patch: mut serde_json::Value`?

crates/cli/src/http/mod.rs
```diff
@@ -432,4 +514,282 @@ mod tests {
...
+    #[actix_web::test]
+    async fn explicit_empty_array_still_clears() {
+        let existing = scenario_of(scenario_json("s1", "before"));
+        let patch = serde_json::json!({ "overrides": [] });
+
+        let merged = merge_scenario_patch(&existing, &patch, "s1").unwrap();
+
+        assert!(merged.overrides.is_empty(), "an explicit [] clears");
+        assert_eq!(merged.name, "before", "other fields untouched");
+    }
```
MicaiahReid: Can we remove "still" and "before" wording in this test? Rather just naming the test based on what's being asserted. i.e. `explicit_empty_array_patch_clears_scenario_overrides`?

MicaiahReid (changes requested):

crates/cli/src/http/mod.rs
92Infinitus92: fixed, thanks

crates/cli/src/http/mod.rs
92Infinitus92: fixed, thanks

MicaiahReid (approved):

PR #777, opened 2026-08-25
fix(core): render token uiAmount through the account decoder

pucedoteth:
## The bug

`format_ui_amount_string` in `crates/core/src/surfnet/locker.rs` divided through `f64` before formatting:

```rust
let divisor = 10u64.pow(decimals as u32);
format!("{:.decimals$}", amount as f64 / divisor as f64, decimals = decimals as usize)
```

An `f64` mantissa holds 53 bits; a token amount holds 64. Every amount above 2^53 was rounded before it reached the client:

| amount | decimals | rendered | correct |
|---|---|---|---|
| `18446744073709551615` | 6 | `18446744073709.550781` | `18446744073709.551615` |
| `12345678901234567` | 6 | `12345678901.234568` | `12345678901.234567` |
| `9007199254740993` | 2 | `90071992547409.92` | `90071992547409.93` |

This is not a theoretical range. 2^53 raw units is only ~9B tokens on a 6-decimal mint, so an agent reading a supply-scale balance gets a wrong number back — the same class of defect #758 fixed for `getTokenSupply`, in the helpers that path didn't touch.

Both helpers additionally built their divisor with `10u64.pow(decimals)`, which overflows from `decimals >= 20`. `Mint::decimals` is a plain `u8` with no cap in the token program, so a mint declaring 20+ decimals made this a **panic in debug** and a silently wrong divisor in release:

```
format_ui_amount_string(1, 20) -> panicked: attempt to multiply with overflow
format_ui_amount(1, 20)        -> panicked: attempt to multiply with overflow
```

## Reachability

All four call sites pass real mint decimals, so both defects are live on RPC output:

- `locker.rs:3040` — `getTokenLargestAccounts`
- `types.rs:516`, `:538`, `:732` — pre/post token balances on `getTransaction` and `getBlock`

## The fix

Defer the string to `real_number_string_trimmed` — the helper `getTokenSupply` already uses since a4acb1e (#758) — and give `format_ui_amount` the `Option<f64>` return that `UiTokenAmount::ui_amount` is already typed for. That mirrors `solana_account_decoder::parse_token::token_amount_to_ui_amount_v3` exactly:

```rust
let ui_amount = 10_usize
    .checked_pow(decimals as u32)
    .map(|dividend| amount as f64 / dividend as f64);
(ui_amount, real_number_string_trimmed(amount, decimals))
```

The three `types.rs` call sites drop their now-redundant `Some(...)` wrapper.

**One deliberate behaviour change:** trailing zeros are no longer padded, so 1 USDC renders as `"1"` rather than `"1.000000"`. That is what `real_number_string_trimmed` does, so it is what agave returns on mainnet and what `getTokenSupply` already returned here after #758 — this brings the remaining paths into line rather than away from it.

## Verification

Four tests added to the existing `locker.rs` test module. Reverting **only** the production helper body — tests untouched — turns three of them red with exactly the values above:

```
test_format_ui_amount_string_is_exact_above_2_pow_53 ... FAILED
  left: "18446744073709.550781"   right: "18446744073709.551615"
test_format_ui_amount_string_matches_account_decoder ... FAILED
  left: "1.000000"                right: "1"
test_format_ui_amount_survives_decimals_that_cannot_form_a_divisor ... FAILED
  panicked at core/src/num/mod.rs:1274 (pow overflow)
```

All four pass with the fix. Full `cargo test -p surfpool-core --lib`: **716 passed**. The `test_simulate_add_alt_entries_fetching` variants fail identically on a clean `main` at this commit (`loaded_accounts_data_size` 6090 vs 140134 — it fetches from mainnet), so they are pre-existing and unrelated. `cargo fmt` and `cargo clippy` clean on the changed lines.

### DIFF
--- a/crates/core/src/surfnet/locker.rs
+++ b/crates/core/src/surfnet/locker.rs
@@ -15,7 +15,7 @@ use solana_account_decoder::{
     UiAccount, UiAccountEncoding, UiDataSliceConfig,
     parse_account_data::AccountAdditionalDataV3,
     parse_bpf_loader::{BpfUpgradeableLoaderAccountType, UiProgram, parse_bpf_upgradeable_loader},
-    parse_token::UiTokenAmount,
+    parse_token::{UiTokenAmount, real_number_string_trimmed},
 };
 use solana_address_lookup_table_interface::state::AddressLookupTable;
 use solana_client::{
@@ -3037,8 +3037,8 @@ impl SurfnetSvmLocker {
                     amount: UiTokenAmount {
                         amount: token_account.amount().to_string(),
                         decimals: mint_decimals,
-                        ui_amount: Some(format_ui_amount(token_account.amount(), mint_decimals)),
-                        ui_amount_string: format_ui_amount_string(
+                        ui_amount: format_ui_amount(token_account.amount(), mint_decimals),
+                        ui_amount_string: real_number_string_trimmed(
                             token_account.amount(),
                             mint_decimals,
                         ),
@@ -4548,26 +4548,13 @@ fn update_programdata_account(
     }
 }
 
-pub fn format_ui_amount_string(amount: u64, decimals: u8) -> String {
-    if decimals > 0 {
-        let divisor = 10u64.pow(decimals as u32);
-        format!(
-            "{:.decimals$}",
-            amount as f64 / divisor as f64,
-            decimals = decimals as usize
-        )
-    } else {
-        amount.to_string()
-    }
-}
-
-pub fn format_ui_amount(amount: u64, decimals: u8) -> f64 {
-    if decimals > 0 {
-        let divisor = 10u64.pow(decimals as u32);
-        amount as f64 / divisor as f64
-    } else {
-        amount as f64
-    }
+/// Scales a raw token amount by its mint's decimals for `UiTokenAmount::ui_amount`.
+///
+/// `None` when `decimals` is too large for `10^decimals` to fit a `usize`.
+pub fn format_ui_amount(amount: u64, decimals: u8) -> Option<f64> {
+    10_usize
+        .checked_pow(decimals as u32)
+        .map(|divisor| amount as f64 / divisor as f64)
 }
 
 #[cfg(test)]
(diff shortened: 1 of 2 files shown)


crates/core/src/surfnet/locker.rs
```diff
@@ -4548,26 +4548,27 @@ fn update_programdata_account(
...
-    }
+/// The `uiAmount` float, or `None` when `decimals` is too large to form a divisor.
+///
+/// Matches `solana_account_decoder::parse_token::token_amount_to_ui_amount_v3`, which computes
+/// this as `10_usize.checked_pow(decimals)` mapped over the division: the field is
+/// `Option<f64>` precisely so an undivisible `decimals` has somewhere to go other than a panic.
+pub fn format_ui_amount(amount: u64, decimals: u8) -> Option<f64> {
+    10_usize
+        .checked_pow(decimals as u32)
+        .map(|divisor| amount as f64 / divisor as f64)
```
MicaiahReid: The doc comments on these are classic over-described AI generated comments. Please distill this down to what's important. What does the function do? If you had no context, what comment would help you know what the function does, when to use it, and what to pass in?

The comment doesn't need super technical details of what's happening under the hood (the code itself explains what the code is doing). We don't need references to other PRs and how past behavior worked

crates/core/src/surfnet/locker.rs
```diff
@@ -4548,26 +4548,27 @@ fn update_programdata_account(
...
 
+/// The decimal rendering of a raw token amount, as agent-visible RPC output.
+///
+/// Defers to `real_number_string_trimmed`, the same helper `getTokenSupply` uses since #758,
+/// rather than dividing through `f64`. An `f64` carries 53 bits of mantissa and a token amount
+/// carries 64, so every amount above 2^53 rounded: `u64::MAX` at 6 decimals rendered as
+/// `18446744073709.550781` instead of `18446744073709.551615`. It also drops the
+/// `10u64.pow(decimals)` that overflowed for `decimals >= 20` — `Mint::decimals` is a `u8` with
+/// no cap in the token program, so that was a panic in debug and a wrong divisor in release.
 pub fn format_ui_amount_string(amount: u64, decimals: u8) -> String {
```
MicaiahReid: If this function is just directly calling another - why does the function need to exist at all?

crates/core/src/surfnet/locker.rs
```diff
@@ -7027,4 +7028,63 @@ mod tests {
...
+                real_number_string_trimmed(1, decimals),
+                "decimals {decimals} must render rather than overflow"
+            );
+            assert_eq!(
+                format_ui_amount(1, decimals).is_some(),
+                10_usize.checked_pow(decimals as u32).is_some(),
+                "decimals {decimals} must yield a float exactly when a divisor exists"
+            );
+        }
+    }
```
MicaiahReid: These tests don't seem particularly valuable, they're essentially asserting the one function has the same output as another... when the first function is just calling the other

MicaiahReid (changes requested):
I appreciate the PR, but in future PRs please do more to distill the AI output.

MicaiahReid (approved):
thanks, @pucedoteth!

PR #778, opened 2026-08-25
feat(core): derive geyser slot statuses from a per-slot lifecycle

cds-amal:
## 1. What

This PR refactors #747 starting with its bones (the status set, warp handling, and wiring points) and moves the scattered emission sites behind one state machine, so the ordering guarantees are properties of the machine, not its call sites.

Every slot-status emission now comes from one transition relation, a per-slot lifecycle registry in `surfnet/slot_lifecycle/`:

* `announce` emits `CreatedBank`, `produce` emits `Processed`, and `confirm` emits `Confirmed`
* rooting is a threshold: `root_through(r)` roots every confirmed slot at or below `r`, in slot order, from the registry's own record
* a forward warp emits `Dead` for the abandoned open slot and `CreatedBank` for the destination; a backward warp is a reorg, emitting `Dead` for every rewritten slot, in slot order, before re-announcing the destination

Each transition requires its exact predecessor stage, so a status cannot be skipped or repeated within a bank. Block production follows lifecycle order (block data, then produce/confirm, then root, then announce the next slot), so consumers see every slot announced before its data and its data before confirmation. `UpdateSlotStatus` carries Agave's `SlotStatus`, since `CreatedBank` and `Dead` exist only there.

## 2. Consumer-visible changes

Both streams now derive from the same lifecycle, which makes a few wire-visible changes.

### WS and geyser agree on lifecycle statuses

`slotsUpdatesSubscribe` now mirrors `CreatedBank`, `OptimisticConfirmation`, `Root`, and `Dead` from the same emission point as geyser. Warp statuses, which previously reached geyser only, now reach both.

`Frozen` stays with block production, which alone knows the block's transaction stats. Within a block, `CreatedBank` for the next slot therefore arrives after `Frozen` and `OptimisticConfirmation`, rather than between them.

### Rooting follows the registry

A warp gap drains cleanly: every confirmed slot still roots.

After a persistent-mode restart, the fresh registry knows nothing about prior-run slots, so neither stream emits roots for them. Previously ws `Root` fired unconditionally while geyser `Rooted` did not.

### `CreatedBank` carries the registry's parent

The parent is the highest slot on record below the new slot: the registry's chain tip. A warp destination therefore links to a real slot instead of `slot - 1`. If nothing below the destination remains on record, it is parentless.

Other statuses carry no parent, matching Agave's notifier.

### Deep backward warps replace the timeline

A backward warp may land at or below the root line; time travel to the current epoch does exactly this. Those rooted slots have already left the registry, so the landing re-announces one with a parentless `CreatedBank`.

That is the discontinuity signal: a consumer that sees `CreatedBank` for a slot it previously saw rooted drops state at or above that slot and resyncs.

Time travel is a cheatcode, so calling it suspends the cross-timeline guarantees that rooted slots are final and slots increase. The per-bank guarantees still hold through the warp: announce before data, data before confirmation, statuses in order and once.

## 3. How the contract is checked

The lifecycle has an independent machine and spec, following the startup state machine's precedent:

* `slot_lifecycle/mod.rs` implements the machine
* `slot_lifecycle/spec.rs` encodes the transition table, one row per `(state, event)` cell; warp, root-through, and clear are set-level rules routed through those same cells

The machine never reads the spec, so the two remain independent encodings.

### The machine matches the spec

An exhaustive sweep visits every reachable registry state across every event. For each transition, it compares the machine's emissions and successor state against the spec cell.

A totality test separately ensures that every `(state, event)` cell exists.

### Production preserves the open-slot invariant

A second sweep drives the registry through production's call grammar: close a block, warp, reset.

At every reachable point, exactly one slot must be `Announced`, and it must be the open slot. This rules out both an orphaned announced slot and an open slot whose data could arrive before its announcement.

### The wire contract is tested as consumers see it

SVM-level tests observe both streams and check the external guarantees: announce before data, data before confirmation, statuses in order and once per bank, and warp statuses mirrored to ws.

### The docs come from the spec

The rustdoc table and edge diagram in `slot-lifecycle.md` render from the spec rows. A test flags stale generated blocks; `cargo surfpool-update-slot-spec` regenerates them. This is the startup spec's diagram pipeline, applied to the slot spec.


crates/core/src/runloops/mod.rs
```diff
@@ -238,6 +238,12 @@ pub async fn start_local_surfnet_runloop(
 
     let (plugin_commands_tx, plugin_commands_rx) = unbounded::<PluginCommand>();
 
+    // Startup before traffic: plugins observe startup completion and the open slot
+    // before the RPC listeners bind, so external traffic can't emit block data for
+    // a slot that a plugin is not tracking yet.
+    let _ = svm_locker.with_svm_reader(|svm| svm.geyser_events_tx.send(GeyserEvent::EndOfStartup));
+    svm_locker.with_svm_writer(|svm_writer| svm_writer.announce_open_slot());
```
MicaiahReid, line 245: Should this just happen under one lock? rather than reader than writer? Don't actually know what's better for this case

crates/core/src/surfnet/svm.rs
```diff
@@ -2523,29 +2535,8 @@ impl SurfnetSvm {
                 max_transactions_per_entry: num_transactions,
             },
         });
-        self.notify_slots_updates_subscribers(SlotUpdate::CreatedBank {
```
MicaiahReid, line 2526: I think the diff from hear down is giving me reservations. 

This code used to show me very clearly _what_ event is being emitted (`CreatedBank`, `UpdatedSlotStatus`, etc), _who_ is consuming it (ws subscribers vs `geyser_events_rx`) and _when_ (in `confirm_current_block`, after new block is made, `CreatedBank` is before `OptimisticConfirmation`, etc).

Now if I wanted to answer these same questions I have follow a lot of code. I need to inspect:
 - What events are made when we call `self.slot_lifecycle.produce(slot)`? 
 - What about `self.slot_lifecycle.confirm(slot)`? 
 - How does each of the events generated by this call get interpreted into a Geyser/WS event? 
 - Who is the audience of the events generated by these functions?

Understanding the code is a lot more difficult, and the whole state machine requires a lot of infrastructure around it (spec, tests specific to the spec, tests for each helper function, etc etc). 

It seems like a few good tests ensuring events are emitted in the correct order, and some in-line comments explaining the ordering here in this function would accomplish much of the value of the state machine, while keeping the readability and requiring less code.

Thoughts?

arunimshukla:
Cross-linking #814 because it reproduces an embedded SDK hang in the default transaction-driven block-production mode on 1.6.0.

This PR owns the slot lifecycle and confirmation ordering that the failing path now depends on. The missing coverage is specifically: `Surfnet::builder().offline(true)` with default `BlockProductionMode::Transaction`, submit a transfer, then observe it through `getSignatureStatuses`.

I have asked on #814 whether you want that bounded regression added here if the PR head fixes the hang, or kept as a compact follow-up if a behavioural change is still required. I will avoid opening a parallel implementation until that placement is clear.

PR #783, opened 2026-08-26
feat(scenarios): add Kamino support - state templates & presets

### LINKED ISSUE #742: Add Scenario Support for Kamino
Extend the coverage and provide support for a catalog of reserves and obligations, automatic address derivation, and written guidance for the model on how the protocol represents its values internally. Supported Kamino products:

- Kamino Lending Market
- Kamino Earn
- Kamino Liquidity
- Kamino Swap
- Kamino Scope
- Kamino Farms

bakasura980:
Adds native scenario support for six Kamino programs and extends scenario execution with persistent overrides and indexed IDL paths.

- Registers Kamino lending, Scope, Farms, Swap, Vault, and Liquidity IDLs and override templates.

- Adds persist to the shared override contract and re-applies persistent overrides on subsequent slots.

- Adds Kamino integration coverage, documentation, and a liquidation-arbitrage example.

The PR fixes: https://github.com/solana-foundation/surfpool/issues/742


crates/core/src/surfnet/svm.rs
```diff
@@ -2824,13 +2989,43 @@ impl SurfnetSvm {
...
+                    return Err(e);
+                }
+            };
+
+            // The request is only retired when another attempt could no longer change anything.
+            let fetch_retired = match fetch_outcome {
+                FetchOutcome::Retired => true,
+                FetchOutcome::Unanswered => false,
+                FetchOutcome::NoRemote | FetchOutcome::NotOnRemote => existing_account.is_some(),
+            };
```
MicaiahReid, line 3019: I think this logic belongs on an impl on the `FetchOutcome` struct.

MicaiahReid: Also, can we rename to `is_fetch_retired`?

crates/core/src/surfnet/svm.rs
```diff
@@ -528,6 +657,18 @@ fn synthetic_blockhash_for_slot(slot: Slot, genesis_slot: Slot) -> SyntheticBloc
     SyntheticBlockhash::new(u64::MAX - (genesis_slot - slot - 1))
 }
 
+/// What one `fetch_before_use` attempt settled. Decides whether a persisted override keeps
+/// asking on later slots, which it must while another attempt could still change the answer.
+enum FetchOutcome {
+    Retired,
```
MicaiahReid, line 663:
```suggestion
    /// This account no longer needs to be fetched because it already exists
    /// or has already successfully been fetched.
    Retired,
```
Or some sort of comment explaining this variant would be nice

crates/core/src/surfnet/svm.rs
```diff
@@ -2975,13 +3170,45 @@ impl SurfnetSvm {
...
         }
 
         Ok(())
     }
 
+    /// Re-queues `instance` for the slot after `target_slot`, replacing any copy of itself
+    /// already queued there. One entry per id, so an override cannot be applied twice to one slot.
+    fn reschedule_override_for_next_slot(
+        &mut self,
+        instance: &OverrideInstance,
```
MicaiahReid, line 3185: Is it reasonable to pass in an owned `OverrideInstance` to avoid the clone in here (since we already clone it right before passing in the instance)

crates/core/src/scenarios/protocols/kamino/README.md
```diff
@@ -0,0 +1,275 @@
+# Kamino
+
+Surfpool bundles IDLs and override templates for **six Kamino programs**, so a scenario can put a
+Kamino market into whatever state you need before your code runs against it.
+
+This is a how-to. For how scenarios work in general see the [scenarios README](../../README.md)
+every field's own purpose and units are on the template itself, visible in Studio and via
```
MicaiahReid, line 7:
```suggestion
This is a how-to. For how scenarios work in general see the [scenarios README](../../README.md).
Every field's own purpose and units are on the template itself, visible in Studio and via
```

crates/core/src/scenarios/protocols/kamino/farms/v1/overrides.yaml
```diff
@@ -0,0 +1,219 @@
...
+      3. Keep total_staked_amount consistent with the scaled totals, or reward maths and withdrawal
+         accounting disagree
+
+      Stake and reward_per_share_scaled are scaled by 2^18.
+
+      EXAMPLE - "every staker is owed more":
+      reward_infos.0.reward_per_share_scaled: 5000000
+
+      DO NOT set persist: true here - transactions write these fields, and re-applying the
+      override reverts their writes at the start of every following slot.
```
MicaiahReid, line 101: I am seeing this `DO NOT set persise: true` across the new overrides, but I'm not seeing when they _should_ set `persist: true`.

I'd rather have the direction be: don't set persist unless dictated by llm context. Otherwise, if we have to be sure to add a "DO NOT", then we need to update all other overrides to do the same.

MicaiahReid (changes requested):
I'm not quite following why the new `persist` field is needed on override instance. 

If in slot 1 we set some account state, the account state is still there in slot 2. This is particularly true if no transactions are targetting that account, which the persist docs explicitly call out as being a prerequisite to using the field.

So, why do we need to re-write the scenario override each slot?

MicaiahReid (changes requested):
I'm also seeing red CI - is that due to these changes? LMK if this is an upstream issue that I need to resolve

bakasura980:
> I'm also seeing red CI - is that due to these changes? LMK if this is an upstream issue that I need to resolve

It seems the issue is not related to the Kamino PR. The shared tests were failing because the public Solana RPC was temporarily unreachable.

bakasura980:
> I'm not quite following why the new `persist` field is needed on override instance.
> 
> If in slot 1 we set some account state, the account state is still there in slot 2. This is particularly true if no transactions are targetting that account, which the persist docs explicitly call out as being a prerequisite to using the field.
> 
> So, why do we need to re-write the scenario override each slot?

We introduced persist while addressing Kamino’s recalculated health fields, then fixed the scenario by overriding its Scope price and reserve threshold instead. Kamino indeed does not need persistence, but we kept the generic capability. We use it later for the PMMs.

MicaiahReid (approved):

PR #787, opened 2026-08-27
fix(types): require explicit delivery classes for every known event variant

cds-amal:
The forwarding match ends in a wildcard arm. A future log-class variant could therefore silently default to lossless delivery and block while the SVM lock is held, defeating the safety property enforced by the sender type.

- Name all twelve emit-class variants explicitly.
- Make the compiler reject new `SimnetEvent` variants until their delivery class is chosen.

### DIFF
--- a/crates/types/src/types.rs
+++ b/crates/types/src/types.rs
@@ -633,10 +633,7 @@ impl SimnetEventsTx {
     /// Routes a pre-built event by its class: telemetry and guard-context
     /// state events go through `log`, lifecycle events through `emit`. For
     /// forwarding buffered events (the startup replay); a new call site
-    /// names its event instead.
-    ///
-    /// Wildcard-free on purpose: a new variant does not compile until its
-    /// delivery class is chosen here.
+   /// names its event instead.
     pub fn forward(&self, event: SimnetEvent) {
         match &event {
             SimnetEvent::InfoLog(..)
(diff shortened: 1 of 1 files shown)


crates/types/src/types.rs
```diff
@@ -634,6 +634,9 @@ impl SimnetEventsTx {
     /// state events go through `log`, lifecycle events through `emit`. For
     /// forwarding buffered events (the startup replay); a new call site
     /// names its event instead.
+    ///
+    /// Wildcard-free on purpose: a new variant does not compile until its
+    /// delivery class is chosen here.
```
```suggestion
   /// names its event instead.
```
Rationale: when I'm about to call `forward` and am reading it's doc comments, I don't need to know why internally it doesn't use a wildcard in the match.

IMO, if I see every enum variant listed out, I know it's intentional for this reason. But if you like the reasoning this comment gives (which I do too), I think it belongs internally, over `CoreStarted` when the `| ` matches begin, not in the doc comment.

crates/types/src/types.rs
```diff
@@ -633,7 +633,7 @@ impl SimnetEventsTx {
     /// Routes a pre-built event by its class: telemetry and guard-context
     /// state events go through `log`, lifecycle events through `emit`. For
     /// forwarding buffered events (the startup replay); a new call site
-    /// names its event instead.
+   /// names its event instead.
```
```suggestion
    /// names its event instead.
```

MicaiahReid (approved):

PR #790, opened 2026-08-31
fix(core): serialize postgres table creation with a tx-scoped advisory lock

cds-amal:
## What was failing

`CREATE TABLE IF NOT EXISTS` reads as one step but runs as two: check whether the table exists, then create it. Two connections setting up storage for the same table at the same time can both pass the check. Both then create, and the loser fails with:

    duplicate key value violates unique constraint "pg_type_typname_nsp_index"

So construction errors out over a table that now exists.

The race is only open the first time a table name is created in a database, which makes it a rare CI flake but easy to force:

- PR #730, run 30679928707:
  `test_blockhash_consistency_across_calls::with_postgres_db`
- main at e4e3c546 (push build for #774), run 33210436114:
  `test_account_update_removes_old_indexes::with_postgres_db`
- the reproducer in this PR, which starts the sessions on a shared
  barrier, loses 48 of 50 attempts
  
I modeled the interactions with pml and spin to find race. pml SO GOOD.

## The fix

Before creating the table, each session takes an advisory lock keyed on the table name. An advisory lock is a named mutex the database hands out: it locks no table or row, it just makes sessions that ask for the same name proceed one at a time. The second session waits for the first to commit, so its existence check sees the finished table and skips the create. The lock lives in the database, so constructors in other processes serialize the same way.

PostgreSQL offers two flavors, and the choice is the safety argument:

| Flavor                                | Released by                               | If an error hits between lock and unlock                                                                                                              |
|---------------------------------------|-------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------|
| session (`pg_advisory_lock`)          | an explicit unlock call                   | the lock leaks; on a pooled connection it rides back into the pool on a healthy idle session, and every later constructor of that table waits forever |
| transaction (`pg_advisory_xact_lock`) | the database, at commit or rollback alike | nothing to leak; no code path can forget the release                                                                                                  |

This PR uses the transaction-scoped flavor.

## Tests

- Reproducer (RED commit): four sessions construct storage for a fresh
  table, released together by a barrier. Red before the fix,
  deterministically green after.
- Two contract tests (GREEN commit): a separate session probes that
  the lock is free after construction returns, on the success path and
  after a forced create failure.
- Both contract tests were verified by mutation: swapping in the
  session-scoped lock makes both fail.
- The mutation also exposed a probe bug: a probe drawn from the shared
  pool can land on the very connection that leaked the lock, and a
  session sees its own lock as free. The probe therefore opens a
  dedicated connection, a distinct session by construction.

## Testing

Any PostgreSQL 15+ works; without the environment variable the three new tests skip. A compose file matching the CI service container:

```yml
services:
  db:
    image: postgres
    restart: always
    shm_size: 128mb
    environment:
      POSTGRES_USER: surfpool
      POSTGRES_PASSWORD: surfpool
      POSTGRES_DB: surfpool_test
    ports:
      - 5432:5432
```

    SURFPOOL_TEST_POSTGRES_URL="postgres://surfpool:surfpool@localhost:5432/surfpool_test" \
      cargo test -p surfpool-core --features postgres --lib storage::postgres

### DIFF
--- a/crates/core/src/storage/postgres.rs
+++ b/crates/core/src/storage/postgres.rs
@@ -135,17 +135,8 @@ where
         debug!("Getting connection from pool for table creation");
         let mut conn = self.pool.get().map_err(|_| StorageError::LockError)?;
 
-        // IF NOT EXISTS is a catalog lookup then a create, atomic only
-        // against itself: two sessions can both pass the lookup, and the
-        // loser fails on a duplicate key in pg_type. An advisory lock keyed
-        // by table name serializes the two halves across sessions. The lock
-        // must be transaction-scoped: a session-scoped lock survives an
-        // error between lock and unlock, rides its pooled connection back
-        // into the pool, and parks every later constructor forever. Per
-        // the PostgreSQL docs, an xact lock "is automatically released at
-        // the end of the current transaction and cannot be released
-        // explicitly": release is the database's job alone, on commit and
-        // rollback alike, so no error path can leak it.
+        // pg_advisory_xact_lock serializes this transaction, at the server,
+        // against every other session taking the same key, preventing a race.
         conn.transaction(|conn| {
             sql_query("SELECT pg_advisory_xact_lock(hashtext('surfpool:ddl:' || $1))")
                 .bind::<Text, _>(&self.table_name)
--- a/crates/core/src/storage/postgres.rs
+++ b/crates/core/src/storage/postgres.rs
@@ -379,6 +379,8 @@ where
 mod tests {
     use std::sync::{Arc, Barrier};
 
+    use surfpool_db::diesel::QueryableByName;
+
     use super::*;
     use crate::storage::tests::{POSTGRES_TEST_URL_ENV, random_surfnet_id};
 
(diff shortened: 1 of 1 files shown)


crates/core/src/storage/postgres.rs
```diff
@@ -135,8 +135,24 @@ where
...
+        // against itself: two sessions can both pass the lookup, and the
+        // loser fails on a duplicate key in pg_type. An advisory lock keyed
+        // by table name serializes the two halves across sessions. The lock
+        // must be transaction-scoped: a session-scoped lock survives an
+        // error between lock and unlock, rides its pooled connection back
+        // into the pool, and parks every later constructor forever. Per
+        // the PostgreSQL docs, an xact lock "is automatically released at
+        // the end of the current transaction and cannot be released
+        // explicitly": release is the database's job alone, on commit and
+        // rollback alike, so no error path can leak it.
```
MicaiahReid: I think you know what I'm going to say about this  😅😮‍💨

MicaiahReid (changes requested):

crates/core/src/storage/postgres.rs
cds-amal: :)

MicaiahReid (approved):

PR #795, opened 2026-09-03
fix(rpc): omit empty simulateTransaction return data

xternet:
## Problem

`get_simulate_transaction_result` only omitted empty return data when its program ID was the System Program.

LiteSVM metadata can contain another program ID with an empty data vector. Surfpool therefore returned an empty `returnData` object where canonical Solana RPC returns `null`.

## Fix

Treat any empty return-data vector as absent, regardless of program ID. Non-empty return data remains unchanged.

## Testing

- Added a regression test for empty and non-empty data with a non-system program ID.
- The regression test fails before the fix and passes afterward.
- `cargo test -p surfpool-core simulate_transaction_omits_only_empty_return_data`
- `cargo test -p surfpool-core --lib --features ignore_tests_ci -- --test-threads=20` (710 passed, 64 ignored)
- `cargo clippy --workspace --all-targets`

### DIFF
--- a/crates/core/src/rpc/full.rs
+++ b/crates/core/src/rpc/full.rs
@@ -27,7 +27,6 @@ use solana_pubkey::Pubkey;
 use solana_rpc_client_api::response::Response as RpcResponse;
 use solana_sdk_ids::compute_budget;
 use solana_signature::Signature;
-use solana_system_interface::program as system_program;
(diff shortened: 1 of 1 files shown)


MicaiahReid (approved):
LGTM, thanks @xternet

PR #799, opened 2026-09-05
fix(rpc): handle JSON-RPC batch requests

### LINKED ISSUE #717: JSON-RPC batch requests are rejected

michael-moffett:
Fixes #717.

`@solana/web3.js` `Connection.getTransactions([sigA, sigB])` now works against a surfnet instead of
failing with `TypeError: unsafeRes.map is not a function`.

One file, `crates/core/src/rpc/mod.rs`. `SurfpoolMiddleware::on_request` bound only
`Request::Single(Call::MethodCall(..))` and rejected everything else, so a top-level array was discarded
before the handler saw it. `Request::Batch` is now handled there: the `surfnet_` cheatcode guard runs
once per element, through the same function the single-request path calls, and the elements that pass it
are forwarded together. Clients correlate on `id` rather than on array position.

The rest of JSON-RPC 2.0 §6 follows from that. A notification (an element with no `id`) is omitted
from the response array. A batch of nothing but notifications is answered with no response at all,
rather than with an empty array. An element the cheatcode guard rejects, or one the handler answers with
an error, comes back inline next to the elements that succeeded rather than failing the whole batch. An
empty array is answered with a single Invalid Request Response object, byte-identical to what this
endpoint answered `[]` with before this change.

Five tests cover it: a batch of method calls answered with an array, a notification omitted from the
batch response, an all-notification batch answered with nothing, a rejected element failing alone while
its neighbour succeeds, and an empty array answered with one Invalid Request.

A lone notification sent as a single request still returns `-32600`, exactly as before. This change
does not touch the single-request path.

Verified with a `@solana/web3.js` 1.98.4 client. `connection.getTransactions([sigA, sigB])` against a
patched surfnet returned the two transactions; the same script and the same client against a stock
unpatched build raised `TypeError: unsafeRes.map is not a function`, the issue's string character for
character.

+209 / -44, one file.

### DIFF
--- a/crates/core/src/rpc/mod.rs
+++ b/crates/core/src/rpc/mod.rs
@@ -6,7 +6,8 @@ use std::{
 use blake3::Hash;
 use crossbeam_channel::Sender;
 use jsonrpc_core::{
-    BoxFuture, Error, ErrorCode, FutureResponse, Metadata, Middleware, Request, Response,
+    BoxFuture, Call, Error, ErrorCode, FutureResponse, Metadata, Middleware, Output, Request,
+    Response,
     futures::{FutureExt, future::Either},
     middleware,
 };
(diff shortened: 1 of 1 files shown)


MicaiahReid (approved):
This is awesome, @michael-moffett! Thanks for the thorough testing as well.

PR #800, opened 2026-09-05
Add historical forks with --fork-slot

chen-robert:
Adds `--fork-slot <SLOT>` so the backing RPC behaves as if frozen at that slot. Local transactions and slots continue advancing normally. Behavior without the flag stays unchanged.

### DIFF
--- a/crates/cli/src/cli/mod.rs
+++ b/crates/cli/src/cli/mod.rs
@@ -266,6 +266,14 @@ pub struct StartNetworkOptions {
         long_help = "Fork from this datasource RPC URL. Cannot be used with --network.\n\nThis can also be set with SURFPOOL_DATASOURCE_RPC_URL.\n\nExample: surfpool start --rpc-url https://api.mainnet-beta.solana.com"
     )]
     pub rpc_url: Option<String>,
+    /// Load remote accounts at a finalized slot (requires Alchemy Account Archive support).
+    #[arg(
+        long,
+        value_name = "SLOT",
+        requires = "rpc_url",
+        conflicts_with = "offline"
+    )]
+    pub fork_slot: Option<u64>,
     /// Fork from a predefined Solana network.
     #[arg(
         long = "network",
@@ -647,6 +655,7 @@ impl StartSimnet {
         SurfnetSvmConfig {
             surfnet_id: self.accounts.surfnet_id.clone(),
             feature_config: self.feature_config(),
+            fork_slot: self.network.fork_slot,
             slot_time: self.svm.slot_time,
             instruction_profiling_enabled: !self.observability.disable_instruction_profiling,
             max_profiles: self.observability.max_profiles,
@@ -672,6 +681,7 @@ impl StartSimnet {
 
         SimnetConfig {
             remote_rpc_url,
+            fork_slot: self.network.fork_slot,
             slot_time: self.svm.slot_time,
             block_production_mode: self.svm.block_production_mode.clone(),
             airdrop_addresses,
--- a/crates/core/src/rpc/full.rs
+++ b/crates/core/src/rpc/full.rs
@@ -2088,12 +2088,21 @@ impl Full for SurfpoolFullRpc {
         meta: Self::Metadata,
         slot: Slot,
     ) -> BoxFuture<Result<Option<UnixTimestamp>>> {
-        let svm_locker = match meta.get_svm_locker() {
-            Ok(locker) => locker,
+        let SurfnetRpcContext {
+            svm_locker,
+            remote_ctx,
+        } = match meta.get_rpc_context(()) {
+            Ok(context) => context,
             Err(e) => return e.into(),
         };
 
         Box::pin(async move {
+            if let Some((client, _)) = &remote_ctx
+                && client.fork_slot.is_some()
+                && slot < svm_locker.with_svm_reader(|svm| svm.genesis_slot)
+            {
+                return client.get_block_time(slot).await.map_err(Into::into);
+            }
             let block_time = svm_locker.with_svm_reader(|svm_reader| {
                 Ok::<_, jsonrpc_core::Error>(match svm_reader.blocks.get(&slot)? {
                     Some(block) => Some(block.block_time),
@@ -2203,10 +2212,8 @@ impl Full for SurfpoolFullRpc {
                     let remote_end = effective_end_slot.min(local_min.saturating_sub(1));
                     if start_slot <= remote_end {
                         remote_client
-                            .client
                             .get_blocks(start_slot, Some(remote_end))
-                            .await
-                            .unwrap_or_else(|_| vec![])
+                            .await?
                     } else {
                         vec![]
                     }
@@ -2218,10 +2225,8 @@ impl Full for SurfpoolFullRpc {
                     .as_ref()
                     .unwrap()
                     .0
-                    .client
                     .get_blocks(start_slot, Some(effective_end_slot))
-                    .await
-                    .unwrap_or_else(|_| vec![])
+                    .await?
             } else {
                 vec![]
             };
@@ -2302,11 +2307,18 @@ impl Full for SurfpoolFullRpc {
                 if start_slot < local_min {
                     let remote_end = committed_latest_slot.min(local_min.saturating_sub(1));
                     if start_slot <= remote_end {
-                        remote_client
-                            .client
-                            .get_blocks(start_slot, Some(remote_end))
-                            .await
-                            .unwrap_or_else(|_| vec![])
+                        if remote_client.fork_slot.is_some() {
+                            remote_client
+                                .get_blocks_with_limit(start_slot, limit)
+                                .await?
+                                .into_iter()
+                                .take_while(|slot| *slot <= remote_end)
+                                .collect()
+                        } else {
+                            remote_client
+                                .get_blocks(start_slot, Some(remote_end))
+                                .await?
+                        }
                     } else {
                         vec![]
                     }
@@ -2319,10 +2331,8 @@ impl Full for SurfpoolFullRpc {
                     .as_ref()
                     .unwrap()
                     .0
-                    .client
                     .get_blocks(start_slot, Some(committed_latest_slot))
-                    .await
-                    .unwrap_or_else(|_| vec![])
+                    .await?
             } else {
                 vec![]
             };
--- a/crates/core/src/surfnet/locker.rs
+++ b/crates/core/src/surfnet/locker.rs
@@ -273,10 +273,16 @@ impl SurfnetSvmLocker {
             (epoch_info, epoch_schedule, some_genesis_hash)
         };
         epoch_info.transaction_count = None;
+        let fork_clock = remote_client.get_fork_clock().await?;
 
         self.with_svm_writer(move |svm_writer| {
             svm_writer.cached_genesis_hash = some_genesis_hash;
             svm_writer.initialize(epoch_info, epoch_schedule);
+            if let Some(clock) = fork_clock {
+                svm_writer.updated_at = clock.unix_timestamp as u64 * 1_000;
+                svm_writer.genesis_updated_at = svm_writer.updated_at;
+                svm_writer.inner.set_sysvar(&clock);
+            }
         });
         Ok(())
     }
@@ -1599,6 +1605,32 @@ impl SurfnetSvmLocker {
         pubkey: &Pubkey,
         config: Option<&RpcSignaturesForAddressConfig>,
     ) -> SurfpoolContextualizedResult<Vec<RpcConfirmedTransactionStatusWithSignature>> {
+        let mut historical_config;
+        let config = if let Some((client, _)) = remote_ctx {
+            if client.fork_slot.is_some() {
+                historical_config = config.cloned().unwrap_or_default();
+                let limit = historical_config.limit.unwrap_or(1000);
+                if !(1..=1000).contains(&limit) {
+                    return Err(SurfpoolError::invalid_params(
+                        "Signature limit must be between 1 and 1000",
+                    ));
+                }
+                if historical_config
+                    .min_context_slot
+                    .take()
+                    .is_some_and(|minimum| minimum > self.get_latest_absolute_slot())
+                {
+                    return Err(SurfpoolError::invalid_params(
+                        "Minimum context slot has not been reached",
+                    ));
+                }
+                Some(&historical_config)
+            } else {
+                config
+            }
+        } else {
+            config
+        };
         let results = if let Some((remote_client, _)) = remote_ctx {
             self.get_signatures_for_address_local_then_remote(remote_client, pubkey, config)
                 .await?
@@ -2681,8 +2713,17 @@ impl SurfnetSvmLocker {
         };
         epoch_info.transaction_count = None;
 
+        let fork_clock = match remote_ctx {
+            Some(client) => client.get_fork_clock().await?,
+            None => None,
+        };
         self.with_svm_writer(move |svm_writer| {
             let _ = svm_writer.reset_network(epoch_info, epoch_schedule);
+            if let Some(clock) = fork_clock {
+                svm_writer.updated_at = clock.unix_timestamp as u64 * 1_000;
+                svm_writer.genesis_updated_at = svm_writer.updated_at;
+                svm_writer.inner.set_sysvar(&clock);
+            }
             let _ = svm_writer.offline_accounts.clear();
         });
         Ok(())
@@ -2900,8 +2941,19 @@ impl SurfnetSvmLocker {
             inner: local_accounts,
         } = self.get_token_accounts_by_owner_local(owner, filter, config)?;
 
+        let mut remote_config = config.clone();
+        if remote_client.fork_slot.is_some()
+            && remote_config
+                .min_context_slot
+                .take()
+                .is_some_and(|minimum| minimum > slot)
+        {
+            return Err(SurfpoolError::invalid_params(
+                "Minimum context slot has not been reached",
+            ));
+        }
         let remote_accounts = remote_client
-            .get_token_accounts_by_owner(owner, filter, config)
+            .get_token_accounts_by_owner(owner, filter, &remote_config)
             .await?;
 
         let mut combined_accounts = remote_accounts;
@@ -4710,6 +4762,7 @@ mod tests {
         let expected_hash = Hash::new_from_array([8; 32]);
         let requests = Arc::new(AtomicUsize::new(0));
         let remote_client = SurfnetRemoteClient {
+            fork_slot: None,
             client: RpcClient::new_sender(
                 StartupRpcSender {
                     genesis_hash: expected_hash,
--- a/crates/core/src/surfnet/remote.rs
+++ b/crates/core/src/surfnet/remote.rs
@@ -1,4 +1,9 @@
-use std::{collections::HashMap, str::FromStr, sync::Arc, time::Duration};
+use std::{
+    collections::{HashMap, HashSet},
+    str::FromStr,
+    sync::{Arc, Mutex},
+    time::Duration,
+};
 
 use async_trait::async_trait;
 use serde_json::json;
@@ -18,7 +23,7 @@ use solana_client::{
         RpcTokenAccountBalance,
     },
 };
-use solana_clock::Slot;
+use solana_clock::{Clock, Slot};
 use solana_commitment_config::CommitmentConfig;
 use solana_epoch_info::EpochInfo;
 use solana_epoch_schedule::EpochSchedule;
@@ -33,6 +38,7 @@ use solana_rpc_client_api::client_error::{
     Error as ClientError, ErrorKind as ClientErrorKind, Result as ClientResult,
 };
 use solana_signature::Signature;
+use solana_sysvar_id::SysvarId;
 use solana_transaction_status::{EncodedConfirmedTransactionWithStatusMeta, UiConfirmedBlock};
 use surfpool_types::sanitized_datasource_url;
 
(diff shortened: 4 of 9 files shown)


crates/core/src/surfnet/remote.rs
```diff
@@ -493,8 +974,9 @@ impl SurfnetRemoteClient {
             Ok(res) => Ok(res.value),
             // A mint that exists only on this surfnet is `could not find mint` upstream. That is
             // a definite "no remote accounts", not a failed lookup, and must not discard the
-            // local accounts the caller merges with.
-            Err(e) if is_unknown_mint(filter, &e) => {
+            // local accounts the caller merges with. Historical walks propagate errors since
+            // an error may come from a later page or account hydration.
+            Err(e) if self.fork_slot.is_none() && is_unknown_mint(filter, &e) => {
```
MicaiahReid: I'm not understanding the addition of `self.fork_slot.is_none()` here, and I don't think this comment is really helping explain it

crates/core/src/surfnet/remote.rs
```diff
@@ -616,6 +1108,32 @@ impl SurfnetRemoteClient {
             .map_err(SurfpoolError::get_signatures_for_address)
     }
 
+    pub async fn get_block_time(&self, slot: Slot) -> SurfpoolResult<Option<i64>> {
+        self.client
+            .send(RpcRequest::GetBlockTime, json!([slot]))
```
MicaiahReid, line 1170: Is there a reason for `.send(RpcRequest::GetBlockTime, json!([slot]))` vs `.get_block_time(slot)`?

crates/core/src/surfnet/remote.rs
```diff
@@ -616,6 +1108,32 @@ impl SurfnetRemoteClient {
...
+            .await
+            .map_err(Into::into)
+    }
+
+    /// Historical lookup failures must not become incomplete block lists.
+    pub async fn get_blocks(&self, start: Slot, end: Option<Slot>) -> SurfpoolResult<Vec<Slot>> {
+        match self.client.get_blocks(start, end).await {
+            Err(_) if self.fork_slot.is_none() => Ok(vec![]), // Preserve the live datasource fallback.
```
MicaiahReid: why are we swallowing error if no fork slot?

crates/core/src/surfnet/remote.rs
```diff
@@ -616,6 +1108,32 @@ impl SurfnetRemoteClient {
...
 
+    pub async fn get_block_time(&self, slot: Slot) -> SurfpoolResult<Option<i64>> {
+        self.client
+            .send(RpcRequest::GetBlockTime, json!([slot]))
+            .await
+            .map_err(Into::into)
+    }
+
+    /// Historical lookup failures must not become incomplete block lists.
+    pub async fn get_blocks(&self, start: Slot, end: Option<Slot>) -> SurfpoolResult<Vec<Slot>> {
```
MicaiahReid: I'd prefer for doc comments to be documenting how to use a function, not rationale for internals. Here, can we either remove comment, update to an appropriate function usage-comment, and/or move the note about historical lookup failures where it's relevant inside the function. Or, if the detail is relevant to function consumers, put it in the context of how to use the function

crates/core/src/surfnet/remote.rs
```diff
@@ -616,6 +1108,32 @@ impl SurfnetRemoteClient {
...
+        }
+    }
+
+    pub async fn get_blocks_with_limit(
+        &self,
+        start: Slot,
+        limit: usize,
+    ) -> SurfpoolResult<Vec<Slot>> {
+        self.client
+            .send(RpcRequest::GetBlocksWithLimit, json!([start, limit]))
```
MicaiahReid: Again, why `send` instead of `get_blocks_with_limit`?

crates/core/src/surfnet/remote.rs
```diff
@@ -157,6 +164,421 @@ impl<S: RpcSender + Send + Sync> RpcSender for DeadlineSender<S> {
...
+        let address = params[0]
+            .as_str()
+            .ok_or_else(|| ClientErrorKind::Custom("Invalid address".into()))?;
+        let address_key = Pubkey::from_str(address)
+            .map_err(|_| ClientErrorKind::Custom("Invalid address".into()))?;
+        let mut config: RpcSignaturesForAddressConfig = if params[1].is_null() {
+            Default::default()
+        } else {
+            serde_json::from_value(params[1].clone())?
+        };
```
MicaiahReid: Can we just deserialize to `(String, Option<RpcSignaturesForAddressConfig>)`, something like
```rust
        let (address, config) =
            serde_json::from_value::<(String, Option<RpcSignaturesForAddressConfig>)>(params)
                .map(|(address, config)| (address, config.unwrap_or_default()))
                .map_err(|_| {
                    ClientErrorKind::Custom("Invalid signature request parameters".into())
                })?;
```
I'd like to avoid directly accessing keys on params, it's harder to reason about imo

crates/core/src/surfnet/remote.rs
```diff
@@ -157,6 +164,421 @@ impl<S: RpcSender + Send + Sync> RpcSender for DeadlineSender<S> {
...
+        let limit = config.limit.unwrap_or(1000);
+        if !(1..=1000).contains(&limit) || config.min_context_slot.is_some_and(|min| min > slot) {
+            return Err(ClientErrorKind::Custom(
+                "Invalid limit or minimum context slot for historical signatures".into(),
+            )
+            .into());
+        }
```
MicaiahReid: Can we separate out to two checks and a specific error for each check?

crates/core/src/surfnet/remote.rs
```diff
@@ -157,6 +164,421 @@ impl<S: RpcSender + Send + Sync> RpcSender for DeadlineSender<S> {
...
+
+impl<S: RpcSender + Send + Sync> ForkSender<S> {
+    async fn transaction_slot(&self, signature: &str) -> ClientResult<Slot> {
+        Signature::from_str(signature)
+            .map_err(|_| ClientErrorKind::Custom("Invalid signature cursor".into()))?;
+        let response = self.inner.send(RpcRequest::GetTransaction,
+            json!([signature, {"encoding": "base64", "commitment": "finalized", "maxSupportedTransactionVersion": 0}])).await?;
+        response["slot"].as_u64().ok_or_else(|| {
+            ClientErrorKind::Custom("Cannot resolve historical signature cursor".into()).into()
+        })
```
MicaiahReid: Would prefer a proper building of the `RpcTransactionConfig` struct and serializing, plus deserializing a result

crates/core/src/surfnet/remote.rs
```diff
@@ -157,6 +164,421 @@ impl<S: RpcSender + Send + Sync> RpcSender for DeadlineSender<S> {
...
+                if is_before {
+                    cursor_slot = resolved_slot;
+                } else {
+                    until_slot = Some(resolved_slot);
+                }
+            }
+        }
+        if config.before.is_none()
+            && let Some((cursor, cached_slot)) =
+                self.signature_cursors.lock().unwrap().get(&address_key)
```
MicaiahReid: Can we return error if lock can't be acquired?

crates/core/src/surfnet/remote.rs
```diff
@@ -157,6 +164,421 @@ impl<S: RpcSender + Send + Sync> RpcSender for DeadlineSender<S> {
...
+                        "Invalid or repeating historical signature page".into(),
+                    )
+                    .into());
+                }
+                cursor_slot = entry.slot;
+            }
+            // Binary search the descending page for the first signature at or before S.
+            let cutoff = page.partition_point(|entry| entry.slot > slot);
+            if cutoff > 0 {
+                let mut cursors = self.signature_cursors.lock().unwrap();
```
MicaiahReid: Can we return error if lock can't be acquired?

crates/core/src/surfnet/remote.rs
```diff
@@ -157,6 +164,421 @@ impl<S: RpcSender + Send + Sync> RpcSender for DeadlineSender<S> {
     }
 }
 
+/// Pins account hydration and bounds remote history to one slot.
+/// Queries without a historical source fail rather than mixing current state into the fork.
+struct ForkSender<S> {
+    inner: S,
+    slot: Option<Slot>,
+    signature_cursors: Mutex<HashMap<Pubkey, (String, Slot)>>,
```
MicaiahReid: A doc comment along the lines of
```rust
  /// Stores a map of addresses to their last signature/slot that is after
  /// the fork slot. Used to shortcut against re-scanning post-fork history
  /// when searching signatures for an address.
```

MicaiahReid: Also a rename of the tuple to something like `PostForkBoundary { signature, slot };` would make groking this quicker for future devs

crates/core/src/surfnet/remote.rs
```diff
@@ -157,6 +164,421 @@ impl<S: RpcSender + Send + Sync> RpcSender for DeadlineSender<S> {
...
+            .and_then(serde_json::Value::as_u64)
+            .is_some_and(|minimum| minimum > slot)
+        {
+            return Err(
+                ClientErrorKind::Custom("Minimum context slot is after the fork".into()).into(),
+            );
+        }
+        config.remove("minContextSlot");
+        config.insert("slot".into(), json!(slot));
+        config.insert("commitment".into(), json!("finalized"));
```
MicaiahReid: Something like
```rust
  #[derive(Deserialize)]
  struct GetAccountInfoRequest(
      Pubkey,
      #[serde(default)] Option<RpcAccountInfoConfig>,
  );

  // Archive-specific wire shape sent downstream.
  #[derive(Serialize)]
  struct ArchiveAccountInfoConfig {
      #[serde(flatten)]
      account: RpcAccountInfoConfig,
      slot: Slot,
  }

  Then conversion owns the policy:

  impl GetAccountInfoRequest {
      fn into_archive_request(
          self,
          slot: Slot,
      ) -> ClientResult<(Pubkey, ArchiveAccountInfoConfig)> {
          let (pubkey, mut config) = self;
          let mut account = config.unwrap_or_default();

          if account.min_context_slot.is_some_and(|minimum| minimum > slot) {
              return Err(/* clear error */);
          }

          account.min_context_slot = None;
          account.commitment = Some(CommitmentConfig::finalized());

          Ok((pubkey, ArchiveAccountInfoConfig { account, slot }))
      }
  }
```

would make these transforms much clearer and more explicit. I don't love the direct json key updating

crates/core/src/surfnet/remote.rs
```diff
@@ -157,6 +164,421 @@ impl<S: RpcSender + Send + Sync> RpcSender for DeadlineSender<S> {
...
+                let pubkeys = params[0]
+                    .as_array()
+                    .ok_or_else(|| ClientErrorKind::Custom("Invalid account list".into()))?;
+                // Archive providers may only support historical getAccountInfo.
+                let responses = jsonrpc_core::futures::future::try_join_all(
+                    pubkeys
+                        .iter()
+                        .map(|pubkey| self.account(pubkey.clone(), params[1].clone(), slot)),
+                )
+                .await?;
```
MicaiahReid: This seems like a major performance loss over a "may only support". Is this a known lack of support among archive providers? Should we do an exploratory check first and fallback to this?

crates/cli/src/cli/mod.rs
```diff
@@ -266,6 +266,14 @@ pub struct StartNetworkOptions {
         long_help = "Fork from this datasource RPC URL. Cannot be used with --network.\n\nThis can also be set with SURFPOOL_DATASOURCE_RPC_URL.\n\nExample: surfpool start --rpc-url https://api.mainnet-beta.solana.com"
     )]
     pub rpc_url: Option<String>,
+    /// Load remote accounts at a finalized slot (requires Alchemy Account Archive support).
```
MicaiahReid: Would rather not explicitly call out archive providers. Does the wire format we're using or `ForkSender` requests work for any other archive providers? If so awesome. If not, it would be out of scope for this PR to support all, we can make the `ForkSender` more robust in the future. But if that's the case, I think this wording should be more like "currently, only Alchemy Account Archives are supported"

crates/core/src/rpc/full.rs
```diff
@@ -2088,12 +2088,21 @@ impl Full for SurfpoolFullRpc {
...
+            remote_ctx,
+        } = match meta.get_rpc_context(()) {
+            Ok(context) => context,
             Err(e) => return e.into(),
         };
 
         Box::pin(async move {
+            if let Some((client, _)) = &remote_ctx
+                && client.fork_slot.is_some()
+                && slot < svm_locker.with_svm_reader(|svm| svm.genesis_slot)
```
MicaiahReid, line 2101:
```suggestion
            if let Some((client, _)) = &remote_ctx
                && slot < svm_locker.with_svm_reader(|svm| svm.genesis_slot)
```
Good fix to get remote block time if the requested slot is before local genesis. But I think this fix applies whether `fork_slot` is provided or not

crates/core/src/surfnet/remote.rs
```diff
@@ -206,13 +636,64 @@ impl SurfnetRemoteClient {
     }
 
     pub fn try_new<U: ToString>(remote_rpc_url: U) -> Result<Self, reqwest::Error> {
```
MicaiahReid: This function is still used in `crates/core/src/runloops/mod.rs:573`, and it should also be upgraded to use `try_new_at_slot`. From there I believe we can remove the `pub` from this guy

crates/core/src/surfnet/locker.rs
```diff
@@ -273,10 +273,16 @@ impl SurfnetSvmLocker {
...
 
         self.with_svm_writer(move |svm_writer| {
             svm_writer.cached_genesis_hash = some_genesis_hash;
             svm_writer.initialize(epoch_info, epoch_schedule);
+            if let Some(clock) = fork_clock {
+                svm_writer.updated_at = clock.unix_timestamp as u64 * 1_000;
+                svm_writer.genesis_updated_at = svm_writer.updated_at;
+                svm_writer.inner.set_sysvar(&clock);
+            }
         });
```
MicaiahReid, line 286: The `svm_writer.initialize` call here is updating the `updated_at`, `genesis_updated_at`, and clock sysvar according to the values reeturned rom the epoch info fetch from the remote client.

So if the remote client has done its job and fetched only data up to `fork_slot` for the `get_epoch_info`, `get_epoch_schedule`, and `get_genesis_hash` request, shouldn't these values already be correct from the `initialize` call?
```suggestion

        self.with_svm_writer(move |svm_writer| {
            svm_writer.cached_genesis_hash = some_genesis_hash;
            svm_writer.initialize(epoch_info, epoch_schedule);
```

```suggestion
        self.with_svm_writer(move |svm_writer| {
            svm_writer.cached_genesis_hash = some_genesis_hash;
            svm_writer.initialize(epoch_info, epoch_schedule);
        });
```

crates/core/src/rpc/full.rs
```diff
@@ -2302,11 +2307,18 @@ impl Full for SurfpoolFullRpc {
...
+                                .get_blocks_with_limit(start_slot, limit)
+                                .await?
+                                .into_iter()
+                                .take_while(|slot| *slot <= remote_end)
+                                .collect()
+                        } else {
+                            remote_client
+                                .get_blocks(start_slot, Some(remote_end))
+                                .await?
+                        }
```
MicaiahReid: I'd rather keep such conditional logic regarding `fork_slot` centralized in the `remote` mod as mush as possible. Could we not do
```
                            remote_client
                                .get_blocks_with_limit(start_slot, limit)
                                .await?
                                .into_iter()
                                .take_while(|slot| *slot <= remote_end)
                                .collect()
```
for both cases here?

crates/core/src/surfnet/locker.rs
```diff
@@ -1599,6 +1605,32 @@ impl SurfnetSvmLocker {
...
+                }
+                if historical_config
+                    .min_context_slot
+                    .take()
+                    .is_some_and(|minimum| minimum > self.get_latest_absolute_slot())
+                {
+                    return Err(SurfpoolError::invalid_params(
+                        "Minimum context slot has not been reached",
+                    ));
+                }
```
MicaiahReid: All of these checks should be made anyways, regardless of `fork_slot.is_some()`

crates/core/src/surfnet/locker.rs
```diff
@@ -2681,8 +2713,17 @@ impl SurfnetSvmLocker {
...
+            Some(client) => client.get_fork_clock().await?,
+            None => None,
+        };
         self.with_svm_writer(move |svm_writer| {
             let _ = svm_writer.reset_network(epoch_info, epoch_schedule);
+            if let Some(clock) = fork_clock {
+                svm_writer.updated_at = clock.unix_timestamp as u64 * 1_000;
+                svm_writer.genesis_updated_at = svm_writer.updated_at;
+                svm_writer.inner.set_sysvar(&clock);
+            }
```
MicaiahReid, line 2719: I have the same note here as I had about `initialize` fn in locker - the fork-specific logic here should not be necessary. It should be handled when the epoch info/schedule are fetched from remote above

crates/core/src/surfnet/locker.rs
```diff
@@ -2900,8 +2941,19 @@ impl SurfnetSvmLocker {
...
+        if remote_client.fork_slot.is_some()
+            && remote_config
+                .min_context_slot
+                .take()
+                .is_some_and(|minimum| minimum > slot)
+        {
+            return Err(SurfpoolError::invalid_params(
+                "Minimum context slot has not been reached",
+            ));
+        }
```
MicaiahReid: This check should be made regardless of `remote_client.fork_slot.is_some()`

MicaiahReid (changes requested):
I believe this should have some downstream side effects in the sdk/sdk-node crates as well to update the wire format

crates/cli/src/cli/mod.rs
chen-robert: Yeah it seems like currently only Alchemy supports it. I know Cavey was working on something similar, would be good to add support once that lands. Will update the wording

crates/core/src/rpc/full.rs
chen-robert: This logic is a bit convoluted because we want to keep the `unwrap_or_else(|_| vec![])` for non-fork slots (which arguably is a bug), and fix it for fork slot. I think it's easier to standardize on `unwrap_or_else(|_| vec![])` for now, and we can punt on if we want to fix it later.

crates/core/src/surfnet/locker.rs
chen-robert: I think this code was being overly defensive to preserve existing behavior, but agree these checks are worth landing even if it slightly breaks compatibility

crates/core/src/surfnet/locker.rs
chen-robert: Agreed, same as before

crates/core/src/rpc/full.rs
chen-robert: Will land

crates/core/src/surfnet/locker.rs
chen-robert: The issue is `initialize` sets the timestamp to the current timestamp, not the historical one. Which could break programs relying on an accurate historical timestamp

https://github.com/solana-foundation/surfpool/blob/b74285ff6ec9725c31ecd3f0f6a6310204b342a5/crates/core/src/surfnet/svm.rs#L1188

crates/core/src/surfnet/locker.rs
chen-robert: Same as above, don't think this is correct

crates/core/src/surfnet/remote.rs
chen-robert: Will do

crates/core/src/surfnet/remote.rs
chen-robert: Will do

crates/core/src/surfnet/remote.rs
chen-robert: Will do

crates/core/src/surfnet/remote.rs
chen-robert: Will do

crates/core/src/surfnet/remote.rs
chen-robert: Unfortunately it's only getAccountInfo as of today, but I'll update the comment to be clearer

https://www.alchemy.com/docs/solana/account-archive

crates/core/src/surfnet/remote.rs
```diff
@@ -206,13 +636,64 @@ impl SurfnetRemoteClient {
...
+                &slot,
+                RpcBlockConfig {
+                    commitment: Some(CommitmentConfig::finalized()),
+                    transaction_details: Some(solana_transaction_status::TransactionDetails::None),
+                    rewards: Some(false),
+                    max_supported_transaction_version: Some(0),
+                    ..Default::default()
+                },
+            )
+            .await?;
```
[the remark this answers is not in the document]
chen-robert: Seems too edge-case to be worth addressing

crates/core/src/surfnet/remote.rs
chen-robert: Will delete

crates/core/src/surfnet/remote.rs
chen-robert: Will fix per previous discussion about simplifying the get_blocks logic

crates/core/src/surfnet/remote.rs
chen-robert: `.get_block_time` returns an [error on null](https://github.com/anza-xyz/agave/blob/c4b48df969a9e4f121e14a389bd7bec34c752507/rpc-client/src/nonblocking/rpc_client.rs#L3037), which we can convert instead to `None`, which is a bit more ergonomic

crates/core/src/surfnet/remote.rs
chen-robert: Will revert

crates/core/src/surfnet/remote.rs
chen-robert: This is because in fork mode, the previous `RpcRequest::GetTokenAccountsByOwner` call is doing a lot more work. But I do agree this feels a bit too defensive, we can revert this change.

crates/core/src/surfnet/remote.rs
chen-robert: Will fix

